PR #120, opened 2014-10-02
add supported languages check and convert to smallest denomination in no...

iamkoby:
i found it useful for us.

koby


kyleconroy:
@iamkoby Sorry we never responded to this :(

Since our currency support changes often, we don't want to hard code a list in our helper libraries. Thanks for taking the time to make a PR. I promise we'll do better in the future.

PR #135, opened 2015-01-07
Support Bitcoin

maxwells:
Added Bitcoin Receiver & Transaction objects


maxwells:
WIP

lib/resources/BitcoinReceivers.js
```diff
@@ -0,0 +1,19 @@
...
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+
+  path: 'bitcoin/receivers',
+
+  includeBasic: [
+    'create', 'list', 'retrieve'
```
michelle: Can't update, correct?

lib/resources/BitcoinReceivers.js
```diff
@@ -0,0 +1,19 @@
...
+  ],
+
+  listBitcoinTransactions: stripeMethod({
+    method: 'GET',
+    path: '/{receiverId}/transactions',
```
michelle: Do BitcoinTransactions not need their own resource? Can't retrieve by ID/update?

test/resources/BitcoinReceivers.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+    });
+
+  });
+
+  describe('listBitcoinTransactions', function() {
+
+    it('Sends the correct request', function() {
+
+      stripe.bitcoinReceivers.listBitcoinTransactions('receiverId', {
```
michelle: Is this naming redundant? `listBitcoinTransactions`

michelle:
Few notes, mostly due to me not having state on the API, but LGTM!

lib/resources/BitcoinReceivers.js
maxwells: correct. they're only create/list-able by API users.

lib/resources/BitcoinReceivers.js
maxwells: The API is exposed, but after chatting with ray & christian about it, we decided there wasn't a strong story for adding it to the bindings as a top level object. They are useful as a sublist of a receiver, though so I stuck with that.

test/resources/BitcoinReceivers.spec.js
maxwells: Yeah. `listTransactions` would be better. I'll touch that up.

Thanks for reviewing these changes!

lib/resources/BitcoinReceivers.js
michelle: Is there any functionality we'd be missing out on w/r/t transactions?

maxwells: they're immutable from the user API, so not that I can think of. These transactions represent the transfer of bitcoins to fill the receiver before it's used in a charge -- they're very specific to the receiver itself.

michelle: :100:

PR #144, opened 2015-02-13
Add support for accounts resource

bkrausz:
This adds support for accessing accounts via either the /v1/account or /v1/accounts endpoints.

r? @michelle 
CC @danwang

### DIFF
--- a/lib/resources/Account.js
+++ b/lib/resources/Account.js
@@ -1,41 +1,7 @@
 'use strict';
 
-var StripeResource = require('../StripeResource');
-var stripeMethod = StripeResource.method;
-var utils = require('../utils');
-
-module.exports = StripeResource.extend({
-  // Since path can either be `account` or `accounts`, support both through stripeMethod path
-
-  create: stripeMethod({
-    method: 'POST',
-    path: 'accounts'
-  }),
-
-  list: stripeMethod({
-    method: 'GET',
-    path: 'accounts'
-  }),
-
-  update: stripeMethod({
-    method: 'POST',
-    path: 'accounts/{id}',
-    urlParams: ['id']
-  }),
-
-  retrieve: function(id) {
-    // Old bindings supported no-arg access to /v1/account, support that if ID is not passed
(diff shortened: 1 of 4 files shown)


lib/resources/Account.js
```diff
@@ -2,12 +2,40 @@
 
 var StripeResource = require('../StripeResource');
 var stripeMethod = StripeResource.method;
+var utils = require('../utils');
 
 module.exports = StripeResource.extend({
-  path: 'account',
+  // Since path can either be `account` or `accounts`, support both through stripeMethod path
```
michelle: The semantics of using v1/accounts v. v1/account are pretty different, right? Like it'd be weird to see all these methods available but still be using the /account singleton retrieve. My preference is actually to create a new resource called "Accounts" for the new stuff, so you need to explicitly switch over.

cc @jimdanz since this was brought up during review but I'm not sure we reached a conclusion.

bkrausz: Nope, `/v1/account` is (in theory) 100% identical to `/v1/accounts/your_acct_id`. Essentially the latter acts as a convenience method to the former with a `Stripe-Account: your_acct_id` header.

bkrausz: That being said, permissions will dictate what fields you can see and use.

michelle: Sure, I guess I mean that if you're using "stripe.account" as a singleton
to grab your own account info, your context/use for it is probably
different than if you were using "stripe.accounts"--I imagine not everyone
will need to use or care to think about "stripe.accounts", though I've
definitely not thought about this as much as you have.

- Michelle

bkrausz: I think the people who care about their own account is notoriously small anyway, it seems like 100% a Connect use case. But we will over time consider exposing this to managing your own account (i.e. you can update your own account), so consolidation seems to make sense to me. Happy to IRL though, I'll find you.

bkrausz:
IRLed with @michelle, realized the main FUD was the plural/singular naming. Just aliased one to the other. Shipping.

PR #146, opened 2015-02-18
added structure to support unified customer sources api

maxwells:
added structure to support unified customer sources api; added delete and update methods to BitcoinReceiever

cc @raycmorgan @canderson


lib/resources/Customers.js
```diff
@@ -111,6 +111,40 @@ module.exports = StripeResource.extend({
   }),
 
   /**
+   * Customer: Source methods
+   */
+
+  createSource: stripeMethod({
```
michelle: Are we leaving the card methods in?

maxwells: We are for now to provide a less bumpy upgrade path

SlexAxton:
Let me or @michelle know if you wanna push this out as a release to npm.

PR #148, opened 2015-02-21
Check if a callback is passed to Retrieve Customer

### LINKED ISSUE #147: Cannot retrieve own account without specifying key
Per the API docs I should just be able to do

    var stripe = require("stripe")("<api key>");
    
    stripe.account.retrieve(function(err, account) {
      // asynchronously called
    });

This will throw an ugly error though:

    /Users/matthewarkin/Developer/test1/node_modules/stripe/lib/StripeMethod.js:47
            throw new Error('Stripe: I require argument "' + urlParams[i] + '", bu
                  ^
    Error: Stripe: I require argument "id", but I got: undefined
        at Object.<anonymous> (/Users/matthewarkin/Developer/test1/node_modules/stripe/lib/StripeMethod.js:47:15)
        at Object.module.exports.StripeResource.extend.retrieve (/Users/matthewarkin/Developer/test1/node_modules/stripe/lib/resources/Accounts.js:40:10)
        at Object.<anonymous> (/Users/matthewarkin/Developer/test1/app.js:14:16)
        at Module._compile (module.js:456:26)
        at Object.Module._extensions..js (module.js:474:10)
        at Module.load (module.js:356:32)
        at Function.Module._load (module.js:312:12)
        at Function.Module.runMain (module.js:497:10)
        at startup (node.js:119:16)
        at node.js:906:3

If you use a promise though it works fine:

    stripe.account.retrieve().then(console.log);

Looks like it was caused by the if check introduced in [8db8f0](https://github.com/stripe/stripe-node/blob/8db8f0d5ed5fb47a7f1f4702b4bd96f8ac5d1e0a/lib/resources/Account.js#L28) which fails to handle the case where you pass in a function.

matthewarkin:
If a parameter is passed to Retrieve customer that isn't an
auth key, the library incorrectly assumed it was a account id.

Updated the check to ensure that a function could be passed in as
a sole argument and it would retrieve the account for the pre-set
api key.

Fixes #147

### DIFF
--- a/lib/resources/Accounts.js
+++ b/lib/resources/Accounts.js
@@ -24,8 +24,8 @@ module.exports = StripeResource.extend({
   }),
 
   retrieve: function(id) {
-    // Old bindings supported no-arg access to /v1/account, support that if ID is not passed
-    if (id === undefined || (typeof id === 'string' && utils.isAuthKey(id))) {
+    // Old bindings supported no-arg / callback access to /v1/account, support that if ID is not passed
+    if (id === undefined || typeof id === 'function' || (typeof id === 'string' && utils.isAuthKey(id))) {
       return stripeMethod({
         method: 'GET',
         path: 'account'
--- a/test/resources/Account.spec.js
+++ b/test/resources/Account.spec.js
@@ -42,6 +42,17 @@ describe('Account Resource', function() {
         data: {},
         headers: {},
       });
+    });
+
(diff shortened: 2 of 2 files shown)


michelle:
Ah, sorry, you mean retrieve account, right?

Great catch!

matthewarkin:
Yeah, what I get for writing commit messages at 3am



—

PR #150, opened 2015-02-24
Support transfer reversals

danwang:
r? @michelle @bkrausz @SlexAxton 

Support CRUD of transfer reversal objects, and update `cancel` to use `/reversals` endpoint.

### DIFF
--- a/lib/resources/TransferReversals.js
+++ b/lib/resources/TransferReversals.js
@@ -1,0 +1,21 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+
+/**
+ * TransferReversals is a unique resource in that, upon instantiation,
+ * requires a transferId, and therefore each of its methods only
+ * require the reversalId argument.
+ *
+ * This streamlines the API specifically for the case of accessing reversals
+ * on a returned transfer object.
+ *
+ * E.g. transferObject.reversals.retrieve(reversalId)
+ * (As opposed to the also-supported stripe.transfers.retrieveReversal(transferId,
+ * reversalId))
+ */
+module.exports = StripeResource.extend({
+  path: 'transfers/{transferId}/reversals',
+  includeBasic: ['create', 'list', 'retrieve', 'update']
+});
+
--- a/lib/resources/Transfers.js
+++ b/lib/resources/Transfers.js
@@ -6,10 +6,17 @@ var stripeMethod = StripeResource.method;
 module.exports = StripeResource.extend({
 
   path: 'transfers',
+
   includeBasic: [
     'create', 'list', 'retrieve', 'update',
     'setMetadata', 'getMetadata'
   ],
+
+  reverse: stripeMethod({
+    method: 'POST',
+    path: '/{transferId}/reversals',
+    urlParams: ['transferId']
+  }),
 
   cancel: stripeMethod({
     method: 'POST',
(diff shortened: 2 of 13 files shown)


lib/resources/Transfers.js
```diff
@@ -6,14 +6,21 @@ var stripeMethod = StripeResource.method;
 module.exports = StripeResource.extend({
 
   path: 'transfers',
+
   includeBasic: [
     'create', 'list', 'retrieve', 'update',
     'setMetadata', 'getMetadata'
   ],
 
+  reverse: stripeMethod({
```
michelle: can we make it more clear that `cancel` is just an alias to `reverse`?

test/resources/Transfers.spec.js
```diff
@@ -64,7 +64,7 @@ describe('Transfers Resource', function() {
       stripe.transfers.cancel('transferId4');
       expect(stripe.LAST_REQUEST).to.deep.equal({
         method: 'POST',
-        url: '/v1/transfers/transferId4/cancel',
+        url: '/v1/transfers/transferId4/reversals',
```
michelle: can you add a test for `stripe.transfers.reverse`?

lib/resources/TransferReversals.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+
+/**
+ * TransferReversals is a unique resource in that, upon instantiation,
```
michelle: hm, i'm not sure this sentence reads quite right--what do you mean here?

michelle: Ah, I see. Basically because in order to instantiate them, you need to pass in a transfer ID, when calling methods on them, you don't need to pass the transfer ID again, correct?

lib/resources/TransferReversals.js
```diff
@@ -0,0 +1,21 @@
...
+ * requires a transferId, and therefore each of its methods only
+ * require the reversalId argument.
+ *
+ * This streamlines the API specifically for the case of accessing reversals
+ * on a returned transfer object.
+ *
+ * E.g. transferObject.reversals.retrieve(reversalId)
+ * (As opposed to the also-supported stripe.Transfer.retrieveReversal(transferId,
```
michelle: Do you mean `stripe.transfers.retrieveReversal`?

lib/resources/TransferReversals.js
```diff
@@ -0,0 +1,21 @@
...
+
+var StripeResource = require('../StripeResource');
+
+/**
+ * TransferReversals is a unique resource in that, upon instantiation,
+ * requires a transferId, and therefore each of its methods only
+ * require the reversalId argument.
+ *
+ * This streamlines the API specifically for the case of accessing reversals
+ * on a returned transfer object.
```
michelle: I'd say "retrieved" rather than "returned"--it's all async.

lib/resources/TransferReversals.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * E.g. transferObject.reversals.retrieve(reversalId)
```
michelle: Huh, interesting. Can we do this with cards, refunds, and such?

lib/resources/TransferReversals.js
danwang: Yes, I believe that's what it means.

The comment is copied from `ChargeRefunds`: https://github.com/stripe/stripe-node/blob/master/lib/resources/ChargeRefunds.js

test/resources/Transfers.spec.js
danwang: Done.

lib/resources/TransferReversals.js
danwang: Fixed. Also fixed `ChargeRefunds`.

danwang: Also fixed `ApplicationFeeRefunds`, `CustomerCards.js`, ` CustomerSubscriptions`, and `RecipientCards`.

lib/resources/TransferReversals.js
danwang: This affects all subresources:

```
  ➜  git grep "on a returned"
ApplicationFeeRefunds.js: * on a returned application fee object.
ChargeRefunds.js: * on a returned charge object.
CustomerCards.js: * on a returned customer object.
CustomerSubscriptions.js: * on a returned customer object.
RecipientCards.js: * on a returned recipient object.
TransferReversals.js: * on a returned transfer object.
```

Should all these be changed as well?

lib/resources/Transfers.js
danwang: Added a comment.

matthewarkin:
I don't see a revesal object documented anywhere at https://stripe.com/docs/api/curl#cancel_transfer. The docs use  https://api.stripe.com/v1/transfers/<transfer id>/ and all the other libraries use "cancel" instead of "reverse".

bkrausz:
Matthew: you're correct, we haven't finished documenting this new API quite
yet. It should be done this week :).

michelle:
:sheep:

Issue #145, opened 2015-02-18
Response status not available on error object


pkaminski:
The error object returned from a rejected promise doesn't appear to have the HTTP response's status code available anywhere.  I'm queuing up incoming webhook events and their handling could be arbitrarily delayed, by which point the target object might've gone away.  I'd like to distinguish 404 cases (which are OK to ignore for my system) vs other ones (which are bugs or transient errors).  Right now, it looks like the only way to do this is by scraping the error message, which I'd like to avoid.

Could you please put a status code on the error object when it originates from the REST API?  Thanks.

michelle:
Good point--we'll look into fixing this, but in the meantime, PRs are very welcome!

PR #156, opened 2015-03-08
Provide HTTP response status code to error object of a rejected promise

APTy:
Added the http status code to the error object in the main response handler so that the return from a rejected promise is more descriptive. This is in response to [Issue #145](https://github.com/stripe/stripe-node/issues/145). 

Added test to stripe spec to confirm functionality.

screen shot 2015-03-08 at 3 35 01 pm

All other tests still pass.

screen shot 2015-03-08 at 3 35 26 pm

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -129,6 +129,7 @@ StripeResource.prototype = {
         try {
           response = JSON.parse(response);
           if (response.error) {
+            response.error.code = res.statusCode; // add response status to error object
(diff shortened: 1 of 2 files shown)


kyleconroy:
Errors already have a `code` parameter. I've create a new PR that adds a `statusCode` property to the errors.

kyleconroy:
https://github.com/stripe/stripe-node/pull/198

PR #158, opened 2015-03-19
Add bank account support

djeusette:
No tests added but the bank account resources are required to use the new marketplace features.

I don't expect you to merge this PR as is, but keep in mind that a full support of the bank accounts would be great in a near future.

### DIFF
--- a/lib/resources/Accounts.js
+++ b/lib/resources/Accounts.js
@@ -38,4 +38,39 @@ module.exports = StripeResource.extend({
       }).apply(this, arguments);
     }
   },
+
+  /**
+   * Accounts: Bank account methods
+   */
+
+  createBankAccount: stripeMethod({
+    method: 'POST',
+    path: 'accounts/{accountId}/bank_accounts',
+    urlParams: ['accountId']
+  }),
+
+  listBankAccounts: stripeMethod({
+    method: 'GET',
+    path: 'accounts/{accountId}/bank_accounts',
+    urlParams: ['accountId']
+  }),
+
+  retrieveBankAccount: stripeMethod({
+    method: 'GET',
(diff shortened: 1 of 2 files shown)


matthewarkin:
I can add some tests to this if it'll help :)

djeusette:
Hi Matthew,

I've just added default tests based on what had been done for the customer cards.  Feel free to add new ones if needed.

matthewarkin:
@bkrausz (just wanted to ping you on this and a couple other PRs that could use Stripe input)

bkrausz:
Thanks for the ping! There's a small chance we'll rename this API from /bank_accounts/ to /destinations/ (to be symmetric with `sources` on customers), so let's hold off for a few more days before merging.

CC @russelldavis

lowe0292:
Any word on this? I'm migrating from Balanced to Stripe and I'd definitely appreciate bank account support :wink:

russelldavis:
We're indeed doing something along the lines of @bkrausz mentioned above, so unfortunately we won't be merging this as-is. We will add support for that here once it's ready though.

matthewarkin:
Good news everyone!

So they've started adding this to some of the other libraries!

Bad news!
Instead of the endpoint being bank_accounts its external_accounts (I'm assuming this is to also supported debit cards by July).

So this will need a quick s/bank_accounts/external_accounts

bkrausz:
Yup, if you want to update this to reference external accounts instead of bank accounts, I'm happy to merge in.

djeusette:
Hey, thanks for the news.  I've just renamed all the references.

bkrausz:
Thanks so much for adding this, and seeing it through to completion. CC @michelle but LGTM :)

michelle:
Looks good. Thanks for the patch!
>
>

-- 
Sent from Gmail Mobile

PR #164, opened 2015-03-30
Fix support for both stripe.account and stripe.accounts

bkrausz:
r? @michelle


michelle:
Can probably also get rid of `lib/resources/Account`?

:+1:

PR #169, opened 2015-04-18
Add info on how to use the Stripe-Account header

stith:
The only place I could find this info before was on [this one page](https://stripe.com/docs/connect/payments-fees) of the docs, pointed out to me in IRC. Figured the README was a much better place for it.


kyleconroy:
Agreed! Thanks for the PR

PR #180, opened 2015-06-10
Allow an options hash to be passed in to Upcoming Invoice API

### LINKED ISSUE #176: Cannot access invoices.retrieveUpcoming using arguments
The documentation says that I must supply a customer Id, with optional arguments to generate a prorated preview. My goal is to see what the end-balance looks like when I change from one subscription to another.

However, I have tried two different argument approaches and both result in a JSON error. 

1. ` stripe.invoices.retrieveUpcoming( customerId, {object of arguments}, callback) `
2. ` stripe.invoices.retrieveUpcoming( {object of arguments}, callback);`

```javascript
{ 
  [Error: Invalid JSON received from the Stripe API]
  type: 'StripeAPIError',
  stack: 'Error: Invalid JSON received from the Stripe API\n    
    at Error._Error (node_modules/stripe/lib/Error.js:12:17)\n    
    at Error.Constructor (node_modules/stripe/lib/utils.js:113:13)\n    
    at Error.Constructor (node_modules/stripe/lib/utils.js:113:13)\n    
    at IncomingMessage.<anonymous> (node_modules/stripe/lib/StripeResource.js:140:13)\n    
    at IncomingMessage.emit (events.js:117:20)\n    
    at _stream_readable.js:943:16\n   
    at process._tickDomainCallback (node.js:463:13)',
  rawType: undefined,
  code: undefined,
  param: undefined,
  message: 'Invalid JSON received from the Stripe API',
  detail: undefined,
  raw: 
   { 
     message: 'Invalid JSON received from the Stripe API',
     response: '',
     exception: [SyntaxError: Unexpected end of input] 
   } 
}
```

matthewarkin:
The Upcoming Invoice API now supports a ton of cool optional params
to help calculate prorations.

Since the customer ID is still required, and the subscription id is
optional, the second parameter to the retrieveUpcoming endpoint
can be an options object or a string (string must be the subscription
id for legacy compatibility).

Fixes #176

### DIFF
--- a/lib/resources/Invoices.js
+++ b/lib/resources/Invoices.js
@@ -2,6 +2,7 @@
 
 var StripeResource = require('../StripeResource');
 var stripeMethod = StripeResource.method;
+var utils = require("../utils");
 
 module.exports = StripeResource.extend({
 
@@ -24,12 +25,16 @@ module.exports = StripeResource.extend({
     method: 'GET',
     path: function(urlData) {
       var url = 'upcoming?customer=' + urlData.customerId;
-      if (urlData.subscriptionId) {
-        return url + '&subscription=' + urlData.subscriptionId;
+      // Legacy support where second argument is a the subscription id
+      if (urlData.invoiceOptions && typeof urlData.invoiceOptions === 'string') {
+        return url + '&subscription=' + urlData.invoiceOptions;
+      }
+      else if (urlData.invoiceOptions && typeof urlData.invoiceOptions === 'object') {
+        return url + "&" + utils.stringifyRequestData(urlData.invoiceOptions);
       }
       return url;
     },
-    urlParams: ['customerId', 'optional!subscriptionId']
+    urlParams: ['customerId', 'optional!invoiceOptions']
   })
 
 });
--- a/test/resources/Invoices.spec.js
+++ b/test/resources/Invoices.spec.js
@@ -99,6 +99,20 @@ describe('Invoices Resource', function() {
       });
     });
 
+    describe('With a options object in addition to a customer ID', function() {
+      it('Sends the correct request', function() {
+
+        stripe.invoices.retrieveUpcoming('customerId1', { plan:'planId123' });
+        expect(stripe.LAST_REQUEST).to.deep.equal({
+          method: 'GET',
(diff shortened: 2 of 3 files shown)


matthewarkin:
I do hate the forking on typeOf but its necessary for backwards compatibility.

Also I debated whether or not it should be invoice.retreiveUpcoming(options), where options would include the customer id, vs how I implemented it where customer id is a separate parameter.

I went with the separate one since customer id is required and all the other options are optional.

matthewarkin:
@michelle

test/resources/Invoices.spec.js
```diff
@@ -99,6 +99,20 @@ describe('Invoices Resource', function() {
       });
     });
 
+    describe('With a options object in addition to a customer ID', function() {
+      it('Sends the correct request', function() {
+
```
michelle: oof, JSCS would yell at you if you put in these sorts of whitespaces in our JS! :P

michelle: Whoops, ignore that--wrong PR.

lib/resources/Invoices.js
```diff
@@ -24,12 +25,16 @@ module.exports = StripeResource.extend({
     method: 'GET',
     path: function(urlData) {
       var url = 'upcoming?customer=' + urlData.customerId;
-      if (urlData.subscriptionId) {
-        return url + '&subscription=' + urlData.subscriptionId;
+      // Legacy support where second argument is a the subscription id
+      if (urlData.invoiceOptions && typeof urlData.invoiceOptions === "string") {
```
michelle: let's `s/"/'/g` in this file

matthewarkin:
Changed " to '
Also I updated the readme to reflect the change

michelle:
Thanks for the PR!

> I do hate the forking on typeOf but its necessary for backwards compatibility.

Yeah, I'm actually pretty conflicted about this, given we've never documented this optional second parameter, and I'd love for us to be more aggressive with deprecating functionality. Doesn't seem worth a major version bump just deprecate this undocumented feature though, so I'll file an issue to deprecate the subscription ID second parameter for the next major version bump.


> I went with the separate one since customer id is required and all the other options are optional.

Yes, agreed that this is correct. Options has should only include optional parameters.

michelle:
Thanks!

PR #183, opened 2015-06-30
Switch array rendering syntax to brackets

bkrausz:
Fixes a bug where arrays of hashes do not work properly. Also `a[]=first&a[]=second` is more commonly used than `a=first&a=second`.

r? @michelle


bkrausz:
CC @remistr

michelle:
:shipit:

PR #185, opened 2015-07-07
Remove flaky Invoice Item test

kyleconroy:
Now that we're testing on four different node.js versions, this test
fails half the time. It's not a very helpful test anyways, so just
remove it.


SlexAxton:
Seems fine. Any idea on why it's flaky?

kyleconroy:
@SlexAxton It's filtering invoice items by date. There are four concurrent tests happening at the same time, filtering over the same list. Hence, very inconsistent results.

PR #187, opened 2015-07-15
Added support for verify bank account endpoint

jackdent:
r? @raycmorgan

### DIFF
--- a/lib/resources/Customers.js
+++ b/lib/resources/Customers.js
@@ -144,6 +144,12 @@ module.exports = StripeResource.extend({
     urlParams: ['customerId', 'sourceId']
   }),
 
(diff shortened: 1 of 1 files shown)


matthewarkin:
the path is incorrect, and this likely wouldn't be merged until ACH leaves beta

brandonros:
What is somebody looking to integrate with the ACH beta and node supposed to do? Roll their own HTTP requests?

matthewarkin:
Stripe doesn't have API bindings for any of its betas nor even for the Stripe Connect oauth stuff. You could "fork" the node package or you could roll your own http request. With incoming ACH, the only thing that would actually need to be rolled your own is the verify endpoint, the rest of it should just work.

PR #188, opened 2015-07-18
Add top level disputes

kyleconroy:
r? @cupcait (not urgent, can wait til Monday) cc @michelle

### DIFF
--- a/lib/resources/Disputes.js
+++ b/lib/resources/Disputes.js
@@ -1,0 +1,21 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+
+  path: 'disputes',
+
+  includeBasic: [
+    'list', 'retrieve', 'update', 'setMetadata', 'getMetadata'
+  ],
+
+  close: stripeMethod({
+    method: 'POST',
+    path: '/{id}/close',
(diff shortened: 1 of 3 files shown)


lib/resources/Disputes.js
```diff
@@ -0,0 +1,21 @@
...
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+
+  path: 'disputes',
+
+  includeBasic: [
+    'list', 'retrieve', 'update', 'setMetadata', 'getMetadata'
```
cupcait: What about set/get evidence methods?

kyleconroy: Whoops, you're totally right, will need to add those.

kyleconroy: So I've thought more about this, and I think a "setEvidence" method isn't needed, as the following works:

```js
stripe.disputes.update('dp_123', {evidence: {customer_name: 'Bob'}})
```

cupcait: Yep, that sounds in line with how some of the other libraries do it. In that case, can you remove the metadata methods, too?

kyleconroy: I think we should be consistent here: all other resources have the `setMetadata` and `getMetadata` methods, so a dispute object should be no different.

PR #190, opened 2015-07-29
Added the ability to delete an account

shalecraig:
R? @michelle


michelle:
can we add a test?

looks good though!

Issue #194, opened 2015-09-10
Issues with very old user IDs


stefanosala:
Hi folks,
we're on our way to publish our Application that uses Connect.

We're having some troubles testing it with some users because of their user IDs, which are in the form of 32 alpha-numerical characters.

The problem is that they are interpreted as authentication keys and not user ids when doing calls such as `accounts.retrieve`: https://github.com/stripe/stripe-node/blob/master/lib/resources/Accounts.js#L35

Is there a way to force those calls to treat those user ids correctly not as authentication keys?

Thanks!

matthewarkin:
Well I thought I found a workaround but I believe that work around is abusing a bug in the API so I've pinged some folks to investigate that bug and hopefully once that comes back I can think of a proper workaround / fix, or maybe a Stripe will.

I have a very horrible disgusting work around https://github.com/stripe/stripe-node/commit/2473296a5bd38664be375cc93cb15a39831d9485, basically you'd send an object with the key "id" as the first parameter instead of passing a string.

matthewarkin:
@bkrausz any thoughts on a decent way to handle this? 
My initial thought would be to deprecate passing api key as the first argument and require that the first parameter be an account id or an empty string (to access your own account). 

So
```javascript
 // retrieve a connected account
stripe.Accounts.retrieve("account id", "api key" || {options args}, callback)
 // retrieve self  account / account for api key / stripe_account
stripe.Accounts.retrieve("", "api key" || {options args}, callback) 
```

or add a `retrieveSelf` function that would take no urlParam, just the optional options argument and callback. 

Can't say I'm a *huge* fan of either method.

bkrausz:
Sounds like we should check for a 2nd hash-or-string param, which would indicate that the first string is an account ID. Then you could pass `stripe.Accounts.retrieve("account id", {}, callback)`. Does that sound like it would all permutations of issues here?

matthewarkin:
yeah, so that's essentially my first option and if the first param is an empty string / null the library would request `/account`

stefanosala:
Thanks for the feedback guys, please keep us posted as it seems it will be a BC with the current version.

matthewarkin:
Have a PR open that should handle this in the most backwards compatible way that I could thing of but its still a breaking change.

[...]

PR #160, opened 2015-03-20
Rename StripeInvalidRequest to StripeInvalidRequestError

matthewarkin:
For consistency, lets rename all errors to end in Error.

closes #159

PR #196, opened 2015-09-12
Remove ability for API keys to be passed as 1st param to acct.retrieve

matthewarkin:
Old (pre-prefix) account ids and api keys are hard to differentiate. As
such old account ids would pass the `isAuthKey` tried to identify
whether a user was requesting a connected account or if they were trying
to request the account for the passed api key.

This is a "major" change in that an api key cannot be passed as a string
to the first argument. It can be passed in the options parameter as the
`api_key` value, or as a string if the user passes `null` as the account
id.

[stripe/stripe-node#194]

### DIFF
--- a/lib/resources/Accounts.js
+++ b/lib/resources/Accounts.js
@@ -31,17 +31,21 @@ module.exports = StripeResource.extend({
   }),
 
   retrieve: function(id) {
-    // Old bindings supported no-arg / callback access to /v1/account, support that if ID is not passed
-    if (id === undefined || typeof id === 'function' || (typeof id === 'string' && utils.isAuthKey(id))) {
-      return stripeMethod({
-        method: 'GET',
-        path: 'account'
-      }).apply(this, arguments);
-    } else {
+    // No longer allow an api key to be passed as the first string to this function due to ambiguity between
+    // old account ids and api keys. To request the account for an api key, send null as the id
+    if (typeof id === "string") {
       return stripeMethod({
         method: 'GET',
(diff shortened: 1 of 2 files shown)


matthewarkin:
@kyleconroy @bkrausz if either of you (or another stripe) get a chance it would be appreciated. (I know things may be a tad crazy given relay launch

bkrausz:
If we do a major version bump we can consider this a breaking change, which I'm conceptually OK with. We do have a reasonable number of platforms using our node bindings with `/v1/account` requests, but I'm fine breaking that as long as there's a comment.

@kyleconroy: thoughts?

matthewarkin:
Yeah, ideally platforms wanting to access /v1/account would do `stripe.account.retrieve({api_key:"key"})` instead of just passing in a string, so as a breaking change its a fairly easy one to fix.  I definitely get how its not ideal. The other option would be to make allow the first param to be an object that has a key "id" but that seems really more gross and creates a special case of how url parameters are handled.

matthewarkin:
Unrelated #160 would be nice for 4.0 as a bit of naming OCD

kyleconroy:
I'm :+1: to 4.0 and this change, as long as the `{api_key:"key"}` object is support in all methods, and not just some strange one-off.

matthewarkin:
Yeah, the {api_key:"key"} object in place of just the api key was added for 3.1.0 when idempotency support was added. 8bcd1144d74666a910f294ac1870436014635d17

PR #198, opened 2015-09-16
Add HTTP status code to errors.

### LINKED ISSUE #145: Response status not available on error object

kyleconroy:
Fix #145

r? @michelle

### DIFF
--- a/lib/Error.js
+++ b/lib/Error.js
@@ -42,6 +42,7 @@ var StripeError = _Error.StripeError = _Error.extend({
     this.detail = raw.detail;
     this.raw = raw;
     this.requestId = raw.requestId;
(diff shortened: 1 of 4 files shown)


michelle:
shibbit
:shibbit:

PR #203, opened 2015-10-22
Stick to single-quote string literals for consistency

tejasmanohar:
Because convention :D


kyleconroy:
Sure! Why not :smile:

PR #204, opened 2015-10-22
Prefer object literal over switch evaluation

tejasmanohar:
Objects are awesome and more expressive, standard, and functional.

https://medium.com/@tejasmanohar/flip-the-switch-239d2d479a29


kyleconroy:
Don't think I'm going to merge this one. I think the switch statement is fine here, as it's only three cases.

tejasmanohar:
:\ Alright, fair enough. I almost never use switch, but I do see the point about not changing something that's already working with  little benefit :)

PR #207, opened 2015-10-22
Remove lodash when required but not used

tejasmanohar:
may PR a linter... would help with stuff like this + styling consistency and so on


kyleconroy:
@tejasmanohar Do you have any suggestions for a linter we could use? We already do something similar for stripe-python.

tejasmanohar:
@kyleconroy dropped my suggestions in #209

PR #208, opened 2015-10-22
Switch out iojs for node 4 now that it's LTS

tejasmanohar:
Would recommend since IO.js is no longer supported after the merge w/ Joyent's project for Node foundation. Additionally, Node 4 is LTS now so good to have tests running for that


kyleconroy:
I'd agree as well. Thanks for the PR!

PR #214, opened 2015-11-13
Add Stripe exports for easier mocking

utvara:
In order to be able to stub the library sinonjs needs an property to stub:

> https://github.com/sinonjs/sinon/issues/664
> https://github.com/sinonjs/sinon/issues/664#issuecomment-71980876

After that it would be possible to do:

```
var stripe = require('stripe');
var stubStripe = sinon.stub(stripe, 'Stripe').returns(stubStripeClient);
```

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -161,3 +161,4 @@ Stripe.prototype = {
 };
 
 module.exports = Stripe;
+module.exports.Stripe = Stripe;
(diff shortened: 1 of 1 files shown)


brandur:
Cool! Makes sense to me if it's helpful. Would you mind just commenting the added line to explain why it's there beyond the normal exports? Thanks!

utvara:
@brandur added a small comment for reference, not sure if inline example is in order?

kyleconroy:
No need for an inline example. Thanks for the PR!

PR #216, opened 2015-12-01
Add the ability to verify a customer source

wbronitsky:
r? @stripe/api-libraries

### DIFF
--- a/lib/resources/Customers.js
+++ b/lib/resources/Customers.js
@@ -144,6 +144,12 @@ module.exports = StripeResource.extend({
     urlParams: ['customerId', 'sourceId'],
   }),
 
+  verifySource: stripeMethod({
+    method: 'POST',
+    path: '/{customerId}/sources/{sourceId}/verify',
+    urlParams: ['customerId', 'sourceId'],
+  }),
+
   /**
    * Customer: Discount methods
    */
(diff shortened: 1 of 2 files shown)


lib/resources/Customers.js
```diff
@@ -144,6 +144,12 @@ module.exports = StripeResource.extend({
     urlParams: ['customerId', 'sourceId'],
   }),
 
+  validateSource: stripeMethod({
+    method: 'POST',
+    path: '/{customerId}/sources/{sourceId}/validate',
```
brandur: Is this a different URL from your pulls in other languages?

brandur:
Left a question about the API endpoints in use above, but otherwise this looks good to me.

As with PHP, I'll leave this one open a bit longer in case someone else with more Node bindings experience also wants to take a look.

lib/resources/Customers.js
wbronitsky: shouldn't be. The `customers/` is already taken care of here

brandur: Oh I was thinking the `/validate` part at the end (vs. `/verify` elsewhere).

wbronitsky: :facepalm: you are totally right. Thanks!

brandur: @wbronitsky Cool! And I think the discrepancy might have made it into the function naming as well. Maybe consider a rename from `validateSource` to `verifySource`.

brandur:
Released as 4.1.0.

Issue #215, opened 2015-12-01
Thrown error should be in callback


Nijikokun:
Un-documented thrown error kept taking down my application:

https://github.com/stripe/stripe-node/blob/master/lib/StripeMethod.js#L45

brandur:
Hi @Nijikokun, sorry about the trouble here!

So this exception is a usage error that occurs if an API call method is misused by not passing enough arguments or by passing a `null` when we needed a value. Ideally you'll be able to avoid it at runtime completely by having a test suite that's adequate enough to reasonably exercise your code and/or by performing some basic input checks to make sure that you're not passing a `null` inappropriately. Obviously this is easier said than done, but you should be able to get yourself into a place where you never see this error unexpectedly.

I'm going to close this issue out for now, but of course let us know if you have anymore comments/questions on the problem. Thanks!

Nijikokun:
It should at least be documented.

brandur:
> Finding the callback would be as simple as checking for an argument of type function.

@Nijikokun I'm still a little hesitant to pass it back through the standard runtime paths because this is the type of error that indicates that something is very wrong in the user code, and by extension it should be made very difficult to ignore or not handle. If this was a compiled language, this would be the sort of problem that the compiler would not allow through to run time.

By fair enough. I don't have great Node familiarity, so if you have other specific ideas on how to improve things I'd be happy to take a look. When you talk about documentation, do you just mean the function's documentation or something else?

matthewarkin:
@brandur I get why it makes sense that this is a thrown exception, though the Errors api docs doesn't state anywhere that the node.js sdk will throw an error. One thing for me is that the handling for how invalid requests are handled are different based on if the check happens client-side or server-side, so creating a charge with a null card would return an invalid request error in the callback but trying to update a customer with a null id would instead throw an error. Who is to say that we shouldn't have logic in the create charge method that checks that `amount != null && (customer!= null || source != null)`? Both these instances are invalid requests but they're handled differently.

And per Joyent's Node docs (not necessarily saying these are best practices)
>The general rule is that a function may deliver operational errors synchronously (e.g., by throwing) or asynchronously (by passing them to a callback or emitting error on an EventEmitter), but it should not do both. This way, a user can handle errors by either handling them in the callback or using try/catch, but they never need to do both. Which one they use depends on what how the function delivers its errors, and that should be specified with its documentation.

[...]

PR #218, opened 2015-12-14
Return usage errors through callback

brandur:
Return usage errors through callbacks as described in #215. This has the advantage of never throwing an error unexpectedly in user code.

@kyleconroy Would you mind taking a review pass on this? I literally learned how promises work 45 seconds ago. Thanks!

### DIFF
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -46,7 +46,10 @@ function stripeMethod(spec) {
           continue;
         }
 
-        var err = new Error('Stripe: I require argument "' + urlParams[i] + '", but I got: ' + arg);
+        var err = new Error(
+          'Stripe: Argument "' + urlParams[i] + '" required, but got: ' + arg +
+          ' (on API request to ' + requestMethod + ' ' + commandPath + ')'
+        );
         deferred.reject(err);
         return deferred.promise;
       }
@@ -59,8 +62,9 @@ function stripeMethod(spec) {
 
     if (args.length) {
       var err = new Error(
-        'Stripe: Unknown arguments (' + args + '). Did you mean to pass an options object? ' +
-        'See https://github.com/stripe/stripe-node/wiki/Passing-Options.'
+        'Stripe: Unknown arguments (' + args + '). Did you mean to pass an options ' +
+        'object? See https://github.com/stripe/stripe-node/wiki/Passing-Options.' +
+        ' (on API request to ' + requestMethod + ' ' + commandPath + ')'
       );
       deferred.reject(err);
       return deferred.promise;
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -45,7 +45,10 @@ function stripeMethod(spec) {
           urlData[param] = '';
           continue;
         }
-        throw new Error('Stripe: I require argument "' + urlParams[i] + '", but I got: ' + arg);
+
+        var err = new Error('Stripe: I require argument "' + urlParams[i] + '", but I got: ' + arg);
+        deferred.reject(err);
+        return deferred.promise;
       }
 
       urlData[param] = args.shift();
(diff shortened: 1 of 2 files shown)


kyleconroy:
I, too, don't know much about promises.

On a related note, are we sure we want to do this? I've been following #215, and I thought we decided that these errors would be strange to send through the callback, because they're programmer error, not API errors.

brandur:
> On a related note, are we sure we want to do this? I've been following #215, and I thought we decided that these errors would be strange to send through the callback, because they're programmer error, not API errors.

Yeah, I think it's still a bit subjective, but I think that all errors through callbacks may be the best route for the time being.

I agree that it's strange for this type of usage problem to be sent back along an error channel that's mostly used for runtime errors, but it seems that in this case it's too easy to accidentally run into this problem by passing a value that's unexpectedly `null` or `undefined`, which can then lead to a program crash. Matt also [pointed to the Joyent docs](https://www.joyent.com/developers/node/design/errors) which suggest using only exceptions _or_ callback errors, which seems to be as good of a convention as any as far as I'm concerned.

Throwing the exception rather than passing by an error by callback doesn't buy you that much even in testing and development, so I'm not too unhappy to see it go. It's still open for debate though, so if you have any other thoughts/ideas, I'd be happy to hear them.

Regarding review: any chance that @michelle might want to take a look at this to make sure that I haven't committed any unforgivable promise taboo? (And see also #215 for more background.)

kyleconroy:
Just a heads up, but @pifantastic is going to take a look for us.

Since this is the only place in the entire code base where we throw an exception, I'm okay with changing the behavior. This is a breaking change, so we'll need to version in accordingly.

pifantastic:
I'm taking a look, but I'll just say, in general it's not expected to mix asynchronous and synchronous errors, so my opinion is that it's better for these to be wrapped in the promise.

test/resources/Charges.spec.js
```diff
@@ -120,9 +120,9 @@ describe('Charge Resource', function() {
     });
 
     it('Throws an error on incorrect arguments', function() {
```
pifantastic: Nit: Maybe change the description since it doesn't throw anymore? Maybe something like `Incorrect arguments result in an error`.

lib/StripeMethod.js
```diff
@@ -45,7 +45,10 @@ function stripeMethod(spec) {
           urlData[param] = '';
           continue;
         }
-        throw new Error('Stripe: I require argument "' + urlParams[i] + '", but I got: ' + arg);
+
+        var err = new Error('Stripe: I require argument "' + urlParams[i] + '", but I got: ' + arg);
```
pifantastic: Nit: It maybe helpful to include the `path` in the error so it's immediately obvious what method is causing the problem.

pifantastic:
Save a few nits that aren't vital, this LGTM.

test/resources/Charges.spec.js
brandur: +1. Improved the spec name in 6b482f3.

lib/StripeMethod.js
brandur: Good call. More debugging output is rarely unhelpful. Also fixed in 6b482f3.

brandur:
@pifantastic Thanks a lot for taking a look at this!

Regarding breaking change: we're changing the nature of the error return, but in a way that the current version would have led to a broken program anyway. User programs would have already had to handle the error-by-callback case anyway, and now we're just sending a few more things through that channel (but again, only in broken cases). It's true that there may now be useless rescue blocks that might get left around, but IMO that's not too big of a deal.

kyleconroy:
@brandur but a new type of error will now be coming through the callback. If you had code that assumed that error would always be from the API, your code may now break.

brandur:
> @brandur but a new type of error will now be coming through the callback. If you had code that assumed that error would always be from the API, your code may now break.

True, but I a couple things:

1. Because the old path was such a hard crash, there's a good chance that most users have succeeded in curbing its occurrences by this point.
2. We're just sending the top-level `Error` class here. It's certainly with the realm of possibility that you could write a program that conditionally handles only certain types of errors coming back from an error channel, but I suspect that it's very rare in practice.

Anyway, I don't feel too strongly about it. We can bump the major version if we're still worried about it.

kyleconroy:
https://stripe.com/docs/api/node#errors also needs to be updated to remove the caveat about exceptions. I'm pretty sure the sample code in that section needs a default case, as these errors will not have a `type`.

nijikokun:
:+1:

brandur:
> https://stripe.com/docs/api/node#errors also needs to be updated to remove the caveat about exceptions. I'm pretty sure the sample code in that section needs a default case, as these errors will not have a type. 

Yeah, +1. Opened an internal change to address these problems.

Issue #222, opened 2016-01-13
Expose Request-ID header


olalonde:
Exposing the Request-ID header field would be extremely useful (e.g. to prevent double processing response of API call vs webhook events).

olalonde:
I exposed the whole res object instead as a non enumerable property to maintain backwards compatibility and cleaner console.logs.

brandur:
See also #223.

brandur:
This was fixed by #223.

PR #223, opened 2016-01-13
Expose ._res property on resources. Fixes issue #222

olalonde:


brandur:
Nice! Thanks for the patch!

What do you think about calling it something like `lastResponse` to match the naming in the similar PHP functionality (see stripe/stripe-php#206)?

olalonde:
Yes, sure. `lastResponse` or `_lastResponse`?

brandur:
@olalonde I believe that the leading underscore here is a convention being used in place of JavaScript's missing visibility modifiers. Because we're intending to use `lastResponse` from outside of `StripeResource` objects, I'd vote to have no leading underscore.

olalonde:
Ok, done :+1:

brandur:
Great! And thanks a lot for adding the test as well.

brandur:
Released as 4.2.0.

olalonde:
Thanks!

PR #229, opened 2016-02-06
Added support for the CountrySpec endpoint

remistr:
r? @brandur 

The tests are really lacking here but other specs were as empty. Let me know if we should ensure that the data returned is valid.


brandur:
Holy Remi, you are on a mission! Thanks!

brandur:
Released as 4.4.0.

PR #241, opened 2016-04-19
Remove eval from url interpolator

implausible:
makeURLInterpolator requires use of eval to create a function which interpolates url strings in constant time.

By taking a small increase in interpolator complexity to linear time with respect to parts interpolated, we can improve the security of stripe-node. This also helps this package conform to security requirements an electron application might have.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -31,7 +31,6 @@ var utils = module.exports = {
   },
 
   /**
-   * https://gist.github.com/padolsey/6008842
    * Outputs a new function with interpolated object property values.
    * Use like so:
    *   var fn = makeURLInterpolator('some/url/{param1}/{param2}');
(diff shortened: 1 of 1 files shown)


matthewarkin:
To be fair, you most likely do not want to be using stripe-node in an electron application environment. Doing so would mean you're shipping your secret API key as part of the application which ends up being a major security issue. Opening up an electron application package and viewing your source code is pretty easy, so essentially anyone can then steal your API key and create fraudulent charges, perform card testing, or refund all your charges.

brandur:
> To be fair, you most likely do not want to be using stripe-node in an electron application environment. Doing so would mean you're shipping your secret API key as part of the application which ends up being a major security issue

+1.

FWIW, I'm generally supportive of an effort to remove evals whether it's for Electron or not. This is a subjective observation, but I find that the new version of the method is generally quite hard to follow compared to the current implementation, which I think is a drawback that somewhat counteracts the added benefit.

I wonder if there's something we can do here to refactor/clarify the new implementation? Ideally, we'd probably be looking for something that's as legible as the current one.

implausible:
> Opening up an electron application package and viewing your source code is pretty easy, so essentially anyone can then steal your API key and create fraudulent charges, perform card testing, or refund all your charges.
This is a fair assessment. I was doing a mock-up of 'what could be', not 'what will be' when I discovered that the CSP was triggered. Still, I thought it was worth it to refactor.
>  I find that the new version of the method is generally quite hard to follow compared to the current implementation, which I think is a drawback that somewhat counteracts the added benefit.
I was also unsure of how much the current stripe team cares about the runtime efficiency of this function, so I tried to keep it as close to quick as possible. If the current team is in agreement, I could definitely lop off a ton of code and improve the readability by just performing a runtime regex replace. What do you think?

implausible:
I've gone ahead and rewritten it to be as readable as before.

brandur:
@implausible Cool, this works for me. It may be less runtime performant than the other version, but doing a regex find/replace shouldn't be too bad, and I kind of like how the function becomes quite a bit easier to follow (i.e. [doesn't need a dedicated blog post to explain it](http://james.padolsey.com/javascript/straight-up-interpolation/)). Thanks!

PR #248, opened 2016-05-25
Support returning orders

rasmusrygaard:
r? @stripe/api-libraries 
cc @jimdanz

### DIFF
--- a/lib/resources/Orders.js
+++ b/lib/resources/Orders.js
@@ -23,4 +23,10 @@ module.exports = StripeResource.extend({
     urlParams: ['id'],
   }),
 
+  order_return: stripeMethod({
+    method: 'POST',
+    path: '/{id}/returns',
+    urlParams: ['id'],
+  }),
+
 });
--- a/lib/resources/Orders.js
+++ b/lib/resources/Orders.js
@@ -19,14 +19,14 @@ module.exports = StripeResource.extend({
 
   pay: stripeMethod({
     method: 'POST',
-    path: '/{id}/pay',
-    urlParams: ['id'],
+    path: '/{orderId}/pay',
+    urlParams: ['orderId'],
   }),
 
-  order_return: stripeMethod({
+  returnOrder: stripeMethod({
     method: 'POST',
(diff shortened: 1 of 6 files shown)


lib/resources/Orders.js
```diff
@@ -23,4 +23,10 @@ module.exports = StripeResource.extend({
     urlParams: ['id'],
   }),
 
+  order_return: stripeMethod({
```
brandur: I notice a lot of the other binding pulls were flipped compared to this one (i.e. "return order" versus "order return"). Is there a reason

Also, I think the convention in Node (or within this library at least) is to use camel casing for method names.

lib/resources/Orders.js
```diff
@@ -23,4 +23,10 @@ module.exports = StripeResource.extend({
...
+    method: 'POST',
+    path: '/{id}/returns',
+    urlParams: ['id'],
```
brandur: I notice that other resource helpers seem to name parameters with a resource prepended (i.e. `orderId`), and we might want to match that here. See [here](https://github.com/stripe/stripe-node/blob/71e34ef44e80675cfc7c16420233013e6355a7ef/lib/resources/Invoices.js#L12,L16) for example.

brandur:
@rasmusrygaard Left a couple minor comments, but awesome work! Thanks for fixing up the list in the README too.

rasmusrygaard:
Thanks for the comments @brandur. I cleaned up the code to be more consistent and use `returnOrder`!

test/resources/Orders.spec.js
```diff
@@ -102,6 +102,22 @@ describe('Order Resource', function() {
     });
   });
 
+  describe('order_return', function() {
```
brandur: Looks like one last vestigial `order_return` here. Looks good otherwise!

rasmusrygaard: Well that's embarrassing, good catch! Fixed

brandur:
Released as 4.7.0.

PR #254, opened 2016-07-19
Add support for the Source resource

spolu:
This PR adds support for the Source API resource.

### DIFF
--- a/lib/resources/Sources.js
+++ b/lib/resources/Sources.js
@@ -1,8 +1,6 @@
 'use strict';
 
 var StripeResource = require('../StripeResource');
-var stripeMethod = StripeResource.method;
-var utils = require('../utils');
 
 module.exports = StripeResource.extend({
 
--- a/lib/resources/Sources.js
+++ b/lib/resources/Sources.js
@@ -1,0 +1,15 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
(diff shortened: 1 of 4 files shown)


lib/resources/Sources.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+var utils = require('../utils');
```
brandur: I know that imports are a bit of a mess right now, but maybe just for strict correctness, drop both `stripeMethod` and `utils` here; neither are needed.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "stripe",
-  "version": "4.8.0",
+  "version": "4.8.1",
```
brandur: I think by semver we'd bump the minor version rather than patch.

Want to just leave this out of the pull though? I'll take care of it and the rest of the release separately.

brandur:
@spolu Left a couple comments above, but overall looks good! Thanks.

ptal @spolu

spolu:
Thanks @brandur! updated. PTAL

brandur:
LGTM!

brandur:
Released as 4.9.0.

spolu:
Thanks @brandur!

PR #267, opened 2016-09-16
Add SubscriptionItem object

jxu577:
Add SubscriptionItem object to support list, retrieve, create, update, deletion on subscriptions items through `v1/subscription_items`. Note that documentation for this endpoint in the README has yet to be filled out since this feature is still in development.


kyleconroy:
👍

PR #259, opened 2016-08-22
URI-encode all fields of User-Agent

brandur:
URI-encodes all fields of the JSON version of the Stripe User-Agent
field in case something like the system `uname` contains a special
character.

Fixes #246.

r? @dpetrovics Mind taking a look at this?

PR #270, opened 2016-09-26
Remove Lodash dependency

brandur:
Lodash is being used in relatively few places and as described in #259
can cause some problems in transforming objects to versions that cannot
be properly encoded.

Here we remove Lodash. Much of its functionality has been added to ES6
anyway, so we can eventually even remove some of the vendored functions
that we have.

Fixes a report received in #259.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -3,100 +3,18 @@
 var qs = require('qs');
 
 var hasOwn = {}.hasOwnProperty;
-var toString = {}.toString;
+var isPlainObject = require('lodash.isplainobject');
 
 var utils = module.exports = {
-
-  // Note that in ES6 all uses of this can just be replaced with Object.assign.
-  // Given a few more years we should drop this function completely.
-  assign: function(o1, o2) {
-    for (var attr in o2) {
-      o1[attr] = o2[attr];
-    }
-    return o1;
-  },
 
   isAuthKey: function(key) {
     return typeof key == 'string' && /^(?:[a-z]{2}_)?[A-z0-9]{32}$/.test(key);
   },
 
   isOptionsHash: function(o) {
-    return utils.isPlainObject(o) && ['api_key', 'idempotency_key', 'stripe_account'].some(function(key) {
+    return isPlainObject(o) && ['api_key', 'idempotency_key', 'stripe_account'].some(function(key) {
       return o.hasOwnProperty(key);
     });
-  },
-
-  /**
-   * Checks if `value` is a plain object, that is, an object created by the
-   * `Object` constructor or one with a `[[Prototype]]` of `null`.
-   *
-   * Note that this function is vendored from Lodash with some minor
-   * modifications.
-   *
-   * @static
-   * @memberOf utils
-   * @category Lang
-   * @param {*} value The value to check.
-   * @returns {boolean} Returns `true` if `value` is a plain object, else `false`.
-   * @example
-   *
-   * function Foo() {
-   *   this.a = 1;
-   * }
-   *
-   * utils.isPlainObject(new Foo);
-   * // => false
-   *
-   * utils.isPlainObject([1, 2, 3]);
-   * // => false
-   *
-   * utils.isPlainObject({ 'x': 0, 'y': 0 });
-   * // => true
-   *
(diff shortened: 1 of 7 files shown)


brandur:
@floatingLomas @SlexAxton Hey! One thing that we used from Lodash was a function that would allow us to check whether an object was "plain" (i.e. just one with basic key/value pairs) which seems to be useful when checking certain function parameters so that we can ensure that nothing crazy gets passed in. Here I ended up vendoring this method into `utils` because it seems like something that couldn't hurt for us to keep doing. I'll be the first to admit that I'm a bit of a JS novice though, and wouldn't mind a second opinion if either one of you would be willing to give one.

The premise of removing Lodash in the first place is also not a sacred cow. It seems to be causing a few issues (#259) and we have such a minimal need for it that it seems reasonable to me to get rid of it, but let me know if you disagree here.

lib/utils.js
```diff
@@ -1,25 +1,102 @@
...
 var hasOwn = {}.hasOwnProperty;
 var toString = {}.toString;
 
 var utils = module.exports = {
 
+  // Note that in ES6 all uses of this can just be replaced with Object.assign.
+  // Given a few more years we should drop this function completely.
+  assign: function(o1, o2) {
+    for (var attr in o2) {
+      o1[attr] = o2[attr];
```
asessa: if (typeof o2[attr] !== 'function')

asessa:
Hey @brandur, I've found the problem and it's external to stripe-node. In my case Object is extended by the web framework.
As utils.assign is used for headers, I recommend to check also for content type to avoid conflicts.

lib/utils.js
```diff
@@ -1,25 +1,102 @@
...
+    }
+    return o1;
+  },
```
SlexAxton: This is a pretty naive implementation. I'd feel much more comfortable bringing in something like the `object-assign` module to do just this. https://www.npmjs.com/package/object-assign

lib/utils.js
```diff
@@ -1,25 +1,102 @@
...
   },
 
-  isObject: function(o) {
-    return _.isPlainObject(o);
+  /**
+   * Checks if `value` is a plain object, that is, an object created by the
+   * `Object` constructor or one with a `[[Prototype]]` of `null`.
+   *
+   * Note that this function is vendored from Lodash with some minor
+   * modifications.
```
SlexAxton: Seem again like we could just pull in https://www.npmjs.com/package/lodash.isplainobject and be safer.

SlexAxton:
I don't feel suuuuuuuuuper strongly, but I don't think it's harmful to pull in dependencies like this, especially in the node world.

floatingLomas:
Looks kool, but as an alternative, all the lodash methods are individually packaged, so we could just include [isPlainObject](https://www.npmjs.com/package/lodash.isplainobject) on its own.

brandur:
@SlexAxton @floatingLomas Nice. Thanks a lot for the quick responses guys!

I think I've integrated all your feedback. I pulled out `lodash` in favor of `lodash.isplainobject` (which I didn't know about before — nice), and I've also put in the `object-assign` ponyfill for now.

It looks like @asessa's original problem may have been unrelated to lodash, but I feel like overall this change is still for the best: we get a path to a complete ES6 implementation instead of the relatively specialized `extend`, and we also just decrease the total dependency surface area.

r? @floatingLomas Would you mind taking a more complete review pass here if you have some time? Thanks!

asessa:
Hey @brandur, latest commit using object-assign also solves my problem.

brandur:
> Hey @brandur, latest commit using object-assign also solves my problem.

Excellent to hear! Thanks for following up.

floatingLomas:
lgtm

I'm not 100% certain on the differences in behavior (if any) between `for (var k in data)` vs `_.forEach()`, but it's all just swapping ~equivalent functions, so I'm pretty sure it's cool.

SlexAxton:
The only difference is the scope that the function runs in, which isn't an issue here. Also we're filtering on `hasOwnProperty` so it's all pretty safe/good.

floatingLomas:
Awesomo. :+1:

brandur:
@floatingLomas @SlexAxton Thanks again guys!

PR #271, opened 2016-09-27
Add tests for items encoding

kyleconroy:


lib/StripeMethod.js
```diff
@@ -22,6 +22,7 @@ function stripeMethod(spec) {
                   : utils.makeURLInterpolator(spec.path || '');
   var requestMethod = (spec.method || 'GET').toUpperCase();
   var urlParams = spec.urlParams || [];
+  var encode = spec.encode || function(data) {return data;};
```
brandur: Mind documenting this with the other parameters above?

lib/resources/Subscriptions.js
```diff
@@ -3,15 +3,37 @@
 var StripeResource = require('../StripeResource');
 var stripeMethod = StripeResource.method;
 
+function encode(data) {
+  if (Array.isArray(data.items)) {
+    var items = {};
+    data.items.map(function(item, i) {
```
brandur: Can we put this in `utils` under `arrayToObject` a little like in Ruby? (and with a test)

lib/StripeMethod.js
kyleconroy: Fixed

lib/resources/Subscriptions.js
kyleconroy: Fixed

kyleconroy:
@brandur can you take a second look?

brandur:
@kyleconroy Thanks!

LGTM.

PR #272, opened 2016-10-14
Support 403 status codes

jflinter:


brandur:
Similar to PHP, can we call this one `StripePermissionsError` for consistency? Thanks!

jflinter:
re-r? @brandur

Issue #290, opened 2016-12-26
Bug of card


jet-snowman:
Hi there! I have a problem when I tried to create card for some customer. I also tried example on index page and got the same error. 

This is the example:

```node
stripe.customers.create({
  email: 'foo-customer@example.com'
}).then(function(customer){
  return stripe.customers.createSource(customer.id, {
    object: 'card',
    exp_month: 10,
    exp_year: 2018,
    number: '4242 4242 4242 4242',
    cvc: 100
  });
}).then(function(source) {
  return stripe.charges.create({
    amount: 1600,
    currency: 'usd',
    customer: source.customer
  });
}).then(function(charge) {
  // New charge created on a new customer
}).catch(function(err) {
  // Deal with an error
});
```

Response:
```
{ [Error: Missing required param: source.]
  type: 'StripeInvalidRequestError',
  rawType: 'invalid_request_error',
  code: undefined,
  param: 'source',
  message: 'Missing required param: source.',
  detail: undefined,
  raw:
   { type: 'invalid_request_error',
     message: 'Missing required param: source.',
     param: 'source',
     statusCode: 400,
     requestId: 'req_9ofKRcFXZEvl2X' },
  requestId: 'req_9ofKRcFXZEvl2X',
  statusCode: 400 }
```

I've read the documentation and the field source is required but in example and even in test files I don't see this field.

Stripe: 4.14.0
Node js: 5.11
npm: 3.8.6

sosnovsky:
I have the same problem :(

fred-stripe:
Hi @jet-snowman @sosnovsky, what you need to do is pass the `source` as a dictionary as similar to the Stripe API example for creating a token:

https://stripe.com/docs/api/node#create_card_token

```
stripe.customers.create({
  email: 'customer@example.com'
}).then(function(customer){
  return stripe.customers.createSource(customer.id, {
    source: {
        object: 'card',
        exp_month: 10,
        exp_year: 2018,
        number: '4242 4242 4242 4242',
        cvc: 100
     }
  });
}).then(…)
```

See also:
- https://stripe.com/docs/api/node#create_token
- https://stripe.com/docs/api/node#create_card


Unfortunately the example in the README on this repository is incorrect, and Stripe's API example shows how to create a card from a token, not directly from a card number. You should generally be using Stripe.js to tokenize card numbers instead of sending them to the API directly:

sosnovsky:
@fred-stripe thanks for solution!

[...]

PR #291, opened 2016-12-28
Fix example in README.md

fred-stripe:
Source properties should be under a top-level `source` key in the create source request. See confusion in #290.

r? @brandur


dpetrovics-stripe:
Thanks for this, @fred-stripe!

LGTM

PR #305, opened 2017-02-14
Ignore `tags` file for NPM, fixes #302

### LINKED ISSUE #302: npm package has an extra 4.6 MB file
Hi, do you need that ctags-generated 4.6 MB `tags` file in package root, or could it be included in either `.gitignore` or `.npmignore` to reduce npm package size by 50%.

(did `npm install stripe@4.15`)

fred-stripe:
Dug into this issue, and it looks like NPM defaults to "include everything" even if `tags` would be excluded, for example, by a global `.gitignore` rule.

https://docs.npmjs.com/misc/developers

r? @brandur-stripe

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .npm-debug.log
+tags


brandur-stripe:
Good call! Thanks again for tracking this down :)

LGTM.

PR #309, opened 2017-03-08
Update qs

### LINKED ISSUE #308: High Severity Vulnerability in "qs" - Prototype Override Protection Bypass
Thanks for this useful package!

Could you update your 'qs' dependency to 6.4.0 in order to avoid a high severity vulnerability.

Many thanks!

References:
https://snyk.io/vuln/npm:qs:20170213
https://www.sourceclear.com/registry/security/prototype-override-protection-bypass/javascript/sid-3638

brandur-stripe:
There appears to be an open security vulnerability on earlier versions:

https://snyk.io/vuln/npm:qs:20170213

Fixes #308.

This may cause a problem in that it looks like from their changelog they dropped support for Node 4 already ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,14 +30,14 @@
     "chai-as-promised": "~4.1.1",
     "jscs": "^2.3.5",
     "mocha": "~2.1.0",
-    "qs": "~6.0.2",
+    "qs": "~6.0.4",
     "stripe-javascript-style": "^1.0.1"
   },
   "dependencies": {
     "bluebird": "^2.10.2",
     "lodash.isplainobject": "^4.0.6",
     "object-assign": "^4.1.0",
-    "qs": "~6.0.2"
+    "qs": "~6.0.4"
   },
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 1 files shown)


gastonelhordoy:
Hey @brandur-stripe, if old node versions should be considered, you can use qs@v6.0.4. The fix was backported to the following list of releases: 6.3.2, 6.2.3, 6.1.2, 6.0.4.

In the [6.0.x branch](https://github.com/ljharb/qs/blob/master/CHANGELOG.md#602) they seem to support node down to v0.8.

brandur-stripe:
@gastonelhordoy Excellent find. Thank you!

It seems that the build actually passed on the lastest version (maybe because we're using a very limited subset of qs?), but I've moved back to `~6.0.2` anyway because it seems less risky.

brandur-stripe:
And by `~6.0.2` I mean `~6.0.4` because that's where the actual fix came in. Oops.

brandur-stripe:
Alright we're having some trouble with Travis, but given the success of the previous build, I'm fairly confident in this one going through. I'm going to pull this in and cut a release.

PR #310, opened 2017-03-11
Added proxy configuration to README.md

perevernihata:
There is no information about how to use node sdk behind the proxy, tho it was implemented a while ago.  Please consider adding this to the readme, might save someone an hour or two


brandur-stripe:
Thanks! I'm going to rework this a little in master, but this looks like a solid addition.

PR #312, opened 2017-03-13
Refresh README

brandur-stripe:
Refreshes the README to do some general cleanup (I removed the "Version
2" update notice from 2013 and the section on the original author for
example) and to just rebuild the structure to look more like stripe-ruby
(and which will hopefully long term be a standard for all our READMEs).
For example:

* Documentation is linked first thing on top.
* Succinct installation and basic usage sections.
* All "special topics" like promises or using a Proxy get their own h3
  section under "Usage".
* "Development" section stays out of the way at the bottom.

One notable thing: I've started to eliminate the use of the word
"binding" in preference of "SDK". I spoke to @romain-stripe a bit about
this one, and we both agreed that although we're all used to "binding"
internally, more widely it's a little unusual of a word to use.

[Here's a link to the new rendered README](https://github.com/stripe/stripe-node/blob/brandur-readme-update/README.md).

r? @romain-stripe Putting your name on this, but if anyone else has feedback
just drop it in here. I'm going to merge this relatively soon, but will word in any
other commentary async.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Install the package with:
 ## Usage
 
 The package needs to be configured with your account's secret key which is
-available in your [Stripe Dashboard][dashboard]. Require it with the key's
+available in your [Stripe Dashboard][api-keys]. Require it with the key's
 value:
 
 ``` js
@@ -149,7 +149,7 @@ $ export STRIPE_TEST_API_KEY='sk_test....'
 $ npm test
 ```
 
+[api-keys]: https://dashboard.stripe.com/account/apikeys
 [connect]: https://stripe.com/connect
-[dashboard]: https://dashboard.stripe.com/account
 [https-proxy-agent]: https://github.com/TooTallNate/node-https-proxy-agent
 [stripe-js]: https://stripe.com/docs/stripe.js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,17 +141,15 @@ Run a single test (case sensitive):
...
 
-*Note: On Windows use `SET` instead of `export` for setting the `STRIPE_TEST_API_KEY` environment variable.*
-
-If you don't have a prefixed key (in the form `sk_test_...`) you can get one by rolling your "Test Secret Key". This can be done under your dashboard (*Account Setting -> API Keys -> Click the roll icon next to the "test secret key"*). This should give you a new prefixed key ('sk_test_..'), which will then be usable by the node mocha tests.
-
-## Author
-
-Originally by [Ask Bjørn Hansen](https://github.com/abh) (ask@develooper.com). Development was sponsored by YellowBot. Now officially maintained by Stripe.
+[connect]: https://stripe.com/connect
+[dashboard]: https://dashboard.stripe.com/account
```
romain-stripe: If we want we can use this direct link to API keys: https://dashboard.stripe.com/account/apikeys

romain-stripe:
Fantastic work, @brandur-stripe! I’m super glad we’re improving this README! I think my preference would go for _library_ instead of _SDK_ (both are better than _binding_!), but other than that I’m big +1! 👌

README.md
brandur-stripe: Great point! Pushed an edit to take care of that.

brandur-stripe:
Thanks Romain! All requests have been addressed. Pulling this in.

PR #318, opened 2017-03-26
Added support for the new Payout and RecipientTransfer objects

remi-stripe:
The Transfer object used to represent all movements of funds in Stripe. It split in three resources:
- Transfer: this describes the movement of funds between Stripe accounts and is specific to Stripe Connect.
- Payout: this describes the movement of funds from a Stripe account to a bank account, debit card or any future payout method.
- RecipientTransfer: this describes the movement of funds from a Stripe account to a Recipient's card or Bank Account. This is here for legacy reasons and can only be accessed from an expanded BalanceTransaction.

This change is behind an API version so old API versions would still use the Transfer object for everything while new API version would see the split.

This applies beyond the new object as some properties/methods are removed from Transfer and other properties are renamed on other objects.

### DIFF
--- a/lib/resources/Payouts.js
+++ b/lib/resources/Payouts.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+
+  path: 'payouts',
+
+  includeBasic: [
+    'create', 'list', 'retrieve', 'update',
+    'setMetadata', 'getMetadata',
+  ],
+
+  cancel: stripeMethod({
+    method: 'POST',
+    path: '{payoutId}/cancel',
+    urlParams: ['payoutId'],
+  }),
+
+  listTransactions: stripeMethod({
+    method: 'GET',
+    path: '{payoutId}/transactions',
+    urlParams: ['payoutId'],
+  }),
+});
+
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -37,6 +37,7 @@ var resources = {
   Events: require('./resources/Events'),
   Invoices: require('./resources/Invoices'),
   InvoiceItems: require('./resources/InvoiceItems'),
+  Payouts: require('./resources/Payouts'),
   Plans: require('./resources/Plans'),
   RecipientCards: require('./resources/RecipientCards'),
   Recipients: require('./resources/Recipients'),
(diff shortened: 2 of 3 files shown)


brandur-stripe:
You rock Remi. Let me know the schedule of starting to get these pulled in when you have it.

LGTM.

remi-stripe:
@brandur-stripe Added support for `RecipientTransfer`. Similar to Python, I couldn't figure out how to test/confirm this one unfortunately.

lib/resources/RecipientTransfers.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+// This resource is only used when expanding a balance tranasaction
+// it can not be listed, retrieved or updated
```
brandur-stripe: Thank you for the comment here!

brandur-stripe: I don't think we need this file though. Resources should just be coming back as plain old objects:

```
$ npm run mocha debug -- test/stripe.spec.js --grep 'Will expose'

> stripe@4.16.1 mocha /Users/brandur/stripe/stripe-node
> mocha "debug" "test/stripe.spec.js" "--grep" "Will expose"

< Debugger listening on [::]:5858
connecting to 127.0.0.1:5858 ... ok
break in node_modules/mocha/bin/_mocha:7
  5  */
  6
> 7 var program = require('commander')
  8   , path = require('path')
  9   , fs = require('fs')
cont


<   Stripe Module
<     Callback support
<       Any given endpoint
break in test/stripe.spec.js:99
  97             },
  98           }, function(err, customer) {
> 99           debugger;
 100             cleanup.deleteCustomer(customer.id);
 101
debug> repl
Press Ctrl + C to leave debug repl
> customre
ReferenceError: customre is not defined
> customer
{ id: 'cus_ANyfLDmaBoZSJW',
  object: 'customer',
  account_balance: 0,
  created: 1490916606,
  currency: null,
  default_source: 'card_ANyfWchTFHrEWG',
  ...
```

Note how other subresource-only models are not defined (or as far as I can tell anyway).

brandur-stripe:
ptal @remi-stripe

remi-stripe:
I removed the `RecipientTransfer` resource. That explains why we never added one for `AlipayAccount` either.

PTAL @brandur-stripe

brandur-stripe:
Thanks Remi!

LGTM.

brandur-stripe:
Released as 4.17.0.

PR #320, opened 2017-03-28
Add section "Examining Responses"

brandur-stripe:
Adds a section that includes some basic information on looking at
response information.

Requested on #315.

r? @ob-stripe 

@stripe/api-libraries

PR #321, opened 2017-03-29
Switch API docs link from `/python` to `/node`

jlomas-stripe:
Re my comment on #320 

r? @brandur-stripe


brandur-stripe:
Oops, yes this is much better :) Thanks for the additional proofread @jlomas-stripe!

PR #328, opened 2017-04-12
Check to make sure the `id` is a String

jlomas-stripe:
If you accidentally do `stripe.charges.capture(chargeObject)` instead of `stripe.charges.capture(chargeObject.id)`, you currently get this error back from Stripe:

```
{ 
"type": "invalid_request_error", 
"message": "No such charge: [object Object]", 
"param": "id", 
"statusCode": 404, 
"requestId": "req_ABC123xyz" 
}
```

It would be better if you got something like this before the API call was even made:

```
Unhandled rejection Error: Stripe: "id" must be a string, but got: object (on API request to `POST /charges/{id}/refund`)
```

So ... that's what this does.  ¯\\\_(ツ)_/¯

### DIFF
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -44,6 +44,16 @@ function stripeMethod(spec) {
         var isOptional = OPTIONAL_REGEX.test(param);
         param = param.replace(OPTIONAL_REGEX, '');
 
+        if (param == 'id' && typeof arg !== 'string') {
+          var path = this.createResourcePathWithSymbols(spec.path);
+          var err = new Error(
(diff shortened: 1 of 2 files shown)


jlomas-stripe:
r? @brandur-stripe 

PS: Only Node and Golang take this approach, but I'm pretty sure Golang doesn't have this problem.

_something something dynamically typed_

brandur-stripe:
Nice! Seems like a pretty unambiguous UX improvement.

LGTM.

brandur-stripe:
Released as 4.18.0.

PR #330, opened 2017-05-09
Add webhook signing

jlomas-stripe:
\o/

As the Node library doesn't have 'classes' and 'instances' of things, I'm not 100% sold that `constructEvent()` is the right name for the function here - though that's what it's set to for consistency.

I'm happy to discuss any pieces/parts of this, so let me know!

r? @ob-stripe 
cc @brandur-stripe

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,57 +132,7 @@ charge.lastResponse.statusCode
 
 Stripe can optionally sign the webhook events it sends to your endpoint, allowing you to validate that they were not sent by a third-party.  You can read more about it [here](https://stripe.com/docs/webhooks#signatures).
 
-Here's an example of how to use it with Express:
-
-```js
-const Express = require('express');
-const bodyParser = require('body-parser')
-
-const Stripe = require('stripe')('sk_test_...');
-
-const webhookSecret = 'whsec_...';
-
-// We use a router here so we can use a separate `bodyParser` instance that
-// adds the raw POST data to the `request`
-const router = Express.Router();
-
-router.use(bodyParser.json({
-  verify: function(request, response, buffer) {
-    request.rawBody = buffer.toString();
-  },
-}));
-
-router.post('/webhooks', function(request, response) {
-  var event;
-
-  try {
-    // Try adding the Event as `request.event`
-    event = stripe.webhooks.constructEvent(
-      request.rawBody,
-      request.headers['stripe-signature'],
-      webhookSecret
-    );
-  } catch (e) {
-    // If `constructEvent` throws an error, respond with the message and return.
-    console.log('Error', e.message);
-
-    return response.status(400).send('Webhook Error:' + e.message);
-  }
-
-  console.log('Success', event.id);
-
-  // Event was 'constructed', so we can respond with a 200 OK
-  response.status(200).send('Signed Webhook Received: ' + event.id);
-});
-
-// You could either create this app, or just return the `Router` for use in an
-// existing Express app - up to you!
-const app = Express();
-app.use(router);
-app.listen(3000, function() {
-  console.log('Example app listening on port 3000!')
-});
-```
+You can find an example of how to use this with [Express](https://expressjs.com/) in the [`examples/webhook-signing`](examples/webhook-signing) folder.
 
 ## More Information
 
--- a/README.md
+++ b/README.md
@@ -128,6 +128,62 @@ charge.lastResponse.requestId // see: https://stripe.com/docs/api/node#request_i
 charge.lastResponse.statusCode
 ```
 
+### Webhook signing
+
+Stripe can optionally sign the webhook events it sends to your endpoint, allowing you to validate that they were not sent by a third-party.  You can read more about it [here](https://stripe.com/docs/webhooks#signatures).
+
+Here's an example of how to use it with Express:
+
+```js
+const Express = require('express');
+const bodyParser = require('body-parser')
+
+const Stripe = require('stripe')('sk_test_...');
+
+const webhookSecret = 'whsec_...';
+
+// We use a router here so we can use a separate `bodyParser` instance that
+// adds the raw POST data to the `request`
+const router = Express.Router();
+
+router.use(bodyParser.json({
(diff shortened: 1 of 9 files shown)


jlomas-stripe:
OMG IT FINALLY PASSED :joy:

README.md
```diff
@@ -128,6 +128,62 @@ charge.lastResponse.requestId // see: https://stripe.com/docs/api/node#request_i
...
+
+// You could either create this app, or just return the `Router` for use in an
+// existing Express app - up to you!
+const app = Express();
+app.use(router);
+app.listen(3000, function() {
+  console.log('Example app listening on port 3000!')
+});
+```
+
```
brandur-stripe: Thanks for updating the README :)

This is minor, but I think that in order to keep our READMEs succinct we should probably be aiming to provide code snippets in the range of 5 to 10 lines and moving full examples to an `examples/` directory in the same spirit [as stripe-python](https://github.com/stripe/stripe-python/tree/master/examples).

brandur-stripe:
Will leave full review to OB, but this is sweet!

ob-stripe:
Awesome! Thanks a lot for this @jlomas-stripe.

Overall, this lgtm, but I'm not sure if we should use a try/catch block here. I realize it's a bit silly to use a promise for a synchronous, purely client-side operation, but maybe for consistency's sake we should make the interface look like this:

```js
stripe.webhooks.constructEvent({
  payload: request.rawBody,
  sigHeader: request.headers['stripe-signature'],
  secret: webhookSecret
}, function(err, event) {
  // "asynchronously" called
});
```

wdyt @jlomas-stripe @brandur-stripe? I don't have nearly enough Node experience to decide what is better.

README.md
jlomas-stripe: I totally went back and forth on this exact thing.  In fact, I created and then deleted an `examples/` folder - at least once. :)

I'll create a `examples/webhook-signing` (or maybe `examples/webhook-signing/express` because `hapijs` / `sailsjs` / etc) folder and put this file and the appropriate `package.json` in there; I can actually make this a runnable example that way.

jlomas-stripe:
@ob That is a good point.  Here are my thoughts on this:

- It's not asynchronous code, so I don't think [pretending it is makes sense](http://stackoverflow.com/questions/30159316/q-promisify-synchronous-operations-for-chaining-promises);
- Our current [error handling approach](https://github.com/stripe/stripe-node/wiki/Error-Handling) is "we only throw for bad arguments; otherwise, check the first arg of the callback";  in this case, the possible failures are all a result of a bad argument, so I think the `try`/`catch` approach/logic still stands;

The Webhook Signing exists is to save an (async) API call to verify payload; it seems to me that a synchronous function here is the right solution - and I believe this still fits in with our existing 'Error Handling' documentation.  That said, it is like two lines of code to add a `constructEventAsPromise()` method using Bluebird's `Promise.method`...  ¯\_(ツ)_/¯

I'm happy hear @brandur-stripe's thoughts on the matter though. :)

**Also** happy to further discuss the idea of an 'arguments object' vs individual arguments; my thoughts here is that this is a fixed method, with ~fixed arguments, so this approach makes sense - in contrast to the API calls, where there is an object with a variable number of (optional) parameters that are being used for either modifying something or searching for something.

ob-stripe:
That all makes perfect sense, thanks!

Let's wait just in case @brandur-stripe has more to share, but I'm fine with keeping the interface as-is.

README.md
brandur-stripe: Hah nice. I like it.

What do you think about keeping a small bit of code in the README too to show the basic concept? Something like:

``` node
event = stripe.webhooks.constructEvent(
    webhookBody,
    webhookStripeSignatureHeader,
    webhookSecret
);
```

brandur-stripe:
Good points @ob-stripe, especially around maintaining API consistency. I'm +1 what @jlomas-stripe said though, and I don't think we should make a CPU-bound operation look like its asynchronous. There's a fair bit of precedent in the standard library for this kind of stuff; around their crypto package for example:
https://nodejs.org/dist/latest-v7.x/docs/api/crypto.html

README.md
jlomas-stripe: I think that's a great idea. :)

jlomas-stripe:
@brandur-stripe README updated; let me know if this needs any further tweaks, as I'm happy to make them!

README.md
```diff
@@ -128,6 +128,20 @@ charge.lastResponse.requestId // see: https://stripe.com/docs/api/node#request_i
...
 ```
 
+### Webhook signing
+
+Stripe can optionally sign the webhook events it sends to your endpoint, allowing you to validate that they were not sent by a third-party.  You can read more about it [here](https://stripe.com/docs/webhooks#signatures).
+
+You can find an example of how to use this with [Express](https://expressjs.com/) in the [`examples/webhook-signing`](examples/webhook-signing) folder, but here's what it looks like:
+
+```
+event = stripe.webhooks.constructEvent(
```
brandur-stripe: Maybe just a `js` hint to this block. Otherwise looks good.

brandur-stripe:
> @brandur-stripe README updated; let me know if this needs any further tweaks, as I'm happy to make them!

LGTM. If @ob-stripe's happy, then lets get this merged!

README.md
jlomas-stripe: Done.

_something something force-push_

ob-stripe:
🚢

brandur-stripe:
Great work here again! I really like how it uses `crypto.timingSafeEqual` if at all possible and only falls back to an internal timing safe implementation if necessary.

Bringing this in.

brandur-stripe:
Released as 4.19.0.

Would either of you happen to be aware if there's already work in progress to get some sample code for Node added to our docs?

jlomas-stripe:
I don't believe so - I can take a run at it though!

brandur-stripe:
> I don't believe so - I can take a run at it though!

Cool. @ob-stripe may know more, but we should make an effort to get it in there.

PR #333, opened 2017-05-15
Add `setAppInfo()`

jlomas-stripe:
I did a thing.  ¯\\\_(ツ)_/¯

r? @ob-stripe 
cc @brandur-stripe

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,8 +149,8 @@ If you're writing a plugin that uses the library, we'd appreciate it if you iden
 ```js
 stripe.setAppInfo({
   name: MyAwesomePlugin,
-  version: '1.2.34',
-  url: 'https://myawesomeplugin.info',
+  version: '1.2.34', // Optional
+  url: 'https://myawesomeplugin.info', // Optional
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -142,6 +142,20 @@ event = stripe.webhooks.constructEvent(
 );
 ```
 
+### Writing a Plugin
+
+If you're writing a plugin that uses the library, we'd appreciate it if you identified using `stripe.setAppInfo()`:
+
+```js
+stripe.setAppInfo({
+  name: MyAwesomePlugin,
+  version: '1.2.34',
+  url: 'https://myawesomeplugin.info',
+});
+```
+
+This information is passed along when the library makes calls to the Stripe API.
+
 ## More Information
 
  * [REST API Version](https://github.com/stripe/stripe-node/wiki/REST-API-Version)
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -140,6 +140,10 @@ Stripe.prototype = {
       throw new Error('AppInfo must be an object.');
     }
 
+    if (info && !info.name) {
+      throw new Error('AppInfo.name is required');
+    }
+
     info = info || {};
 
     var appInfo = APP_INFO_PROPERTIES.reduce(function(accum, prop) {
@@ -151,10 +155,6 @@ Stripe.prototype = {
 
       return accum;
     }, undefined);
-
-    if (appInfo && Object.keys(appInfo).length > 0 && !appInfo.name) {
-      throw new Error('AppInfo.name is required');
-    }
 
     // Kill the cached UA string because it may no longer be valid
     Stripe.USER_AGENT_SERIALIZED = undefined;
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -21,6 +21,8 @@ Stripe.USER_AGENT = {
 
 Stripe.USER_AGENT_SERIALIZED = null;
 
+var APP_INFO_PROPERTIES = ['name', 'version', 'url'];
+
 var exec = require('child_process').exec;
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -142,6 +142,20 @@ event = stripe.webhooks.constructEvent(
...
+
+```js
+stripe.setAppInfo({
+  name: MyAwesomePlugin,
+  version: '1.2.34',
+  url: 'https://myawesomeplugin.info',
+});
+```
+
+This information is passed along when the library makes calls to the Stripe API.
```
ob-stripe: All good, but can you clarify that `version` and `url` are optional?

test/stripe.spec.js
```diff
@@ -68,6 +68,71 @@ describe('Stripe Module', function() {
...
+        countOfRadishes: 512,
+      });
+      expect(stripe._appInfo).to.eql({
+        name: 'MyAwesomeApp',
+        version: '1.2.345',
+        url: 'https://myawesomeapp.info',
+      });
+    });
+
+    it('should be included in the ClientUserAgent and be adddd to the UserAgent String', function(done) {
```
ob-stripe: Typo: `adddd`

lib/stripe.js
```diff
@@ -173,10 +197,32 @@ Stripe.prototype = {
...
 
+      if (self._appInfo) {
+        userAgent.application = self._appInfo;
+      }
+
       cb(JSON.stringify(userAgent));
     });
   },
 
+  getAppInfoAsString() {
```
ob-stripe: It looks like this is not working on older Node versions. I guess something like `getAppInfoAsString: function() {` would work?

ob-stripe:
Left some comments, ptal @jlomas-stripe

README.md
jlomas-stripe: Absolutely.

test/stripe.spec.js
jlomas-stripe: Yup.  Thanks. :)

lib/stripe.js
jlomas-stripe: WHAT DID I EVEN DO THERE.

Yes, that would not only work, it would also be what I meant to type.  I'm not even sure how what I typed even worked.

ob-stripe:
@jlomas-stripe I think you forgot to stage the README file in your latest commit? :)

jlomas-stripe:
Thanks to the magic of force-push, I have no idea what you're talking about.  ¯\\\_(ツ)_/¯

...But really, I ... forgot to save the file.

ob-stripe:
lgtm!

lib/stripe.js
```diff
@@ -133,6 +135,26 @@ Stripe.prototype = {
...
 
+  setAppInfo: function(info) {
+    if (info && typeof info !== 'object') {
+      throw new Error('AppInfo must be an object.');
+    }
+
+    // Kill the cached UA string because it may no longer be valid
+    Stripe.USER_AGENT_SERIALIZED = undefined;
+
+    this._appInfo = APP_INFO_PROPERTIES.reduce(function(accum, prop) {
```
brandur-stripe: I like the format checking here. Is there something that we can do to make `name` required? I believe that's what we're doing in most of the other bindings.

brandur-stripe:
I left one minor comment, but looks great. Thanks for tackling it!

lib/stripe.js
jlomas-stripe: @brandur-stripe Do you mean ensuring that if it contains _something_, then it must contain _at least_ `name`?

brandur-stripe: I was thinking _at least `name`_.

Here's the stripe-ruby method signature:

``` ruby
def self.set_app_info(name, version: nil, url: nil)
```

As you can see we've tried to force presence on `name` and kept `version` and `url` as additional flavoring if you're so inclined.

jlomas-stripe: Awesomo.  Thanks.  I'm on it.

lib/stripe.js
```diff
@@ -133,6 +135,33 @@ Stripe.prototype = {
...
+      if (typeof info[prop] == 'string') {
+        accum = accum || {};
+
+        accum[prop] = info[prop];
+      }
+
+      return accum;
+    }, undefined);
+
+    if (appInfo && Object.keys(appInfo).length > 0 && !appInfo['name']) {
```
jlomas-stripe: @brandur-stripe That look better?

brandur-stripe: Nice! What do you think about tweaking this a bit so that it's simply "must include name"? It seems that calling `setAppInfo` with an empty object is fairly counterproductive and we may as well error on it.

brandur-stripe:
I talked to @jlomas-stripe a little out of band on this one. Looks great to me, but I think we might make one further tweak here to just make `name` always mandatory (i.e. even an empty object is not allowed).

brandur-stripe:
Oops, I didn't see the commit bump here.

LGTM.

brandur-stripe:
Released as 4.20.0.

PR #334, opened 2017-05-16
Check if object property isPrototype before adding it to Constructor.

bendiy:
Some JavaScript environments incorrectly expose `prototype` in the `Object.hasOwnProperty()` method. The `protoExtend()` function should handle this.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -5,6 +5,7 @@ var crypto = require('crypto');
 
 var hasOwn = {}.hasOwnProperty;
 var isPlainObject = require('lodash.isplainobject');
+var objectAssign = require('object-assign');
 
 var utils = module.exports = {
 
@@ -97,17 +98,11 @@ var utils = module.exports = {
     var Constructor = hasOwn.call(sub, 'constructor') ? sub.constructor : function() {
       Super.apply(this, arguments);
     };
+
+    objectAssign(Constructor, Super);
     Constructor.prototype = Object.create(Super.prototype);
-    for (var i in sub) {
-      if (hasOwn.call(sub, i)) {
-        Constructor.prototype[i] = sub[i];
-      }
-    }
-    for (i in Super) {
-      if (i !== 'prototype' && hasOwn.call(Super, i)) {
-        Constructor[i] = Super[i];
-      }
-    }
+    objectAssign(Constructor.prototype, sub);
+
     return Constructor;
   },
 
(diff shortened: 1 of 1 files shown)


brandur-stripe:
@bendiy It looks like the patch isn't working quite as expected. There's some more information on Travis.

Also, can you clarify a little more about context/impetus? Is this causing a problem for you right now? If so, what's your setup like? I ask because if this was a problem seen with any kind of frequency, I would've expected it to have come up a long time ago.

bendiy:
@brahn-stripe I fixed the issue with the original commit. I also made a second commit that uses `objectAssign` for `protoExtend`, which I think is a better solution.

> Also, can you clarify a little more about context/impetus? Is this causing a problem for you right now? If so, what's your setup like?

It's kind of a long story, but I'm using the `stripe-node` library with a different JavaScript engine than Node.js. That [engine](https://doc.qt.io/qt-5/qtscript-index.html) is old and based off of a much earlier WebKit engine around 2009. It has a bug with `prototype` in the `Object.hasOwnProperty()`. Because of this, the original code:

```js
    for (i in Super) {		
      if (hasOwn.call(Super, i)) {		
        Constructor[i] = Super[i];		
      }		
    }
```
would include the `prototype` property in the `for` loop and wipe out the `Constructor.prototype` that the previous `for` loop set using `sub`.

This PR fixes `protoExtend` to support that environment by building `Constructor` in a way that preserves the correct `Constructor.prototype`.

brahn-stripe:
cc: @brandur-stripe :-)

bendiy:
oops :)

brandur-stripe:
Thanks for the context here @bendiy!

Alright, I was a little afraid that this would change behavior because we're no longer running through a `hasOwnProperty` check, but it seems from [the `Object.assign` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) that only own enumerable properties get assigned, which seems to match the current behavior:

> The Object.assign() method is used to copy the values of all enumerable own properties from one or more source objects to a target object. It will return the target object.

[Here's a pretty good reference doc on ownership and enumerability](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties) for anyone else like me who doesn't understand this stuff all that well.

brandur-stripe:
Released as 4.19.1.

PR #339, opened 2017-05-24
Make response headers accessible on errors

### LINKED ISSUE #335: Include lastResponse in error object
Hi

If a stripe error occurs (e.g. StripeCardError), it would be useful to be able to access the headers within the error object - particularly the date header.

Would it be possible for the error object to include either lastResponse or lastReponse.headers please?

Thanks very much.

Robin

brandur-stripe:
It can't hurt to have these accessible when handling an error. We
already have a few other pieces of response information like request ID
and status code, so it's fairly conventional.

Fixes #335.

r? @jlomas-stripe

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -140,9 +140,7 @@ StripeResource.prototype = {
           if (response.error) {
             var err;
 
-            // These are now available on the top-level resource's
-            // lastResponse, but we keep them here too for backwards
-            // compatibility.
+            response.error.headers = headers;
             response.error.statusCode = res.statusCode;
             response.error.requestId = res.requestId;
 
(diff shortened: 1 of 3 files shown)


lib/StripeResource.js
```diff
@@ -140,9 +140,7 @@ StripeResource.prototype = {
           if (response.error) {
             var err;
 
-            // These are now available on the top-level resource's
```
brandur-stripe: I removed this comment because on a second read I realized that it's not accurate :/ `response.error` is all that we sent into error initialization, so these properties must be on there.

jlomas-stripe:
👍 

lgtm

jlomas-stripe:
lgtm

...Did that need to be on the first line? 😂

brandur-stripe:
Thanks @jlomas-stripe! I think CI Bot is chronically broken on the public organization :/

jlomas-stripe:
😭

brandur-stripe:
Released as 4.22.0.

PR #344, opened 2017-06-06
Fix two small typos in Readme

peterkhayes:
* In the *Writing a Plugin* section, `MyAwesomePlugin` should be a string, not a variable name.
* In the *Configuring for Connect* section, the `stripe_account` property should use single quotes, as all other strings in the examples do.


brandur-stripe:
Oops, yep. Thank you!

PR #347, opened 2017-06-08
Add support for ephemeral keys

grey-stripe:
This adds the `EphemeralKey` resource to the bindings.  Currently, the only supported operations are creation and deletion.

Creation is a bit of a special case, in that it requires `Stripe-Version` to be set explicitly (because this will be requested by the backend for use in the frontend, which may be expecting an arbitrary API version).  This means that it is an error if the header is set by the globally configured API version, and it must be provided explicitly in the `create` call.  As such, `create` will pass on an `Error` if `stripe_version` is not specified.

Because no other resources validate their arguments like this, I added a new field to the `StripeMethod` spec: `validator`, which receives the `data` and `options` to validate before the API is called, any error will be passed along to the callback/promise chain.


lib/utils.js
```diff
@@ -85,6 +85,9 @@ var utils = module.exports = {
         if (params.stripe_account) {
           opts.headers['Stripe-Account'] = params.stripe_account;
         }
+        if (params.stripe_version) {
+          opts.headers['Stripe-Version'] = params.stripe_version;
+        }
```
brandur-stripe: Can you add a test case for this to `test/utils.spec.js`? There's some similar ones in there already.

lib/StripeMethod.js
```diff
@@ -87,6 +87,17 @@ function stripeMethod(spec) {
...
+
+      if (spec.validator) {
+        try {
+          spec.validator(data, options);
+        } catch (err) {
+          reject(err);
+          return;
+        }
+      }
+
```
brandur-stripe: So this seems okay, but I'm always a little wary of breaking out a new global pattern for a new resource, especially when we've already got dozens of resources in here that didn't need it up until now. Have you considered the trade off of just allowing the API to return an error in case `Stripe-Version` wasn't specified in the way that it would if you forgot say, a required parameter? Agreed generally that client-side validation is better where possible for faster feedback, but it seems like a way that we could make ephemeral keys less special cased and avoid new general purpose infrastructure.

lib/utils.js
```diff
@@ -14,7 +14,7 @@ var utils = module.exports = {
   },
 
   isOptionsHash: function(o) {
-    return isPlainObject(o) && ['api_key', 'idempotency_key', 'stripe_account'].some(function(key) {
+    return isPlainObject(o) && ['api_key', 'idempotency_key', 'stripe_account', 'stripe_version'].some(function(key) {
```
brandur-stripe: Nice! And wow, pretty surprised to see this wasn't in here already.

test/resources/EphemeralKeys.spec.js
```diff
@@ -0,0 +1,78 @@
...
+          headers: {
+            'Stripe-Version': '2017-06-05',
+          },
+        });
+      });
+    };
+
+    describe('when an api version is set', function() {
+      beforeEach(function() {
+        stripe.setApiVersion('2017-05-25');
```
brandur-stripe: Similar as to stripe-python, can we just persist whatever global value was already set, and reset it after finishing?

test/resources/EphemeralKeys.spec.js
```diff
@@ -0,0 +1,78 @@
...
+        data: {
+          customer: 'cus_123',
+        },
+        headers: {
+          'Stripe-Version': '2017-05-25',
+        },
+      });
+    });
+
+    function versionTests() {
```
brandur-stripe: Wow, this works!?

Seems a little unusual to wrap up `it` cases like this :) Do you think instead we could just have two helper functions for `testErrorsOnNoStripeVersion` and `sendsStripeVersion` and just invoke them from within a few duplicated `it` cases? It's a tiny bit more boilerplate, but seems less weird, and I think it'll play nicer with the test framework in case someone wants to run one individual test case.

lib/StripeMethod.js
grey-stripe: So, unfortunately there is no tradeoff there without more severe changes to how we do things.  The API server can't tell the difference between `Stripe-Version` passed explicitly to `stripe.ephemeralKeys.create` versus the `Stripe-Version` assigned to `stripe.setApiVersion`.  The second should be an error, as the version has to come from the frontend.

brandur-stripe: Ah, yes I see. Good point.

PR #348, opened 2017-06-09
Remove `body-parser` from the Express webhook signing example

### LINKED ISSUE #341: Webhook signature verification and bodyParser.json issue
If you use in your express app `bodyParser.json()` for all routes, and then have a dedicated route for stripe's webhook, then the second call to `bodyParser.json({verify:  ...})` as done in the example has no effect.

I have the following code:
```
app.use(bodyParser.json());
...
app.use('/stripeUrl', stripeRoute);
app.use('/user', userRoute);
....
```

The **StripeRoute** is defined exactly as the example. 

Unfortunately it doesn't work. Only if I comment out the first call to `bodyParser.json` does it work. Which mean that I cannot have a call to `bodyParser` just for my route, I need to add the `verify` function my top call to `bodyParser.json` and to avoid having to pay for the conversion all the time check for the Url to match my stripe route.

Maybe the doc should reflect that, and if it is supposed to work then some guidance as it doesn't for me.

Thanks!

floatingLomas:
Fixes #341 where `body-parser` was conflicting with other instances of `body-parser` in the app.  Replaced it with a middleware function that accomplishes the same thing.

### DIFF
--- a/examples/webhook-signing/express.js
+++ b/examples/webhook-signing/express.js
@@ -2,7 +2,6 @@
 
 const Stripe = require('stripe');
 const Express = require('express');
-const bodyParser = require('body-parser')
 
 /**
  * You'll need to make sure this is externally accessible.  ngrok (https://ngrok.com/)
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Works for me. Thanks for taking care of this @floatingLomas!

PR #357, opened 2017-06-23
Add utils.arrayToObject() to retrieveUpcoming() for `subscription_items`

jlomas-stripe:
Given the following: 

```
stripe.invoices.retrieveUpcoming(customerId, {
    subscription: subscriptionId,
    subscription_items: [
      {
        id: itemToDelete,
        deleted: true
      },
      {
        plan: 'something-1'
      },
      {
        plan: 'something-2'
      }
    ],
    subscription_prorate: true
  });
```

The result would only contain one new plan - specifically, `something-2`, because the API can't find the boundary there.

This PR makes use of the same thing that was [used in `Subscription.update`](https://github.com/stripe/stripe-node/blob/bd4bd44975d4427aad60a488e959f10677f396e9/lib/resources/Subscriptions.js#L28) to make sure the arrays get converted into indexed objects.


brandur-stripe:
Yep, LGTM. Thank you!

brandur-stripe:
Released as 4.23.1.

PR #358, opened 2017-06-27
Add parameters when calling `pay` on an invoice

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @ark-stripe

Users can now provide a `source` parameter when paying invoices. The Node library actually already supports this, I just updated the test to make sure that sending parameters is supported.


brandur-stripe:
Great test addition :) Thanks!

PR #365, opened 2017-08-02
Handle Buffer.from TypeError on node versions prior to 4.5.x

cxam:
The tests are currently failing on node v4 and versions prior to v4.5.x due to a mismatch in the **Buffer.from** API. This was fixed in https://github.com/nodejs/node/pull/7562 and backported to v4.5.x. PR fixes the current use of the function by handling the **TypeError** being thrown.

Alternative solutions:
- Exclude the failing node versions in package.json
- Use semver checking to use the correct API instead of handling the exception

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -162,5 +162,18 @@ function asBuffer(thing) {
     return thing;
   }
 
-  return Buffer.from ? Buffer.from(thing) : new Buffer(thing);
+  if (Buffer.from) {
+    // Buffer.from fix for node versions prior to 4.5.x
+    try {
+      return Buffer.from(thing);
+    } catch (e) {
+      if (e instanceof TypeError) {
+        return new Buffer(thing);
+      } else {
+        throw e;
+      }
+    }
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -162,5 +162,18 @@ function asBuffer(thing) {
     return thing;
   }
 
-  return Buffer.from ? Buffer.from(thing) : new Buffer(thing);
+  if (Buffer.from) {
+    // Buffer.from fix for node versions prior to 4.5.x
+    try {
+      return Buffer.from(thing);
+    } catch (e) {
+      if (e instanceof TypeError) {
```
brandur-stripe: Thanks for trying to minimize the scope of errors that we'll try to catch.

brandur-stripe:
Thanks @cxam! This works for me.

brandur-stripe:
Released as 4.23.2.

cxam:
That's great @brahn-stripe, thank you.

PR #371, opened 2017-08-22
Explicitly specify utf8 encoding when calculating signature

onevcat:
This is a pull-request for a supporting ticket titled as "Signatures failing when using UTF-8 in charge". (Sorry I am not sure where to find the ticket number)

In some Ubuntu environments, the `payload` might follows another encoding and even setting the request encoding to 'utf8' will not affect the received data. Explicitly setting the encoding to 'utf8' will ensure that Stripe side and user side using the same encoding when computing signature.

### DIFF
--- a/lib/Webhooks.js
+++ b/lib/Webhooks.js
@@ -20,7 +20,7 @@ var signature = {
 
   _computeSignature: function(payload, secret) {
     return crypto.createHmac('sha256', secret)
-      .update(payload)
+      .update(payload, 'utf8')
       .digest('hex');
   },
 
--- a/test/flows.spec.js
+++ b/test/flows.spec.js
@@ -432,7 +432,7 @@ describe('Flows', function() {
     });
 
     it('should emit a `request` event to listeners on request', function(done) {
-      var apiVersion = '2017-06-05';
+      var apiVersion = '2017-08-15';
       var idempotencyKey = Math.random().toString(36).slice(2);
 
       function onRequest(request) {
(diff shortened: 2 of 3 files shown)


jlomas-stripe:
Hi @onevcat!  Thanks for submitting this!   I've fixed the tests in #374 so if you can either wait until that lands and rebase master (or just rebase on that branch right now) so that this PR only includes the one specific change, that would be awesome.

@brandur-stripe should definitely take a peek, but since we're (presumably) always going to send a `utf-8` encoded body, hard-coding this where you've got it seems like a reasonable solution to me. 

My only concern here is 'will this break folks who _don't_ need this hardcode override', but I'm pretty sure the answer is 'no' given that it passes all the tests, and 'it works on my machine.'  ¯\\\_(ツ)_/¯

onevcat:
@jlomas-stripe Sure. I will open another p-r for it.

And yes, it might break those ones do not need 'utf8', but I wonder is there a situation they want to use a different encoding? Possibly no. 

Another way might be adding an encoding option in the constructEvent API, so we could keep current implementation untouched while give the user (like me) a chance to get rid of struggling with the encoding settings on the server.

Thanks!

onevcat:
Another p-r is opened in #376 , in which only the utf8 changes is contained.

ob-stripe:
@jlomas-stripe @brandur-stripe FYI, I made a similar change in the Java library a few weeks back: https://github.com/stripe/stripe-java/pull/394.

brandur-stripe:
Thanks @ob-stripe!

PR #374, opened 2017-08-22
Fix request headers in tests

jlomas-stripe:
_My tests were bad and I should feel bad._

I was sending `api_version` in the header, instead of `stripe_version`, so it ... wasn't actually setting them.

Couldn't understand why the versions needed updating in #371 but after investigating I realized it's because I wasn't _actually_ specifying them, and `latest` no longer mapped to the value I was using.

r? @brandur-stripe

### DIFF
--- a/test/flows.spec.js
+++ b/test/flows.spec.js
@@ -456,7 +456,7 @@ describe('Flows', function() {
         currency: 'usd',
         card: 'tok_chargeDeclined',
       }, {
-        api_version: apiVersion,
+        stripe_version: apiVersion,
         idempotency_key: idempotencyKey,
         stripe_account: connectedAccountId,
       }).then(null, function() {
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Haha, I guess this is one the downfall of unit tests — you're testing what you already expect to occur :) Happens to all of us though.

How can you just change this without updating any mainline code though? Has the build been broken this whole time, or are the tests not actually checking anything?

jlomas-stripe:
It broke when Stripe released `2017-08-15`, so I think it's simply a result of nothing having triggered a build since before `2017-08-15` was released.

That said, the tests were 'magically working' because `latest` happened to equal the hard-coded version I'd thought I was passing through (but was not).

brandur-stripe:
Nice! Okay, I think I misunderstood what we were changing initially, but totally makes sense now. Thanks!

PR #376, opened 2017-08-24
Use utf8 to update HMAC payload

onevcat:
This is a follow for #371


onevcat:
The failing tests should be related to #374

brandur-stripe:
Sorry for the delay. This looks very reasonable to me!

I'm quite sure that you're right about the failing test, so I'm just going to merge this in and keep an eye on CI.

brandur-stripe:
Released as 4.24.1. Thanks again!

PR #378, opened 2017-09-05
Test more versions

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Travis now supports Node.js 7 and 8, so we should test against those versions as well.


brandur-stripe:
Nice! LGTM.

PR #379, opened 2017-09-05
Use Bearer auth

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Use Bearer auth instead of Basic auth to authenticate with the API.

I don't have much experience working with Node so it's possible what I've written isn't very idiomatic.

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -214,6 +214,31 @@ StripeResource.prototype = {
     }
   },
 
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Nice one OB! Thanks!

Yeah, I'm not sure I can write purely idiomatic Node either, but I think this is pretty good.

brandur-stripe:
Released as 4.25.0.

PR #380, opened 2017-09-07
Prevent package-lock.json files from being created

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

npm 5 creates a `package-lock.json` and recommends commiting it to the the repo. When present, the lock file is used instead of `package.json`.

This makes sense for apps, but not for libraries like stripe-node. This PR adds a `.npmrc` config file so that npm will not create `package-lock.json` files.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock = false


brandur-stripe:
I'm not super familiar with Node best practices, but your explanation seems sound to me. +1.

ob-stripe:
Same here, but I figure we have `Gemfile.lock` in the [`.gitignore`](https://github.com/stripe/stripe-ruby/blob/master/.gitignore#L4) file for stripe-ruby for basically the same reason, and this seems slightly cleaner.

It's also the approach chosen by eslint (https://github.com/eslint/eslint/issues/8742).

brandur-stripe:
Ah, thanks for the extra information. Alright, that sounds good to me. Pulling this in!

Issue #367, opened 2017-08-03
`jscs` ded.  Can we replace it?


jlomas-stripe:
Per [this](https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.3qrw9jwnz), jscs is no more.  And ... I end up with a 'JavaScript heap out of memory' error every time I try to run it locally...

Can we move to ESLint?

brandur-stripe:
Yep, make sense to me.

jlomas-stripe:
\o/

Ok, I will take a run at it.  Thanks! :)

jlomas-stripe:
Welp, I don't think that's going to work.  [stripe/eslint-config-stripe](https://github.com/stripe/eslint-config-stripe) is a thing, but it seems to be all ES6y and meant for / coupled with React - i.e., includes some React/JSX-specific dependencies, which definitely don't need to be included in this repo just to get ESLint working.

It would be good to move this repo towards using ESLint, but I'm not sure what the best strategy is here - thoughts?

AdriVanHoudt:
@jlomas-stripe try `eslint --init` ;)

deontologician-stripe:
Looks like @ob-stripe is working on this in #383

ob-stripe:
Closed via #383.

PR #382, opened 2017-09-12
Drop support for Node < 4

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

As discussed internally, we should drop support for Node 0.x. The last 0.x branch (v0.12.x) reached EOL on 2016-12-31 [0], and per #367 we need to switch the linter to ESLint which only supports Node >= 4.

This is a breaking change and will require a major version bump, as happened last time we dropped support for an older Node version (stripe-node 3.0.0 [1]).

[0] https://github.com/nodejs/LTS#release-schedule1
[1] https://github.com/stripe/stripe-node/blob/master/CHANGELOG.md


brandur-stripe:
LGTM.

brandur-stripe:
Released as 5.0.0.

PR #383, opened 2017-09-12
Use ESLint instead of jscs

ob-stripe:
This PR is an attempt at fixing #367. It replaces jscs with ESLint.

The main problem is that we were using a custom ruleset for jscs (https://github.com/stripe/javascript-style), which I think is going to be difficult to port to ESLint.

What I did in this PR is use `eslint --init` to create a new configuration file from scratch. ESLint can scan existing code to write a configuration file that more or less matches the style in the current code.

After doing so, there were still ~70 linting errors which I fixed manually. Most of them were legitimate errors (unused variables, unnecessary semicolons), a few were style errors but they were sensible so I fixed them as well.

I'm not sure if this is how we want to handle the switch to ESLint. I'm not a fan of the very long custom configuration file. I also tried using one of the 3 standard rulesets offered by ESLint (Javascript, Google and Airbnb) but all of them resulted in thousands of linting errors.

Flagging as WIP so we can discuss what's the best way forward.

cc @brandur-stripe @jlomas-stripe

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,292 @@
+module.exports = {
+    "env": {
+        "node": true
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "accessor-pairs": "error",
+        "array-bracket-spacing": [
+            "error",
+            "never"
+        ],
+        "array-callback-return": "off",
+        "arrow-body-style": "error",
+        "arrow-parens": "error",
+        "arrow-spacing": "error",
+        "block-scoped-var": "off",
+        "block-spacing": "off",
+        "brace-style": [
+            "error",
+            "1tbs",
+            {
+                "allowSingleLine": true
+            }
+        ],
+        "callback-return": "error",
+        "camelcase": [
+            "error",
+            {
+                "properties": "never"
+            }
+        ],
+        "capitalized-comments": "off",
+        "class-methods-use-this": "error",
+        "comma-dangle": "off",
+        "comma-spacing": "off",
+        "comma-style": [
+            "error",
+            "last"
+        ],
+        "complexity": "error",
+        "computed-property-spacing": [
+            "error",
+            "never"
+        ],
+        "consistent-return": "off",
+        "consistent-this": "off",
+        "curly": "error",
+        "default-case": "off",
+        "dot-location": [
+            "error",
+            "property"
+        ],
+        "dot-notation": "error",
+        "eol-last": "error",
+        "eqeqeq": "off",
+        "func-call-spacing": "error",
+        "func-name-matching": "error",
+        "func-names": "off",
+        "func-style": [
+            "error",
+            "declaration"
+        ],
+        "generator-star-spacing": "error",
+        "global-require": "off",
+        "guard-for-in": "off",
+        "handle-callback-err": "off",
+        "id-blacklist": "error",
+        "id-length": "off",
+        "id-match": "error",
+        "indent": "off",
+        "init-declarations": "off",
+        "jsx-quotes": "error",
+        "key-spacing": "error",
+        "keyword-spacing": [
+            "error",
+            {
+                "after": true,
+                "before": true
+            }
+        ],
+        "line-comment-position": "off",
+        "linebreak-style": [
+            "error",
+            "unix"
+        ],
+        "lines-around-comment": "error",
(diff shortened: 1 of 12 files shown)


brandur-stripe:
@ob-stripe Wow, that's unfortunate! Any chance it comes with any autofixer that we could use in the same style as Rubocop?

If not, it may be easiest just to accept the custom configuration and see if we could switch to a more common convention as part of another phase.

ob-stripe:
@brandur-stripe There is, actually!

I started another branch in [`ob-eslint-standard`](https://github.com/stripe/stripe-node/compare/ob-eslint-standard) using the "Standard" linting rules, then ran `eslint --fix` to autocorrect errors. There were still a few things that I needed to fix manually, though nothing too painful (mostly handling some ignored errors).

The diff is pretty painful to read because those rules remove the semicolon at the end of each line, but I quite like the result.

If you'd like, I can do the same thing for the "Google" and "Airbnb" rules to compare the results.

brandur-stripe:
> I started another branch in ob-eslint-standard using the "Standard" linting rules, then ran eslint --fix to autocorrect errors. There were still a few things that I needed to fix manually, though nothing too painful (mostly handling some ignored errors).
>
> The diff is pretty painful to read because those rules remove the semicolon at the end of each line, but I quite like the result.

Nice! This looks fine to me, but admittedly I don't care enough about the language to have developed any passionate beliefs.

I was a little surprised to hear that the standard ruleset eliminated semicolons though! I don't mind reading code without them, but I'd always thought that the true JS disciples loved them. It might be worth running this by at least a couple people with a stronger stance on this stuff. Maybe @jlomas-stripe or @alex-stripe.

ob-stripe:
I've played around for a bit with Airbnb's style rules. I think it would require a lot more effort to port our codebase to be compliant with their style guide, but it might be worth it: it enforces a lot of nice ES6 features (arrow functions, string interpolation, etc.) and seems to be pretty popular in the JS community.

I pushed the result of `eslint --fix` in [`ob-eslint-airbnb`](https://github.com/stripe/stripe-node/compare/ob-eslint-airbnb). Note that there are still 366 issues that would need to be manually fixed.

brandur-stripe:
> I've played around for a bit with Airbnb's style rules. I think it would require a lot more effort to port our codebase to be compliant with their style guide, but it might be worth it: it enforces a lot of nice ES6 features (arrow functions, string interpolation, etc.) and seems to be pretty popular in the JS community.

From what I can see, this one looks pretty great. Like you said, the arrow functions and string interpolation are really nice. Another one that jumped out at me is the use of `const`/`let` — given the unscoped nature of `var`, this just seems like one of those things that so _obviously_ better because it's likely to keep you more bug-free.

jlomas-stripe:
We do have a thing (https://github.com/stripe/eslint-config-stripe) for ESLint but it's got a bunch of React stuff intermixed; if we could peel that React-specific stuff out of there (into an eslint-config-stripe-react-whatever) that instead references/includes the basic ESLint spec, we could probably then use that basic spec here.

¯\\\_(ツ)_/¯

SlexAxton:
Yea, I'm simultaneously in favor of something that quickly moves us over to
using eslint and also in favor of a fast-follow that enforces more or less
the same rules as the dashboard. So if we think this commit keeps us
roughly in the same spot as before, but updates our toolchain, then got for
it.

jlomas-stripe:
@SlexAxton I think in this case it would be a matter of splitting stripe/eslint-config-stripe into a not-React-y repo and a React-y repo, and then having `stripe-node` make use of the not-React-y one; I briefly looked at pulling it in, but ... there were _so many dependencies_ it scared me off.

Is that what you meant?  I.e., 'yes' to splitting `stripe/eslint-config-stripe`?

ob-stripe:
@brandur-stripe @jlomas-stripe Okay, I think we should go ahead and merge this PR. It keeps us more or less in the same place as before but upgrades the toolchain to use ESLint.

Using the `eslint-config-stripe` config will require a bigger effort as @jlomas-stripe mentioned (i.e. we'd need to split the config to remove all the frontend/React stuff before we could use it here), but we can tackle that later.

brandur-stripe:
Thanks for shaving this yak Olivier. Looks good to me as well.

PR #384, opened 2017-09-12
Bump dependencies versions

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

While working on #383, I noticed that npm was complaining about some deprecated packages, so I thought I'd just bump all dependencies to their latest versions. Should probably have done this at the same time as #382, sorry!

I had to modify some tests to account for a breaking change in chai 4.x (cf. https://github.com/chaijs/chai/issues/781).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,16 +26,16 @@
   },
   "main": "lib/stripe.js",
   "devDependencies": {
-    "chai": "~1.10.0",
-    "chai-as-promised": "~4.1.1",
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Frankly surprised this went so smoothly :) The test suite in stripe-node is pretty good, so +1.

It looks like another one of your open PRs collided with `package.json`. Want to rebase?

ob-stripe:
Done!

PR #385, opened 2017-09-12
Remove object-assign package

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

The [`object-assign`](https://www.npmjs.com/package/object-assign) package is only useful for code targeting Node < 4. Since this is no longer the case for stripe-node, we can safely get rid of it and call `Object.assign` directly.


brandur-stripe:
So much better!

LGTM.

PR #386, opened 2017-09-12
Use test token rather than raw test PAN

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Updates the sample request in the README file to use the `tok_visa` test token rather than the raw test PAN.


brandur-stripe:
Nice catch on this one!

LGTM.

PR #392, opened 2017-09-15
Minimum requirement is Node 4

mantoni:


brandur-stripe:
Ah yep, that's right. Thank you!

PR #394, opened 2017-10-03
Fix the comparison for the args/options warnings

### LINKED ISSUE #393: Bad warning when retrieving plan for connected account
Stripe-node version: 5.1.0;
Node version: 8.6.0 (Oddly, I only discovered this on upgrading from node 6.x.x to 8.x.x)

Behavior works as expected for retrieving a subscription plan on a connected account; however, a warning is logged.  I believe this warning is wrong.

``` 
stripe
.plans
.retrieve('foo-sub-plan', { stripe_account: 'foo-stripe-account' })
.then(plan => console.log(plan))
```
logs the following:
```
Stripe: Options found in arguments (stripe_account). Did you mean to pass an options object? See https://github.com/stripe/stripe-node/wiki/Passing-Options.

{
  id: 'foo-sub-plan'
  ....plan-details
}
```
Since, I'm calling `plans.retrieve` the way the documentation suggests, and since it's behaving as I would expect, I assume that's a bad warning.

Anyway, confused me a bit as I was debugging a separate, but subscription-plans related issue.

Best,

jlomas-stripe:
Fixes #393.  The initial change (#306) didn't properly handle the case where the object was all `options` - it was `console.warn`ing in that case - but now it handles it properly.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -70,7 +70,10 @@ var utils = module.exports = {
       return OPTIONS_KEYS.indexOf(key) > -1;
     });
 
-    if (optionKeysInArgs.length > 0) {
+    // We need to be sure that it's not _both_ an options and args object:
+    // If the args keys and options keys are not the same length,
+    // then it contains both arguments and options, and that's not good.
+    if (argKeys.length !== optionKeysInArgs.length) {
       console.warn(
         'Stripe: Options found in arguments (' + optionKeysInArgs.join(', ') + '). Did you mean to pass an options ' +
         'object? See https://github.com/stripe/stripe-node/wiki/Passing-Options.'
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -70,10 +70,11 @@ var utils = module.exports = {
       return OPTIONS_KEYS.indexOf(key) > -1;
     });
 
-    // We need to be sure that it's not _both_ an options and args object:
-    // If the args keys and options keys are not the same length,
-    // then it contains both arguments and options, and that's not good.
-    if (argKeys.length !== optionKeysInArgs.length) {
+    // In some cases options may be the provided as the first argument.
+    // Here we're detecting a case where there are two distinct arguments
+    // (the first being args and the second options) and with known
+    // option keys in the first so that we can warn the user about it.
+    if (optionKeysInArgs.length > 0 && optionKeysInArgs.length !== argKeys.length) {
       console.warn(
         'Stripe: Options found in arguments (' + optionKeysInArgs.join(', ') + '). Did you mean to pass an options ' +
         'object? See https://github.com/stripe/stripe-node/wiki/Passing-Options.'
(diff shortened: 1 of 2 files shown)


jlomas-stripe:
Oops. :) 

r? @brandur-stripe

lib/utils.js
```diff
@@ -70,7 +70,10 @@ var utils = module.exports = {
       return OPTIONS_KEYS.indexOf(key) > -1;
     });
 
-    if (optionKeysInArgs.length > 0) {
+    // We need to be sure that it's not _both_ an options and args object:
+    // If the args keys and options keys are not the same length,
+    // then it contains both arguments and options, and that's not good.
+    if (argKeys.length !== optionKeysInArgs.length) {
```
brandur-stripe: Sorry @jlomas-stripe, I think I'm a bit thick, but I've been looking at this for a couple minutes now and still don't quite understand why this condition is written like this.

Is the idea that we're for the presence of `optionKeysInArgs` _unless_ args and options are the same object? If it's something like that, is there anyway that we could rewrite this to be a little more easily readable (even if we have to make it a bit longer by making it two separate conditions)? I think it might help a bit for people to wrap their head around the logic when they see it.

jlomas-stripe: We're not checking for the presence of `options` keys, we're checking for a _difference_ between the number of `options` keys and the number of `args` keys.

If all the `args` keys are `options`, then this is fine and it's an `options` object so we can ignore it; if there's some of both, then we need to warn.

I.e., if we have 3 `argKeys` but only 2 `optionKeysInArgs` then there's 1 non-`options` key in there - i.e., an argument - so we need to warn.  If we have 3 and 3, then this _is_ an `options` object, so it's fine and we'll return `{}` here and we don't need to warn.

Does that help?

jlomas-stripe: More explicitly, this previously tested for "are there options keys in here?" but that turned out to be wrong since this can be an `options` object - which obviously has options keys. :)

...Maybe this would be more clear?

```js
if (optionsKeysInArgs.length > 0 && argKeys.length !== optionsKeysInArgs) {}
```

...?

brandur-stripe: Cool, thanks for the explanation!

Yeah, actually would you mind expanding to the two conditions you have above? Also, can we write a more elaborate comment to hopefully help with clarity? Maybe something like this:

> In some cases options may be the provided as the first argument. Here we're detecting a case where there are two distinct arguments (the first being args and the second options) and with known option keys in the first so that we can warn the user about it.

jlomas-stripe: That's a way better comment. :)

Done and done. :)

brandur-stripe:
Nice work as usual @jlomas-stripe! I'll cut a release.

jlomas-stripe:
Awesome!  Thanks for helping make it more clear! :)

I did just add another PR (#395) that we might want to sneak into the release, assuming it makes sense.

PR #395, opened 2017-10-04
More explicitly call out that `constructEvent` needs the raw body

jlomas-stripe:
Since this is something that comes up a lot, I thought it might be worth calling it out a bit more explicitly, both in the Readme and in the Error we return when the signatures don't match.

r? @brandur-stripe


brandur-stripe:
Oops. I did the release before seeing this, but we can always do another one.

Can you check the build failure? I think there's a check on line length that you might need to fix.

jlomas-stripe:
No worries - this is one of those 'whenever' things, I think.

...I still can't get `jscs` to run on my local machine so I ... rely on Travis to do my linting.  And then I forget to ask Travis how it went. 😢

PR #397, opened 2017-10-10
Remove old CHANGELOG

brandur:
We initially had this in place so that old links would still have
somewhere to point to, but I think it's been long enough at this point
(we've shipped *many* releases with the proper link since then). Worst
case scenario is that people just go up to the repository and find
`CHANGELOG.md` manually.


brandur-stripe:
Thanks Remi!

PR #398, opened 2017-10-10
Linkify changelog

jlomas-stripe:
This way we can link into specific versions as needed.

(`sed` for the win \o/ )


brandur-stripe:
Good call!

You may have seen this already, but [we're working on coming up with a standardized changelog format over in stripe-dot](https://github.com/stripe/stripe-dotnet/pull/1026). Very similar to what you're proposing here in that we're trying to have linkable versions and whatnot.

jlomas-stripe:
I actually hadn't seen that, but it ... may have been inspired by someone in that thread. :joy:

PR #399, opened 2017-10-19
Tarek salah bank account support

myvcc657:
`https://github.com/ myvcc /dexteros/shopifyFire.git` 
`perbaiki` kesalahan


ob-stripe:
Hi @myvcc657, thanks for the contribution and sorry for the late reply!

You should already be able to manage bank accounts for customers using the existing `source` methods, e.g.:

```node
stripe.customers.createSource(
  "cus_...",
  { source: "btok_..." },
  function(err, card) {
    // asynchronously called
  }
);
```

or, if you don't want to use tokens:

```node
stripe.customers.createSource(
  "cus_...",
  { source: {
    object: "bank_account",
    account_number: "110000000",
    routing_number: "000123456789",
    country: "US",
    currency: "usd",
  },
  function(err, card) {
    // asynchronously called
  }
);
```

This is documented in the [ACH Guide](https://stripe.com/docs/ach) as well as in the [API reference](https://stripe.com/docs/api/node#customer_create_bank_account).

Is there any specific reason why you the existing methods would not fit your use case?

ob-stripe:
Closing this PR as it's likely unneeded (see my previous comment).

PR #400, opened 2017-10-20
Reformat the changelog for uniformity

brandur:
Reformats the changelog a little bit so that it looks identical to the format we're using in four other Stripe languages now.

[Link to a rendered version of the changelog](https://github.com/stripe/stripe-node/blob/brandur-reformat-changelog/CHANGELOG.md).


brandur-stripe:
Again, this one is fairly minor so I'm pulling it in, but let me know if there are any doubts.

PR #415, opened 2017-12-08
Update docs to take advantage of Node 8's `await`

mikermcneil:
Now that Node 8 is LTS, it might be nice to show `await` syntax first  (here's some [reasoning](https://gist.github.com/mikermcneil/c1028d000cc0cc8bce995a2a82b29245) on that)


brandur-stripe:
Yeah that makes sense. Thanks!

PR #426, opened 2018-02-13
Update subscriptions tests to pass with new API changes

brandur:
Version 2018-02-05 came with a major change to plans and products:

* Products now have a `type` which is `service` or `good`.
* Plans have `name` change to `nickname`.
* Plans now take a product (of `type=service`) on creation.

Testing on `master` is currently broken. This patch updates some test suites to
allow for the new API changes and fixes the suite.

r? @ob-stripe

### DIFF
--- a/test/flows.spec.js
+++ b/test/flows.spec.js
@@ -43,7 +43,10 @@ describe('Flows', function() {
             amount: 1700,
             currency: CURRENCY,
             interval: 'month',
-            name: 'Gold Super Amazing Tier',
+            nickname: 'Gold Super Amazing Tier',
+            product: {
+              name: 'product' + testUtils.getRandomString(),
+            },
           }),
           stripe.customers.create(CUSTOMER_DETAILS)
         ).then(function(j) {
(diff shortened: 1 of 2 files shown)


test/flows.spec.js
```diff
@@ -38,12 +38,18 @@ describe('Flows', function() {
...
+            name: 'product' + testUtils.getRandomString(),
+            type: 'service',
+          }).then(function(product) {
+            return stripe.plans.create({
+              id: 'plan' + testUtils.getRandomString(),
+              amount: 1700,
+              currency: CURRENCY,
+              interval: 'month',
+              nickname: 'Gold Super Amazing Tier',
+              product: product.id,
```
ob-stripe: You can create the plan and product in a single API request, by passing a dictionary for the [`product`](https://stripe.com/docs/api#create_plan-product) parameter. In the interest of making the test suite faster (until we move to stripe-mock 🤞 ), maybe that'd be preferable?

brandur-stripe: Ah, excellent point! Updated.

ptal @ob-stripe

ob-stripe:
👍  LGTM

PR #433, opened 2018-03-13
Add TypeScript usage to README

elliotaplant:
The ES modules usage fails to compile in TS, so this should help TS users get started.


brandur-stripe:
Works for me. Thanks!

Issue #436, opened 2018-03-22
BUG - Path must be a string. Received undefined


reisel:
Node version 8.9.4
Stripe version: 5.5.0

when creating a plan I get the error.
the reason is that the create spec has no path.
then when calling createResourcePathWithSymbols(spec.path) from StripeMethod
createResourcePathWithSymbols is called with undefined.

inside createResourcePathWithSymbols the method calls path.join with the undefined value and this throws the exception.

ob-stripe:
Hi @reisel, thanks for the report. Can you provide more information about this issue? If you could provide sample code to reproduce the issue, it would be very helpful.

reisel:
@ob-stripe 
sure.
```javascript
class StripeChargeRequest {
	constructor({ amount, currency, source }) {
		this.amount = amount
		this.currency = currency
		this.source = source
	}
	
	toJSON() {
		return {
			amount: this.amount,
			currency: this.currency,
			source: this.source
		}
	}
}
const stripe = require('stripe')('sk_test_xxx')
const request = new StripeChargeRequest({ amount: 1200, currency: 'USD', source: 'tok_visa' })
stripe.plans.create(request)
```

please note,
I know the stripe library wasn't designed to support the custom class.
but still in this case there is an error message that need to be displayed:

``` javascript
err = new Error(
          'Stripe: Unknown arguments (' + args + '). Did you mean to pass an options ' +
          'object? See https://github.com/stripe/stripe-node/wiki/Passing-Options.' +
          ' (on API request to ' + requestMethod + ' `' + path + '`)'
        );
```

ob-stripe:
Ah, gotcha. Indeed, it appears that for API methods where `spec.path` is undefined, the code crashes while trying to format the error message.

ob-stripe:
Fixed in 5.6.1. Thanks for the contribution @reisel!

PR #437, opened 2018-03-22
fix invalid path join #436

reisel:
pathWithSymbols may be undefined and then an exception is thrown from the path.join method.
bug details: #436

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -72,7 +72,7 @@ StripeResource.prototype = {
   createResourcePathWithSymbols: function(pathWithSymbols) {
     return '/' + path.join(
       this.resourcePath,
-      pathWithSymbols
+      pathWithSymbols || ''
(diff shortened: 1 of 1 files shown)


ob-stripe:
@reisel Thanks for the fix, however we cannot use it as-is as the default parameter syntax was only enabled by default in Node v6 [1], and the library still needs to support v4.

Can you change the fix to do something like `var pathWithSymbols = typeof pathWithSymbols  !== 'undefined' ?  pathWithSymbols  : '';`?

[1] https://node.green/#ES2015-syntax-default-function-parameters

reisel:
@ob-stripe how about
```javascript
createResourcePathWithSymbols: function(pathWithSymbols) {
    return '/' + path.join(
      this.resourcePath,
      pathWithSymbols || ''
    ).replace(/\\/g, '/'); // ugly workaround for Windows
  }
```

reisel:
@ob-stripe updated.

PR #439, opened 2018-03-23
Use native promises

### LINKED ISSUE #438: Consider removing Bluebird
Hi,

I was wondering why do you need Bluebird as a dependency? I am packaging this module for AWS Lambda and want to optimize the size of the final bundle. Here is what I see: stripe-node takes 102 KB, bluebird requires 613 KB!

I think the library may not need such a heavy dependency. Is it possible to remove it?

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @OrKoN

- Use native ES6 promises instead of Bluebird promises
- ~Add a `setPromise` configuration option for users to provide their own `Promise` implementation~

Fixes #438.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
     "env": {
-        "node": true
+        "node": true,
+        "es6": true
     },
     "extends": "eslint:recommended",
     "rules": {
--- a/lib/StripeMethod.basic.js
+++ b/lib/StripeMethod.basic.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var Promise = require('bluebird');
 var isPlainObject = require('lodash.isplainobject');
 var stripeMethod = require('./StripeMethod');
 
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var Promise = require('bluebird');
 var utils = require('./utils');
 var OPTIONAL_REGEX = /^optional!/;
 
--- a/package.json
+++ b/package.json
@@ -29,13 +29,12 @@
     "chai": "~4.1.2",
     "chai-as-promised": "~7.1.1",
     "coveralls": "^3.0.0",
-    "eslint": "^4.6.1",
+    "eslint": "^4.19.1",
     "eslint-plugin-chai-friendly": "^0.4.0",
-    "mocha": "~3.5.3",
+    "mocha": "~5.0.5",
     "nyc": "^11.3.0"
   },
   "dependencies": {
-    "bluebird": "^3.5.0",
     "lodash.isplainobject": "^4.0.6",
     "qs": "~6.5.1",
     "safe-buffer": "^5.1.1"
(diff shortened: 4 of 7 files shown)


brandur-stripe:
Awesome!

I'm just thinking about the idea of allowing the library's `Promise` implementation to be set — TBH, I don't know enough about the JS ecosystem to know whether people would want to do this, but it seems to me that given the historical here, i.e., JS didn't have promises so various libraries appeared to add them, but then the core language started supporting them, it seems to me that the overwhelmingly preferred is going to be wide convergence on native promises.

With that assumption, we may even want to not expose a way to configure a promise if possible so that we won't have to deal with that additional API over the long term. I wonder if we should hide that configuration for now, and then add it in if/when someone asks for it?

ob-stripe:
> TBH, I don't know enough about the JS ecosystem to know whether people would want to do this

TBH, I don't either 😆 

That said, I've done some quick research and it seems people prefer third-party libraries like Bluebird over native promises for two reasons:
- performance
- much more extensive API

In our case, performance is probably a non-issue (the time spent in the Promise implementation is likely negligible compared to the time waiting for network requests to complete), and this PR proves we can get by with the native promises API. (Of course, users can still use third-party promise libraries in their own code.)

So, yeah, it probably makes sense for us to remove the `setPromise` method entirely and just use native promises always. I'll update the PR.

brandur-stripe:
Haha, thanks OB. I just hope that I wasn't too far off point and that we don't just have to bring this back to your original implementation in a week ;)

LGTM.

OrKoN:
I have run the latest benchmark on my laptop:

```
results for 10000 parallel executions, 1 ms per I/O op

file                                       time(ms)  memory(MB)
callbacks-baseline.js                           192       28.40
callbacks-suguru03-neo-async-waterfall.js       233       35.93
promises-bluebird-generator.js                  272       38.94
promises-bluebird.js                            294       47.81
promises-cujojs-when.js                         369       64.38
promises-tildeio-rsvp.js                        435       92.23
promises-lvivski-davy.js                        452       94.56
promises-native-async-await.js                  482      103.48
promises-ecmascript6-native.js                  495       96.97
promises-then-promise.js                        531       70.89
generators-tj-co.js                             532       86.63
callbacks-caolan-async-waterfall.js             615       81.26
promises-calvinmetcalf-lie.js                   876      159.32
promises-dfilatov-vow.js                        905      114.75
promises-obvious-kew.js                         967      116.49
streamline-generators.js                        973       78.28
promises-medikoo-deferred.js                   1175      107.42
observables-pozadi-kefir.js                    1205      143.38
streamline-callbacks.js                        1606       99.76
observables-Reactive-Extensions-RxJS.js        2328      183.53
promises-kriskowal-q.js                        7993      283.82
observables-caolan-highland.js                10067      496.00
observables-baconjs-bacon.js.js               22531      728.57

Platform info:
Linux 4.4.0-116-generic x64
Node.JS 8.10.0
V8 6.2.414.50
Intel(R) Core(TM) i7-4710HQ CPU @ 2.50GHz × 8
```

brandur-stripe:
@OrKoN Thanks for doing that! Looks liike amazingly, Bluebird _is_ still faster. I wouldn't have thought that the native promises would have so many more opportunities for optimization compared to a userspace library.

PR #441, opened 2018-04-02
Support custom agent with keepAlive: true

kesla:
When using a custom http agent and keepAlive is set the request won't create new sockets, meaning that the `'connect'` or `'secureConnect'` events won't run.

This fixes that.


brandur-stripe:
Oh, nice. Thanks for the fix!

brandur-stripe:
Released as 5.7.0.

PR #442, opened 2018-04-03
Fix indentation in StripeResource.js

kesla:
Sorry for missing this in #441


kesla:
Also I'm a bit surprised this wasn't caught by eslint?

ob-stripe:
> Also I'm a bit surprised this wasn't caught by eslint?

Unfortunately the indent rule is currently disabled -- it's something we should fix one of these days.

In the meantime, thanks for the contribution! Much appreciated.

PR #444, opened 2018-04-04
Flexible/Metered Billing API support

alexander-stripe:
Implement support for flexible billing options on plans and creating usage records.

### DIFF
--- a/lib/resources/UsageRecords.js
+++ b/lib/resources/UsageRecords.js
@@ -1,0 +1,16 @@
+'use strict';
+
+var stripeMethod = require('../StripeMethod');
+
+module.exports = require('../StripeResource').extend({
+  path: 'subscription_items',
+  create: function(args) {
+    var requestPath = args.subscription_item + '/usage_records';
+    var reqArgs = Object.assign({}, args);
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Excellent, thanks.

LGTM.

brandur-stripe:
Released as 5.8.0.

thinkocapo:
Ok so I upgraded to `v5.8.0` so I now have access to `stripe.usageRecords`, thank you as v5.6.1 didn't have it.

Is anyone finding the `stripe.usageRecords.create` call never triggers the callback?

I can see in my Dashboard that the usage record is getting created, which is great, but I'm not getting response. Seems its hanging on the stripe end? I'm calling this stripe method the same why that i'm calling other things like `stripe.subscriptions.create` and `stripe.customers.create`, and handling the callback the same, per the api documentation.

remi-stripe:
@thinkocapo Can you open an issue instead of commenting on the original PR? I can then follow up (I seem to be able to reproduce)

thinkocapo:
good idea, I did https://github.com/stripe/stripe-node/issues/451 thanks

PR #453, opened 2018-05-01
Fix UsageRecords so it passes through the rest of the arguments

### LINKED ISSUE #451: usageRecords Create not triggering callback
#### Background
I upgraded from v5.6.1 to **v5.8.0** to get access to `stripe.usageRecords`. Thank you for this improvement.

#### Problem
The **callback** to `stripe.usageRecords.create` is not executing. The call hangs. I went to my Dashboard.Stripe and saw the usage record was in fact created. Maybe something is going wrong in your stripe server endpoint?
```
router.post('/subscription/usage', async (req,res) => {
  // ...
  stripe.usageRecords.create({
    quantity: 1,
    timestamp: timestamp,
    subscription_item: subscription_item
  }, function(err, usageRecord) {
    console.log('usageRecord', usageRecord) // DOES NOT LOG
    if (err) {
      console.log('\n===== stripe usage error ====\n', err) // DOES NOT LOG
    } else {
      res.send(200) // DOES NOT HAPPEN, because my front-end says the call to /subscription/usage is still 'PENDING'
    }
```

#### Dashboard.Stripe verification the Usage Record was Created
Billing > Subscriptions > Selection Customer > Products/Pricing Plan > click 'View Usage This Period' > see timestamp for the call I just made in my testing, its now +1 more than previously

#### Overall, the Stripe Client is Working for me
I'm calling this stripe method the same why that i'm calling other things like `stripe.subscriptions.create` and `stripe.customers.create`, and handling the callback the same, per the api documentation. Those methods are working for me.

jlomas-stripe:
Makes sure that all of the elements of the original `arguments` array - with the first one being modified to extract and remove the SubscriptionItem ID - are passed through to the `stripeMethod()` call.

Fixes #451.

r? @brandur-stripe

### DIFF
--- a/lib/resources/UsageRecords.js
+++ b/lib/resources/UsageRecords.js
@@ -4,13 +4,19 @@ var stripeMethod = require('../StripeMethod');
 
 module.exports = require('../StripeResource').extend({
   path: 'subscription_items',
-  create: function(args) {
-    var requestPath = args.subscription_item + '/usage_records';
-    var reqArgs = Object.assign({}, args);
+  create: function() {
+    var args = Array.prototype.slice.call(arguments);
+
+    var reqArgs = args.shift();
+    var requestPath = reqArgs.subscription_item + '/usage_records';
+
     delete reqArgs.subscription_item;
+
+    args.unshift(reqArgs);
+
     return stripeMethod({
       method: 'POST',
       path: requestPath
-    }).bind(this)(reqArgs);
+    }).apply(this, args);
   }
 });
--- a/lib/resources/UsageRecords.js
+++ b/lib/resources/UsageRecords.js
@@ -1,22 +1,14 @@
 'use strict';
 
-var stripeMethod = require('../StripeMethod');
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
 
-module.exports = require('../StripeResource').extend({
+module.exports = StripeResource.extend({
   path: 'subscription_items',
-  create: function() {
-    var args = Array.prototype.slice.call(arguments);
 
-    var reqArgs = args.shift();
-    var requestPath = reqArgs.subscription_item + '/usage_records';
-
-    delete reqArgs.subscription_item;
-
-    args.unshift(reqArgs);
-
-    return stripeMethod({
-      method: 'POST',
-      path: requestPath
-    }).apply(this, args);
-  }
+  create: stripeMethod({
+    method: 'POST',
+    path: '{subscriptionItem}/usage_records',
+    urlParams: ['subscriptionItem'],
+  }),
 });
--- a/test/resources/UsageRecords.spec.js
+++ b/test/resources/UsageRecords.spec.js
@@ -9,18 +9,56 @@ describe('UsageRecords Resource', function() {
       stripe.usageRecords.create({
         subscription_item: 'si_123',
         quantity: 123,
-        timestmap: 123321,
+        timestamp: 123321,
         action: 'increment'
       });
+
       expect(stripe.LAST_REQUEST).to.deep.equal({
         method: 'POST',
         url: '/v1/subscription_items/si_123/usage_records',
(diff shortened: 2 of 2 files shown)


brandur-stripe:
@jlomas-stripe Thanks for taking a crack at this!

Just to make sure that I understand this correctly: what exactly is unusual about this function that means we need this custom code here, but not most other places?

fred-stripe:
cc @alexander-stripe

brandur-stripe:
@jlomas-stripe Sorry, I totally lost track of this one!

Looking over this again ... I'm thinking that I might have been a little hasty merging the original PR. This should be modeled closely after application fee refunds, which have a similar setup in that their a subresource of another resource. Its implementation is quite simple:

``` node
  path: 'application_fees/{feeId}/refunds',
});
```

I'm actually not 100% sure that the `create` there works as expected because there's no test for it, but the method for creating a new refund that's recommended on our docs is this:

``` node
stripe.applicationFees.createRefund(
  "fee_1B73DOKbnvuxQXGuhY8Aw0TN",
  function(err, applicationFee) {
    // asynchronously called
  }
);
```

Which also has a very simple implementation:

``` node
  createRefund: stripeMethod({
    method: 'POST',
    path: '/{feeId}/refunds',
    urlParams: ['feeId'],
  }),
```

I'm sort of thinking that we should change usage records over to use that system instead of what we have now ... what do you think?

jlomas-stripe:
@brandur-stripe Agreed.  I actually had those changes queued up in my local repo already.

This just adds the `create` (similar to `createRefund`), since that's all we can currently do anyways.

brandur-stripe:
Ah, nice. Thanks @jlomas-stripe! (And my mistake — you said this before, but I didn't quite get it.)

Alright, I just looked at stripe-node usage on this endpoint, and unfortunately there is some already, so I think we're going to have to roll this as a major version bump given the breaking API change. Still, it just be a very easy migration.

remi-stripe:
could we support both and deprecate the old one?

brandur-stripe:
> could we support both and deprecate the old one?

Yeah probably, but because both methods would have the same name (`create`), it might require some more involved code that'll be harder to maintain.

IMO, given that this is a relatively new product and there can't be that many users on it yet, we should try to just roll forward on this one. Thoughts?

remi-stripe:
I'm assuming you meant keeping the move forward as a major version? If so I agree, it's just that major versions should also be used to remove/deprecate things we've wanted remove for a while.

brandur-stripe:
> I'm assuming you meant keeping the move forward as a major version? If so I agree, it's just that major versions should also be used to remove/deprecate things we've wanted remove for a while.

Yeah, agreed in principle — did you have anything in mind?

If there's nothing that we want to do opportunistically though, I'd rather not block too long on this. If we want to change something again later, we can always bump the major version again — users upgrading to compensate for 5 breaking changes across two version numbers is the same amount of work as compensating for 5 breaking changes across one version number.

remi-stripe:
Okay, I'm fine to leave as is then and we can always prioritize a v6 soon! Sorry for the chunder

brandur-stripe:
> Okay, I'm fine to leave as is then and we can always prioritize a v6 soon! Sorry for the chunder

Just to be clear, I'm suggesting that we release V6 now (and if there are any immediate backwards-incompatible changes I can bundle them in), then release V7 later if we want to bundle up any larger backwards-incompatible changes. Are you still okay with that?

remi-stripe:
yes, sorry, I got my major versions number mixed up. Please go ahead and ignore most of what I said earlier.

brandur-stripe:
Okay, great. Thanks @jlomas-stripe @remi-stripe! Pulling this in.

brandur-stripe:
(By the way, I'll take updating the docs.)

brandur-stripe:
Released as 6.0.0.

PR #456, opened 2018-05-07
Add support for v1/issuer_fraud_records endpoint

fay-stripe:
need to wait for update to stripe mock

### DIFF
--- a/lib/resources/IssuerFraudRecords.js
+++ b/lib/resources/IssuerFraudRecords.js
@@ -4,13 +4,25 @@ var StripeResource = require('../StripeResource');
 var stripeMethod = StripeResource.method;
 
 module.exports = StripeResource.extend({
-  path: 'issuer_fraud_records',
-  includeBasic: ['list', 'retrieve'],
+  path: '',
 
-  retrieveFromCharge: stripeMethod({
+  list: stripeMethod({
+    method: 'GET',
+    path: 'issuer_fraud_records',
+  }),
+
+  retrieve: stripeMethod({
     method: 'GET',
     path: function(urlData) {
-      return '?charge=' + urlData.chargeId;
+      return 'issuer_fraud_records/' + urlData.issfrId;
+    },
+    urlParams: ['issfrId'],
+  }),
+
+  listFromCharge: stripeMethod({
+    method: 'GET',
+    path: function(urlData) {
+      return 'issuer_fraud_records?charge=' + urlData.chargeId;
     },
     urlParams: ['chargeId'],
   }),
--- a/lib/resources/IssuerFraudRecords.js
+++ b/lib/resources/IssuerFraudRecords.js
@@ -9,7 +9,9 @@ module.exports = StripeResource.extend({
 
   retrieveFromCharge: stripeMethod({
     method: 'GET',
-    path: '',
+    path: function(urlData) {
+      return '?charge=' + urlData.chargeId;
+    },
     urlParams: ['chargeId'],
   }),
 });
--- a/lib/resources/IssuerFraudRecords.js
+++ b/lib/resources/IssuerFraudRecords.js
@@ -1,0 +1,15 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+  path: 'issuer_fraud_records',
+  includeBasic: ['list', 'retrieve'],
+
+  retrieveFromCharge: stripeMethod({
+    method: 'GET',
+    path: '',
+    urlParams: ['chargeId'],
+  }),
+});
--- a/lib/resources/IssuerFraudRecords.js
+++ b/lib/resources/IssuerFraudRecords.js
@@ -1,29 +1,9 @@
 'use strict';
 
 var StripeResource = require('../StripeResource');
-var stripeMethod = StripeResource.method;
 
 module.exports = StripeResource.extend({
-  path: '',
+  path: 'issuer_fraud_records',
 
-  list: stripeMethod({
-    method: 'GET',
-    path: 'issuer_fraud_records',
-  }),
-
-  retrieve: stripeMethod({
-    method: 'GET',
-    path: function(urlData) {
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Thanks! Fay, as discussed in chat, stripe-node is not quite on stripe-mock yet.

Can you see if you can get your failing test fixed? You can see CI for details, but should also be able to reproduce it locally.

```
  IssuerFraudRecord Resource
    retrieve
      1) Sends the correct request for issfr ID
  216 passing (39s)
  1 failing
  1) IssuerFraudRecord Resource
       retrieve
         Sends the correct request for issfr ID:
     TypeError: Cannot read property 'retrieve' of undefined
      at Context.<anonymous> (test/resources/IssuerFraudRecords.spec.js:9:35)
```

lib/resources/IssuerFraudRecords.js
```diff
@@ -0,0 +1,17 @@
...
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+  path: 'issuer_fraud_records',
+  includeBasic: ['list', 'retrieve'],
+
+  retrieveFromCharge: stripeMethod({
+    method: 'GET',
+    path: function(urlData) {
+      return '?charge=' + urlData.chargeId;
```
fay-stripe: @brandur @matt-stripe any advice here? This still isn't right, because then the url is /v1/issuer_fraud_records/?charge=.... whereas I want it to be /v1/issuer_fraud_records?charge=....

I haven't found another example of this ^ but please do direct me if there is one

fay-stripe: I also would actually like the test to be `data: {charge: 'ch_123456789'}` versus doing the urlparams. At least, I think that's what we'd prefer to test here

fay-stripe:
I got the tests to pass but I'm not sure if there's a better / more correct way of doing the `?charge=` here. 
ptal @brandur

fay-stripe:
I think I keep tagging your brandur handle instead of brandur-stripe. Sorry about that!

lib/resources/IssuerFraudRecords.js
```diff
@@ -0,0 +1,29 @@
...
+    urlParams: ['issfrId'],
+  }),
+
+  listFromCharge: stripeMethod({
+    method: 'GET',
+    path: function(urlData) {
+      return 'issuer_fraud_records?charge=' + urlData.chargeId;
+    },
+    urlParams: ['chargeId'],
+  }),
```
brandur-stripe: Like in Ruby, `list` allows you to pass parameters to it. Here's an example from the subscriptions test suite:

``` js
      stripe.subscriptions.list({
        limit: 3,
        customer: 'test_cus',
        plan: 'gold',
      });
```

lib/resources/IssuerFraudRecords.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
+
+module.exports = StripeResource.extend({
+  path: '',
```
brandur-stripe: Why empty the `path` up here and then pass one in each method? I think you should be able to implement this almost identically to country spec:

``` js

  path: 'country_specs',

  includeBasic: [
    'list', 'retrieve',
  ],
});
```

test/resources/IssuerFraudRecords.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    });
+    it('Sends the correct request for charge ID', function() {
+      stripe.issuerFraudRecords.listFromCharge('ch_123456789');
+      expect(stripe.LAST_REQUEST).to.deep.equal({
+        method: 'GET',
+        url: '/v1/issuer_fraud_records?charge=ch_123456789',
+        data: {},
+        headers: {},
+      });
+    });
```
brandur-stripe: I think we can safely drop this test case, but if we don't, it should go in the `list` block below instead of this `retrieve` block.

lib/resources/IssuerFraudRecords.js
fay-stripe: 👍 I didn't know list could take a param as I missed the other examples of it. When I implemented a new method, the base path automatically tacks on `/` when anything is appended to it, so I separated it out in order to make sure the `/` didn't get inserted. 

But I'll switch it over to `list` and that should make everything cleaner. Thanks!

fay-stripe:
This build is taking a long time to start. I think I made the right changes though

ptal @brandur-stripe

test/resources/IssuerFraudRecords.spec.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+var stripe = require('../testUtils').getSpyableStripe();
+var expect = require('chai').expect;
+
+describe('IssuerFraudRecord Resource', function() {
+  describe('retrieve', function() {
+    it('Sends the correct request for issfr ID', function() {
```
brandur-stripe: Can you truncate this down to just "Sends the correct request" for consistency?

test/resources/IssuerFraudRecords.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    it('Sends the correct request', function() {
+      stripe.issuerFraudRecords.list();
+      expect(stripe.LAST_REQUEST).to.deep.equal({
+        method: 'GET',
+        url: '/v1/issuer_fraud_records',
+        data: {},
+        headers: {},
+      });
+    });
+    it('Sends the correct request for charge ID', function() {
```
brandur-stripe: Let's just remove this test case please. It's not actually exercising any new code that we've written.

fay-stripe: 👍

test/resources/IssuerFraudRecords.spec.js
fay-stripe: 👍

fay-stripe:
Made those changes!
ptal @brandur-stripe

brandur-stripe:
LGTM. Thanks Fay!

brandur-stripe:
Released as 5.9.0.

PR #459, opened 2018-05-12
Export errors to allow typed access to them

### LINKED ISSUE #457: Export Stripe error classes
Would it be possible for all stripe error classes to be exported so we can catch them via `instanceof`?

brandur:
Exports the errors module and the types it contains as `stripe.errors`.
This allows users to match on an error with the `instanceof` operator
instead of having to match on the string generated by accessing an
object's `type`.

There's a bit of an unfortunate singular/plural mismatch (the file is called
`Error.js`), but I called the export `stripe.errors` to match the
already-plural `stripe.webhooks`.

Fixes #457.

r? @jlomas Hey! Could I get your opinion on this?
cc @stripe/api-libraries


brandur-stripe:
(And oops — I meant @jlomas-stripe. Sorry @JLomas!)

jlomas-stripe:
@brandur-stripe This looks great.  It looks like `Error.js` is only referenced in 3 places:

    lib/StripeResource.js
    lib/Webhooks.js
    test/Error.spec.js

...so maybe we should also take this opportunity rename the file to `Errors.js` as well?  Though ... it should still export `Error`, so maybe not.   ¯\\\_(ツ)_/¯

Either way,

lgtm

brandur-stripe:
> ...so maybe we should also take this opportunity rename the file to Errors.js as well? Though ... it should still export Error, so maybe not. ¯\_(ツ)_/¯

Yeah, that's basically what stopped me. I think this would still be a good cleanup, but we should also rename the symbol `Error` to `Errors` too, and that would've added quite a bit of noise to this PR.

I'll merge this and then see if that tweak works the way we hope on a different PR. Thanks for taking a look!

brandur-stripe:
Alright, I tried to follow this up with a rename, but it turned into quite a deep rabbit hole that I wasn't prepared for. `Error.js` seems to be doing something to manipulate the base `Error` that I don't understand, and a simple rename will cause a failure on this check:

``` node
  it('Populates with type and message params', function() {
    var e = new Error('FooError', 'Foo happened');
    expect(e).to.have.property('type', 'FooError');
```

It also has the additional problem that the author conflated a class with a package, and throughout much of the file, the singular `_Error` is more conceptually correct.

I'm just going to leave it for now.

SimonSchick:
Little bit late here but, why are the errors exported on the client rather than directly from the package, this seems a little counter intuitive since errors are not bound to a stripe instance.

Issue #375, opened 2017-08-22
Bad `options` keys; should we throw?


jlomas-stripe:
A question that came to mind as a result of #374, where I sent in `api_version` instead of `stripe_version` and it didn't surface until the `latest` API Version was bumped (and my set test value no longer matched `latest`):

Should we be throwing if we see something in the `options` that doesn't belong?

> On the one hand, it would be useful to get immediate feedback that you'd provided bad arguments and we're not gonna do what you asked;

> On the other hand, we'll need to keep that list up to date - which isn't really that much of a burden.

brandur-stripe:
We did something a little similar in Ruby where we whitelist known options so that we could warn people when they accidentally put them in the parameters hash instead of the options hash:

https://github.com/stripe/stripe-ruby/pull/545

This pattern helps insulate users from dumb mistakes that take time to debug and I like it. +1 to doing something like what you suggest.

One note is that we ended up settling on a warning to stderr instead of throwing an error. This has the advantage of being backwards compatible, and also forwards compatible to a degree (if we introduced a new option, it could be used even with an older version of the library, albeit with a warning). Thoughts?

brandur-stripe:
And oops, I'm reading my inbox backwards and just saw #306 which is very similar to the stripe-ruby issue I linked.

jlomas-stripe:
No worries.  #306 is actually just "throw if you combine args and options", as we currently just ignore the args if you combine them - so it's a bit different.

I am 100% onside with warning via console or equivalent instead of throwing.  Currently thinking of implementing this using [`process.emitWarning()`](https://nodejs.org/docs/v6.0.0/api/process.html#process_process_emitwarning_warning_name_ctor) which started being available in 6.0.0, and wrapping in a fn that will either use that if present, or just `console.error()` if not.

There's some extra usefulness to using `process.emitWarning()` if available - i.e., loggers using `process.on('warning')`, etc. - so that seems to be the right approach...?

brandur-stripe:
> I am 100% onside with warning via console or equivalent instead of throwing. Currently thinking of implementing this using process.emitWarning() which started being available in 6.0.0, and wrapping in a fn that will either use that if present, or just console.error() if not.

Yep, that sounds good to me.

robz-stripe:
Looks like this was fixed in #465.

PR #465, opened 2018-05-30
Warn on unknown `options`

jlomas-stripe:
I.e., "Do the thing I said I'd do in #375".

I also fixed some tests that weren't running properly (because they were async but didn't include a `done`), and I'm ... also not sure that `emitWarning()` is the right way, and it might be better to just `console.warn()` instead.  Thoughts?

r? @brandur-stripe

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -75,7 +75,7 @@ var utils = module.exports = {
     // (the first being args and the second options) and with known
     // option keys in the first so that we can warn the user about it.
     if (optionKeysInArgs.length > 0 && optionKeysInArgs.length !== argKeys.length) {
-      emitStripeWarning( // eslint-disable-line
+      emitWarning(
         'Options found in arguments (' + optionKeysInArgs.join(', ') + '). Did you mean to pass an options ' +
         'object? See https://github.com/stripe/stripe-node/wiki/Passing-Options.'
       );
@@ -104,7 +104,7 @@ var utils = module.exports = {
         });
 
         if (extraKeys.length) {
-          emitStripeWarning('Invalid options found (' + extraKeys.join(', ') + '); ignoring.');
+          emitWarning('Invalid options found (' + extraKeys.join(', ') + '); ignoring.');
         }
 
         if (params.api_key) {
@@ -205,10 +205,9 @@ var utils = module.exports = {
   },
 };
 
-function emitStripeWarning(warning) {
+function emitWarning(warning) {
   if (typeof process.emitWarning !== 'function') {
-    /* eslint-disable no-console */
-    return console.warn('Stripe: ' + warning);
+    return console.warn('Stripe: ' + warning); /* eslint-disable-line no-console */
   }
 
   return process.emitWarning(warning, 'Stripe');
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -75,8 +75,8 @@ var utils = module.exports = {
     // (the first being args and the second options) and with known
     // option keys in the first so that we can warn the user about it.
     if (optionKeysInArgs.length > 0 && optionKeysInArgs.length !== argKeys.length) {
-      console.warn( // eslint-disable-line no-console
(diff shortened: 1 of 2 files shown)


jlomas-stripe:
OMG THEY ALL FINALLY PASSED :joy:

lib/utils.js
```diff
@@ -75,8 +75,8 @@ var utils = module.exports = {
     // (the first being args and the second options) and with known
     // option keys in the first so that we can warn the user about it.
     if (optionKeysInArgs.length > 0 && optionKeysInArgs.length !== argKeys.length) {
-      console.warn( // eslint-disable-line no-console
-        'Stripe: Options found in arguments (' + optionKeysInArgs.join(', ') + '). Did you mean to pass an options ' +
+      emitStripeWarning( // eslint-disable-line
```
brandur-stripe: Maybe just call it `emitWarning` — I don't think we need the "Stripe" in there.

Also, now that there's no more console call here, can we remove the `eslint-disable-line`?

lib/utils.js
```diff
@@ -195,3 +204,12 @@ var utils = module.exports = {
     return obj;
   },
 };
+
+function emitStripeWarning(warning) {
+  if (typeof process.emitWarning !== 'function') {
+    /* eslint-disable no-console */
```
brandur-stripe: Can you just make sure this lint doesn't need to be re-enabled after the console call? (I'm honestly not sure.) It might be better just to use the single line disable `eslint-disable-line` like we had above.

test/utils.spec.js
```diff
@@ -262,15 +285,30 @@ describe('utils', function() {
...
-  // Un-shim `console.warn`
-  console.warn = _warn;
+    doWithShimmedConsoleWarn();
+
+    process.nextTick(function() {
+      process.removeListener('warning', onProcessWarn);
+    })
+  }
 }
 /* eslint-enable no-console */
```
brandur-stripe: Ah yeah, I think you need one of these `eslint-enable` to re-enable the lint if you don't use the per-line syntax.

lib/utils.js
jlomas-stripe: Yes, and yes.  :)

jlomas-stripe:
@brandur-stripe Apparently I had no idea how eslint actually worked. :joy:

test/utils.spec.js
```diff
@@ -262,15 +285,31 @@ describe('utils', function() {
 });
 
 /* eslint-disable no-console */
-function handleConsoleWarns(doWithShimmedConsoleWarn, onWarn) {
-  // Shim `console.warn`
-  var _warn = console.warn;
+function handleWarnings(doWithShimmedConsoleWarn, onWarn) {
+  /* eslint-disable no-console */
```
brandur-stripe: Another small one, but I think this `eslint-disable` might have been duplicated. Can we just have one of these? (Between the ones on lines 287 and 289.)

test/utils.spec.js
```diff
@@ -262,15 +285,31 @@ describe('utils', function() {
...
 
-  doWithShimmedConsoleWarn();
+    // Un-shim `console.warn`,
+    console.warn = _warn;
 
-  // Un-shim `console.warn`
-  console.warn = _warn;
+    /* eslint-enable no-console */
+  } else {
+    /* eslint-disable no-inner-declarations */
```
brandur-stripe: And here, can we either re-enable this lint after the inner declaration or use `eslint-disable-line`?

brandur-stripe:
> @brandur-stripe Apparently I had no idea how eslint actually worked. 😂

Haha, you and me both!

ptal @jlomas-stripe

jlomas-stripe:
@brandur-stripe I ... totally missed those. 🤦‍♂️   Fixed 'em.

brandur-stripe:
Thanks!

brandur-stripe:
Released as 6.1.0.

Issue #401, opened 2017-10-23
Feature Request - FileUpload should start supporting streams


mckn:
Hello,

I would be super happy if you could start to support uploading files to your service by passing a stream. It would increase performance and improve the structure of the code.

Thanks in advance!

ob-stripe:
Hey @mckn! Thanks for the feature request. This does sound like a good idea!

ob-stripe:
This was implemented by @jjsquillante in #471 and released as 6.3.0.

PR #471, opened 2018-06-24
Stream support for FileUpload

jjsquillante:
Hey all at Stripe! 

First, I want to say: from documentation to code, Stripe communicates its design of the API libraries extremely well - thank you very much!

Ok down to the PR, I was implementing Stripe for a recent project, and figured in order to truly understand the Stripe-node library, I might as well dive in to an issue or feature request to get my hands dirty. I noticed issue #401 and decided to give it a go. 

This Pull Request allows a user to upload a file by passing a stream. Within the `StripeMethod`, we now check for a stream, convert it to a buffer, and begin the  `_request` process as normal. 

I've added tests to confirm that the new feature is working as designed and did not break any existing functionality.

I'm certainly open to any feedback, so if there's anything you'd like to see different or that can be improved, let me know! 

Please reach out if there's anything that I can help further clarify. Thanks!


ob-stripe:
Hi @jjsquillante, thanks for your kind words as well as the contribution! :)

This looks good to me, but going to ask some colleagues with better JS skills than mine to take a look as well.

r? @jlomas-stripe @rattrayalex-stripe

lib/StripeMethod.js
```diff
@@ -115,7 +115,22 @@ function stripeMethod(spec) {
...
+
+      if (isStream) {
+        var bufferArray = [];
+        data.file.data
+          .on('data', function(line) {
+            bufferArray.push(line);
+          })
+          .on('end', function() {
+            var dataBuffered = Object.assign({}, data);
+            dataBuffered.file.data = Buffer.concat(bufferArray);
```
ob-stripe: I believe this is fine, but would you mind adding `var Buffer = require('safe-buffer').Buffer;` at the top of the file? That way if we add more code that uses `Buffer` in this file it will use the safe version.

jjsquillante: @ob-stripe thanks for the quick response and catching my oversight on `safe-buffer`! I noticed `safe-buffer` in the MultipartDataGenerator.js file, but totally forgot about requiring it here. I'll update the code asap and keep a look out for any other feedback from @jlomas-stripe @rattrayalex-stripe .

jlomas-stripe:
Hey hey.  Super-cool. :)

I think there should probably be a test in the `flows.spec.js` that actually hits the API to confirm this does what it says on the tin.  Something like:

```
  describe('FileUpload', function() {
    it('Allows you to upload a file as a stream', function() {
      var testFilename = path.join(__dirname, 'resources/data/minimal.pdf');
      var f = fs.createReadStream(testFilename);

      return expect(
        stripe.fileUploads.create({
          purpose: 'dispute_evidence',
          file: {
            data: f,
            name: 'minimal.pdf',
            type: 'application/octet-stream',
          },
        }).then(null, function(error) {
          return error;
        })
      ).to.eventually.have.nested.property('size', 739);
    });
  });
```

More generally, we have a [`requestDataProcessor` function](https://github.com/stripe/stripe-node/blob/master/lib/StripeResource.js#L49) function on the `StripeResource` - which is only currently used by `FileUpload` - that feels like a better place for this, but that function is unfortunately synchronous at the moment.

I feel like the right way to do this is to refactor the `requestDataProcessor` function to be async and then do this stuff right in the `FileUpload.js`'s [`requestDataProcessor`](https://github.com/stripe/stripe-node/blob/master/lib/resources/FileUploads.js#L12-L20), mostly because it really feels like this is specific to `FileUpload`...?

jlomas-stripe:
You're welcome! :)

cc @ob-stripe / @brandur-stripe / @rattrayalex-stripe on my above suggestion; does that make sense to y'all?  What jjsquillante has done here definitely works just fine, but it feels a bit out of place in terms of where it lives, since it only affects one type of API request.

brandur-stripe:
Thanks for the effort here @jjsquillante!

> cc @ob-stripe / @brandur-stripe / @rattrayalex-stripe on my above suggestion; does that make sense to y'all? What jjsquillante has done here definitely works just fine, but it feels a bit out of place in terms of where it lives, since it only affects one type of API request.

Yeah, agreed that `FileUpload` would be a better home for this — putting it in `StripeMethod` would be okay if we ever wanted to generalize this for some reason, but I don't ever see that happening. Very likely all files and file-like objects will always go the Stripe area specifically through `FileUpload` (or a direct successor to it).

lib/utils.js
```diff
@@ -203,6 +203,17 @@ var utils = module.exports = {
...
     return obj;
   },
+
+  /**
+  * Determine if file data is a derivative of EventEmitter class.
+  * https://nodejs.org/api/events.html#events_events
+  */
+  checkForStream: function (obj) {
+    if (obj.file && obj.file.data) {
+      return typeof obj.file.data.on === 'function' && typeof obj.file.data.emit === 'function';
```
rattrayalex-stripe: Why not do an `instanceof` check? This is probably fine too, just curious.

rattrayalex-stripe:
Nice, thanks @jjsquillante ! Other than a +1 to @jlomas-stripe 's suggestions, this looks good to me.

lib/utils.js
jjsquillante: @rattrayalex-stripe good question! An `instanceof` check would have been more explicit, but I'm not aware of `require('events').EventEmitter` available in the utils file. In retrospect, I was trying to implement the feature with just minimal additions to the existing code. When it comes to collaboration, though, it makes sense to be more explicit - I should have probably just added it! Thanks for the feedback. I plan to refactor the `requestDataProcessor` function to async, and if I need this check, I'll implement via `instanceof `. Let me know if you have any other feedback - thanks!

jjsquillante:
@ob-stripe @brandur-stripe @jlomas-stripe @rattrayalex-stripe thanks for all of the feedback. I appreciate you all being open to my contribution and I have enjoyed working within Stripe's library. Sounds like the best strategy is to close this PR and I'll resolve to refactor the `requestDataProcessor` function to be async. sound good?

lib/utils.js
rattrayalex-stripe: Cool, thanks! That reasoning makes sense. A duck-type check as you have here may make more sense; let's see how it looks after the refactor. Cheers

jlomas-stripe:
@jjsquillante You're very welcome, and I'm glad you like the library!

Since this PR represents the reason/inspiration for the refactor, and since the tests for streaming `FileUpload` will exercise that refactor anyways, it seems reasonable to me to just stick with this PR.

jjsquillante:
@jlomas-stripe ok - sounds good. So to confirm, I'll leave this PR open and will update once the refactor of `requestDataProcessor` function to async is complete.

jlomas-stripe:
Sure, or just make all the changes in here and re-title - whatever you prefer!

jjsquillante:
Hey all - @jlomas-stripe @rattrayalex-stripe @ob-stripe 
I've refactored the code and added the test @jlomas-stripe provided to `flows.spec.js`. Let me know if there's anything you'd like to see different. Thanks!

lib/StripeResource.js
```diff
@@ -239,30 +239,46 @@ StripeResource.prototype = {
 
   _request: function(method, path, data, auth, options, callback) {
     var self = this;
+    var requestProcessor;
     var requestData;
 
     if (self.requestDataProcessor) {
-      requestData = self.requestDataProcessor(method, data, options.headers);
+      requestProcessor = self.requestDataProcessor(method, data);
```
jlomas-stripe: Since this now gets a function, rather than just returning the data, it might be clearer if it were called something like `requestDataProcessorFn` or `getRequestDataProcessorFn` I think?

jlomas-stripe: The variable also might be a bit clearer if it were named `processRequestData` since that's what it's going to to.

lib/MultipartDataGenerator.js
```diff
@@ -39,7 +39,10 @@ function multipartDataGenerator(method, data, headers) {
   }
   push('--' + segno + '--');
 
-  return buffer;
+  if (!callback) {
```
jlomas-stripe: This function doesn't have any async pieces; does it really need the callback in here?

lib/StripeResource.js
```diff
@@ -239,30 +239,46 @@ StripeResource.prototype = {
...
       requestData = utils.stringifyRequestData(data || {});
     }
 
     var apiVersion = this._stripe.getApiField('version');
 
-    var headers = self._defaultHeaders(auth, requestData.length, apiVersion);
+    if (requestProcessor) {
+      requestProcessor(method, data, options.headers, function(err, buffer) {
+        if (err) {
+          var handleError = self._errorHandler({}, callback);
```
jlomas-stripe: I think this one is probably better handled with a different kind of error, rather than overloading the connection error, since this isn't a connection (or other Stripe) error; (i presume) it'll be something related to the Buffer itself, so we should probably expose it as such.

lib/resources/FileUploads.js
```diff
@@ -9,14 +10,45 @@ module.exports = StripeResource.extend({
...
     data = data || {};
 
     if (method === 'POST') {
-      return multipartDataGenerator(method, data, headers);
+      return determineSource(data);
     } else {
       return utils.stringifyRequestData(data);
     }
+
+    function determineSource(data) {
```
jlomas-stripe: With the approach you've got here, this function does more than just `determineSource` - it `getProcessorForSourceType` - so it probably should be renamed so it's clearer as to what you're getting back.

lib/resources/FileUploads.js
```diff
@@ -9,14 +10,45 @@ module.exports = StripeResource.extend({
 
   overrideHost: 'uploads.stripe.com',
 
-  requestDataProcessor: function(method, data, headers) {
+  requestDataProcessor: function(method, data) {
```
jlomas-stripe: This got a bit more complicated than I expected  - I imagined something very similar to what you'd done previously, except that the entire `requestDataProcessor` function would just get an additional `callback` parameter that was always used (either sync or async), and the main code flow would either hit some 'default' function with callback, or the one specified here.

rattrayalex-stripe:
Thanks – I'll defer to @ob-stripe on this one

test/resources/FileUploads.spec.js
```diff
@@ -78,5 +78,40 @@ describe('File Uploads Resource', function() {
       expect(stripe.LAST_REQUEST).to.deep.property('url', '/v1/files');
       expect(stripe.LAST_REQUEST).to.deep.property('auth', TEST_AUTH_KEY);
     });
+
+    it('Streams a file and sends the correct file upload request', function() {
+      var testFilename = path.join(__dirname, 'data/minimal.pdf');
+      var f = fs.createReadStream(testFilename);
+
+      stripe.fileUploads.create({
```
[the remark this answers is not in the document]
jjsquillante: ah, my fault! nice catch @jlomas-stripe - I just ran it through the debugger with an intentional fail and noticed exactly what you were talking about. I'll update this asap.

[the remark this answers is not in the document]
jlomas-stripe: No worries.  This is a super-insidious mocha thing.  Not including a `done` callback and then doing something async used to catch me out unfortunately often in the good ol' days. :joy:

lib/StripeResource.js
jjsquillante: thanks for the feedback. I'll update the variable names to be more clear.

lib/StripeResource.js
jjsquillante: ok, cool - makes sense. Looking back now, it was kind of a 'cop out' to overload the connection error. I'll resolve to update and expose the error as such.

lib/MultipartDataGenerator.js
jjsquillante: you're correct - definitely does not need to have a callback in there. I'll update. thanks @jlomas-stripe

lib/resources/FileUploads.js
jjsquillante: @jlomas-stripe Definitely a bit more complicated but I believe I vetted the scenario you've mentioned and the nature of `streams` posed a problem. If I'm understanding your comment (correct me if I'm wrong), the gist is: `requestDataProcessor` would send the buffered `multipart/form-data` back to `_request` via the additional `callback` - which ever way handled (ie. data processed via `stream` or already buffered data). From there `_request` can process as normal.

The trouble I encountered is that when the `stream` is set by attaching an event listener `.on('data')`, the code is handled via `process.nextTick()` which runs the callback after the rest of the code and before the event loop. If I remember correctly, this caused `data` to be passed back to the `callback` `undefined` and then proceed through `_request` without the buffered data. The way I figured, returns a function in both scenarios so we can wait for the `stream` to finish processing. This will guarantee the buffered data exists when the rest of `_request` is processed by invoking the `setHeadersAndMakeRequest` within the `callback` from either scenario. My apologies if that was long-winded! :slightly_smiling_face:

If that logic may be flawed or there's a better way to go about this - I'm fine with scrapping this solution for the sake of clarity and going back to the drawing board. Let me know!

lib/resources/FileUploads.js
jjsquillante: sounds good, i'll rename and update!

lib/resources/FileUploads.js
jjsquillante: @jlomas-stripe ohhh no I misunderstood! All good, yea that'll definitely work inside `_request` - I have some free time later and can put it to action. thanks for further explaining!

jjsquillante: @jlomas-stripe just tested it and worked perfectly. I appreciate the help simplifying. I'll update the PR with the changes requested in the review.

jjsquillante:
hey all - I've made the requested updates from the review. Let me know if there's any other changes required or something I may have forgot! Thanks

ob-stripe:
@jlomas-stripe Mind doing another review? Cheers :)

jlomas-stripe:
@ob-stripe / @jjsquillante Yup, for sure; sorry for the delay there.  WIll peek it this morning. 👍

lib/resources/FileUploads.js
```diff
@@ -9,14 +10,38 @@ module.exports = StripeResource.extend({
...
 
-  requestDataProcessor: function(method, data, headers) {
+  requestDataProcessor: function(method, data, headers, callback) {
     data = data || {};
 
     if (method === 'POST') {
-      return multipartDataGenerator(method, data, headers);
+      return getProcessorForSourceType(data);
     } else {
       return utils.stringifyRequestData(data);
```
jlomas-stripe: This needs to make use of the callback flow here, so it's probably `return callback(null, utils.stringifyRequestData(data));`.

lib/StripeResource.js
```diff
@@ -241,28 +255,37 @@ StripeResource.prototype = {
...
-    } else {
-      requestData = utils.stringifyRequestData(data || {});
-    }
+    function makeRequestWithData(error, data) {
+      var apiVersion;
+      var headers;
 
-    var apiVersion = this._stripe.getApiField('version');
+      if (error) {
+        var handleError = self._errorHandlerStream(callback);
```
jlomas-stripe: I'd rather see this FileUpload-specific error generation/handling pushed to `FileUploads.js`, and have this just be more generic here so it can handle future cases, i.e. just:

    if (error) {
      return callback(error);
    }

At this point in the code, we don't (or shouldn't) really care what kind of error this is, we should just expose it in the `callback`.

lib/StripeResource.js
```diff
@@ -212,6 +212,20 @@ StripeResource.prototype = {
     }
   },
 
+  _errorHandlerStream: function(callback) {
```
jlomas-stripe: Per comment further down, I think the `StreamProcessingError` should be created (and wrap the incoming error) in `FileUploads.js`, since it's so specific to FileUploads.

lib/resources/FileUploads.js
```diff
@@ -9,14 +10,38 @@ module.exports = StripeResource.extend({
...
+      var bufferArray = [];
+      data.file.data.on('data', function(line) {
+        bufferArray.push(line);
+      }).on('end', function() {
+        var bufferData = Object.assign({}, data);
+        bufferData.file.data = Buffer.concat(bufferArray);
+        var buffer = fn(method, bufferData, headers);
+        callback(null, buffer);
+      }).on('error', function(err) {
+        callback(err, null);
```
jlomas-stripe: Per comments above, this is probably the right place to new up the `StreamProcessorError`.

test/flows.spec.js
```diff
@@ -522,4 +523,24 @@ describe('Flows', function() {
...
+            name: 'minimal.pdf',
+            type: 'application/octet-stream',
+          },
+        }).then(null, function(error) {
+          return error;
+        })
+      ).to.eventually.have.nested.property('size', 739);
+    });
+  });
 });
```
jlomas-stripe: We don't actually test that the error comes back as expected; this should do it:

```
    it('Surfaces stream errors correctly', function(done) {
      var mockedStream = new stream.Readable();
      mockedStream._read = function() {};

      var fakeError = new Error('I am a fake error');

      process.nextTick(function() {
        mockedStream.emit('error', fakeError);
      });

      stripe.fileUploads.create({
        purpose: 'dispute_evidence',
        file: {
          data: mockedStream,
          name: 'minimal.pdf',
          type: 'application/octet-stream',
        },
      }).catch(function(error) {
        expect(error.message).to.equal('An error occurred while attempting to process the file for upload.');
        expect(error.detail).to.equal(fakeError);

        done();
      });
    });
```

We should also test that you can still provide a (non-Stream) file, which is just the same as what you've already got except with `fs.readFileSync()` instead of `fs.createReadStream()`.

jjsquillante:
@jlomas-stripe awesome, thanks for reviewing again! I'll look over the changes requested and revise based on your comments.

lib/StripeResource.js
jlomas-stripe: (Apparently I meant 'further up' :joy:)

test/flows.spec.js
jlomas-stripe: (I tried doing this in a promisey way but ~I don't trust promises in Mocha~ I couldn't get it to work.  ¯\\\_(ツ)_/¯ :) )

jlomas-stripe:
Awesome - you're welcome!  And I'll try to be more timely in my next review.  :)

lib/resources/FileUploads.js
jjsquillante: oh wow, great catch - my fault - I totally missed that! Thanks

lib/StripeResource.js
jjsquillante: good point! - removed and updated `StreamProcessingError` to FileUploads.

lib/StripeResource.js
jjsquillante: updated

lib/resources/FileUploads.js
jjsquillante: updated - thanks!

test/flows.spec.js
jjsquillante: added both tests (for synchronous file uploads and testing for the stream error to return properly) - thanks!

jjsquillante:
@jlomas-stripe updated the code and revised based on your comments - let me know if there's anything that needs amending. thanks

jlomas-stripe:
@jjsquillante Looks fantastic - thank you! :)

@ob-stripe / @rattrayalex-stripe Can one of y'all please take a rip through this and make sure it all makes sense to you as well?  Thanks!

jlomas-stripe:
Hold on - I didn't actually approve this, just the changes - I'm not sure how that happened.  Weird.

I'd really like @ob-stripe or @rattrayalex-stripe to just have a quick peek first.

jjsquillante:
@jlomas-stripe @ob-stripe @rattrayalex-stripe @brandur-stripe 
Thanks for taking the time to work with me on this PR - I've enjoyed collaborating with you all!

lopesc:
Unsubscribe.

On 15 Jul 2018 21:55, James Squillante <notifications@github.com> wrote:

@jjsquillante commented on this pull request.

________________________________

In lib/resources/FileUploads.js<https://github.com/stripe/stripe-node/pull/471#discussion_r202554230>:

>      data = data || {};

     if (method === 'POST') {
-      return multipartDataGenerator(method, data, headers);
+      return getProcessorForSourceType(data);
     } else {
       return utils.stringifyRequestData(data);


oh wow, great catch - my fault - I totally missed that! Thanks

jlomas-stripe:
No problemo - thanks for taking this on! :)

@ob-stripe can you merge && release?  Or ... can ... I? :joy:

ob-stripe:
I'll take :)

ob-stripe:
Released as 6.3.0.

PR #473, opened 2018-06-27
Add support for the PaymentIntent resource

remi-stripe:
This feature is currently gated so the tests are stubbed for now.

r? @ob-stripe 
cc @stripe/api-libraries 
cc @jenan-stripe @michelle-stripe


ob-stripe:
Released as 6.2.0.

PR #474, opened 2018-06-28
Adds support for an optional 'partner_id' parameter in 'setAppInfo'

zachwick:
This adds support for an optional partner_id attribute in the SetAppInfo hash.

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -21,7 +21,7 @@ Stripe.USER_AGENT = {
 
 Stripe.USER_AGENT_SERIALIZED = null;
 
-var APP_INFO_PROPERTIES = ['name', 'version', 'url'];
+var APP_INFO_PROPERTIES = ['name', 'version', 'url', 'partner_id'];
 
 var EventEmitter = require('events').EventEmitter;
 var exec = require('child_process').exec;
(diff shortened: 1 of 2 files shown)


ob-stripe:
This looks good, but I wonder if the parameter should be called `partnerId` in the Node API (but still be serialized as `partner_id`).

cc @jlomas-stripe @brandur-stripe

zachwick:
@ob-stripe @jlomas-stripe @brandur-stripe I'm open to changing the parameter name to whatever to match whatever conventions we already have in `stripe-node`. I used `partner_id` since that it what was used in most of the other Stripe API client libraries.

brandur-stripe:
Hm, good points. Node's kind of a weird one because camel case is definitely the language convention, but we recommend that for this library's API methods, the plain snakecase parameter names be sent. For example:

``` js
      stripe.applePayDomains.create({
        domain_name: 'example.com',
      });
```

I could see the argument though that `setAppInfo` is more like a function than an API call, and its arguments are really arguments rather than API parameters, and therefore should take camelcase naming. I could go either way, but I'd _probably_ lean toward leaving it as is here.

@jlomas-stripe Thoughts?

remi-stripe:
@brandur-stripe @ob-stripe Just wanted to check back in on this PR!

PR #477, opened 2018-07-04
Extra comma in includeBasic

MarkHerhold:
Pretty sure this is a typo which adds an empty element to the `includeBasic` array.


MarkHerhold:
My mistake. No empty element is added.

PR #482, opened 2018-07-27
Add ScheduledQueryRun resource

remi-stripe:
r? @jlomas-stripe @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -83,6 +83,9 @@ var resources = {
     Transactions: require('./resources/Issuing/Transactions'),
     Disputes: require('./resources/Issuing/Disputes'),
   }),
+  Sigma: resourceNamespace('sigma', {
+    ScheduledQueryRuns: require('./resources/Sigma/ScheduledQueryRuns'),
+  }),
 };
 
 Stripe.StripeResource = require('./StripeResource');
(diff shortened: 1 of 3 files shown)


jlomas-stripe:
Looks awesome to me.

lib/stripe.js
```diff
@@ -83,6 +83,9 @@ var resources = {
     Transactions: require('./resources/Issuing/Transactions'),
     Disputes: require('./resources/Issuing/Disputes'),
   }),
+  Sigma: resourceNamespace('sigma', {
+    ScheduledQueryRuns: require('./resources/Sigma/ScheduledQueryRun'),
```
brandur-stripe: It's kind of annoying, but I think the convention in Node is to have the filenames plural. Can we make this `ScheduledQueryRuns`?

brandur-stripe:
Minor comment, but looks good!

lib/stripe.js
remi-stripe: ah nice catch on that one. I just fixed it if you can take a second look

brandur-stripe:
Thanks for the fixes! LGTM.

brandur-stripe:
Released as 6.5.0.

PR #483, opened 2018-08-02
Add support for the File Link resource

remi-stripe:
r? @ob-stripe @brandur-stripe 
cc @stripe/api-libraries


lib/stripe.js
```diff
@@ -42,46 +44,45 @@ var resources = {
...
   TransferReversals: require('./resources/TransferReversals'),
 
   // Namespaced resources
   Issuing: resourceNamespace('issuing', {
     Authorizations: require('./resources/Issuing/Authorizations'),
     Cardholders: require('./resources/Issuing/Cardholders'),
     Cards: require('./resources/Issuing/Cards'),
-    Transactions: require('./resources/Issuing/Transactions'),
     Disputes: require('./resources/Issuing/Disputes'),
+    Transactions: require('./resources/Issuing/Transactions'),
```
brandur-stripe: Hah, thanks for alphabetizing everything!

test/resources/FileLinks.spec.js
```diff
@@ -0,0 +1,54 @@
...
+      });
+    });
+  });
+
+  describe('update', function() {
+    it('Sends the correct request', function() {
+      stripe.fileLinks.update('fl_123', {metadata: {key: 'value'}});
+      expect(stripe.LAST_REQUEST).to.deep.equal({
+        method: 'POST',
+        url: '/v1/file_links/fl_123',
```
brandur-stripe: Quite minor and I didn't notice on your PHP pull, but file link tokens have a `test` or `live` in them. This is unlike most resources, but is needed because you're not using a key that would hint which mode to load it in.

It might be slightly more accurate if we used an identifier like `fl_live_123`.

brandur-stripe:
Minor comment, but otherwise LGTM.

test/resources/FileLinks.spec.js
ob-stripe: Wait, that's only for the `url` attribute though. File links have a regular token, but it starts with `link_`, not with `fl_`.

brandur-stripe: Oh oops. Yeah you're right — I got confused seeing the `fl_`.

remi-stripe: totally misread that from an earlier link I saw that had fl in it. Thanks for catching, I'll also fix in PHP

remi-stripe:
@brandur-stripe Fixed PTAL

brandur-stripe:
LGTM.

brandur-stripe:
Released as 6.6.0.

PR #485, opened 2018-08-03
Add cancel support for topups

foresthu-stripe:
Add support for the new cancel topup endpoint

r? @remi-stripe || @ob-stripe 
cc @stripe/api-libraries @miguel-stripe @kenneth-stripe @kimberly-stripe

### DIFF
--- a/lib/resources/Topups.js
+++ b/lib/resources/Topups.js
@@ -1,8 +1,15 @@
 'use strict';
 
 var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
 
 module.exports = StripeResource.extend({
   path: 'topups',
   includeBasic: ['create', 'list', 'retrieve', 'update', 'setMetadata', 'getMetadata'],
+
+  cancel: stripeMethod({
+    method: 'POST',
+    path: '{topupId}/cancel',
+    urlParams: ['topupId'],
+  }),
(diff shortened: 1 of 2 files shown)


lib/resources/Topups.js
```diff
@@ -1,8 +1,15 @@
...
 var StripeResource = require('../StripeResource');
+var stripeMethod = StripeResource.method;
 
 module.exports = StripeResource.extend({
   path: 'topups',
   includeBasic: ['create', 'list', 'retrieve', 'update', 'setMetadata', 'getMetadata'],
+
+  cancel: stripeMethod({
+    method: 'POST',
+    path: '{topupId}/cancel',
```
remi-stripe: The test seem to confirm it works but other files have a leading `/` in the path.

remi-stripe:
Added a minor comment though it seems to work. Re-assigning to @brandur-stripe to double check and then release!

lib/resources/Topups.js
brandur-stripe: Hm, we seem to be a little inconsistent on this right now — looking through other examples in the library, it seems that both work.

brandur-stripe:
LGTM.

brandur-stripe:
Released as 6.7.0.

PR #492, opened 2018-08-27
Fix test suite for latest API version

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

The test suite is broken because it tries to cancel a subscription with `at_period_end=true`, which is no longer possible with the latest API version. Fixed by changing the request to an update request with `cancel_at_period_end=true`.

### DIFF
--- a/test/flows.spec.js
+++ b/test/flows.spec.js
@@ -121,7 +121,7 @@ describe('Flows', function() {
       });
     });
 
-    it('Allows me to: subscribe then cancel with `at_period_end` defined', function() {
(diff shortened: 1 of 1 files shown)


ob-stripe:
> Shouldn't the test suite set an API version to avoid this?

Yes, but really the test suite shouldn't use the API at all.

remi-stripe:
Well that does not change the issue in that case. If we used stripe-mock it would have the exact same validation. At least in theory, it does not properly validate params on DELETE today because of a bug but it will in the future.

Issue #225, opened 2016-01-20
Automatic pagination with .list functions


joeycozza:
I was wondering, are there built in options for automatic pagination using the node module?

brandur:
We recently got automatic pagination implemented in Ruby, Java, and Python, but unfortunately Node's is still outstanding. It's certainly something we want to do, but I don't think anyone's currently working on it.

brandur:
I'm going to rename this issue slightly and tag it with "future" so that we have an easy place to track this.

TrejGun:
:+1: for implementation of this feature

coffy:
+1

tejasmanohar:
:+1:

brandur:
Would anyone be willing to propose an interface for this one?

Mostly due to the inherit complexity produced by Node's callbacks, I'm having a bit of trouble coming up with something's that's significantly easier to use than what you'd end up implementing yourself. Ideally we'd want something that could iterate through every result transparently and only make fetches for new pages when necessary, but it seems to me that this would require a callback of the form `(err, obj, next)` or `(err, page, next)` that seems like it would be a major hassle to use.

It's also possible to build a list function that aggregates through _every_ result before trigger a callback, but we've avoided this in other language because it's quite easy to produce enough objects to overwhelm a single process.

I think [coroutines](http://bluebirdjs.com/docs/api/promise.coroutine.html) would present a pretty ideal pagination mechanism, but unfortunately the majority of our Node users seem to be on versions < 0.12.

Would definitely love to hear some ideas, especially from anyone more familiar with idiomatic Node than myself. Thanks!

matthewarkin:
The first thing that comes to mind is something like a stream but streams is node are just ugh.

My other though is like how underscore has a .each so something like 
```
stripe.charges.all( function (err,charge, next/done){
    doSomethingWithCharge();
    done();
})
```
but you'd then need to add like an argument you could send through to done that could stop the pagination which seems a bit hacky and horrible to keep track of...

matthewarkin:
I have used async before to do something similar using queues https://github.com/caolan/async#queue
basically the queue has a worker that processes items in the queue, when the queue is empty add more items to it

osuushi:
Generators/coroutines would be pretty much the perfect interface for this. I'd love to have that, but if most API users are still on older versions of node, I guess it doesn't help much. It would be fantastic to have as an optional add on (e.g. `require('stripe/es6')`) though.

brandur:
[...]

PR #496, opened 2018-09-11
Auto-Pagination

rattrayalex-stripe:
Addresses https://github.com/stripe/stripe-node/issues/225

Work remaining: 

- [x] Add tests for a variety of edge-cases, including:
  - [x] Actually reaching the end of a list (where `has_more: false`).
  - [x] Errors being thrown in a variety of places being handled correctly.
  - [x] Attempts to use more than one API at the same time (eg; having a `.then` on the `.list()` _and_ its `.autoPagingEach()`, passing a callback _and_ attempting to use the result in an iterator, etc). 
- [x] Try to verify that I added `autoPageable` to all the places that I should have (I just grepped `list:`). 
- [x] Try to come up with a `.autoPagingToArray` helper of some kind. 

cc @ob-stripe @jlomas-stripe @brandur-stripe  – this is ready for preliminary eyes if you're interested. I included some refactor work, so viewing commit-by-commit may aid your sanity.


jlomas-stripe:
> Try to come up with a .autoPagingToArray helper of some kind.

What does that mean?  Is this to have some kind of 'output' from the whole thing?  If so, I think it's probably not worth doing, tbh; if someone wants to collect 'em all, they can do so, but per comments in 225, I think it's a bit risky for us to keep track of this - and I don't think any of the other libraries do either?

rattrayalex-stripe:
Yeah, I feel that. Pretty on the fence myself.

I'm basically picturing a lot of users writing their own wrappers to do exactly this, though, which makes me cringe a bit, especially given that it's a lot less straightforward/pleasant than in any of the other languages. 

Plus, for-loops and iteratively pushing into an array have (somewhat rightfully) fallen out of style in the JS community, in favor of `.filter`/`.map`-style code, which without a helper is unachievable here.

As a first pass implementation, I have: 
```js
      requestPromise.autoPagingToArray = function(onDone) {
        var promise = new Promise(function(resolve) {
          var items = [];
          requestPromise.autoPagingEach(function(item) {
            items.push(item);
          }).then(function() {
            resolve(items);
          });
        });
        return self.wrapTimeout(promise, onDone);
      }
```

... though I'm leaning towards a mandatory `{max: number}` arg, perhaps with a cap of 10,000 or something, to limit the downside of runaway memory use / api abuse. 

Frankly, looking at the above code, I think we'd ~need to either include a copy/pastable version of that in our docs, or just include the helper directly.

rattrayalex-stripe:
Okay, added an implementation of `autoPagingToArray` for easier discussion.

Using the tests as an example, comparing the ~simplest case of `autoPagingEach`: 
https://github.com/stripe/stripe-node/blob/173eb650dcd0c0086cc5a20f1e82cd3ee1cf2909/test/autoPagination.spec.js#L52-L63
to the equivalent with `autoPagingToArray`: 
https://github.com/stripe/stripe-node/blob/173eb650dcd0c0086cc5a20f1e82cd3ee1cf2909/test/autoPagination.spec.js#L141-L145

And also looking at the 10-line boilerplate wrapper we'd need to recommend for casual use, I'm personally leaning towards including this in some form. My impression is that none of our other languages have such non-idiomatic/ugly code required to for happy-path usecases. 

I'm assuming here that a large percentage of paginated list requests against the Stripe API have fewer than 10k results (eg; ~small users poking around at ~small collections), and that 10k results could easily fit within memory & rate limits. I'm also assuming that users ~frequently open up a node console to poke around at things in an exploratory fashion.

Definitely open to thoughts on this, and bikeshedding re; the API. These are just my hunches.

rattrayalex-stripe:
r? @brandur-stripe @jlomas-stripe 
@ob-stripe & @remi-stripe , eager for your 👀 as well!

jlomas-stripe:
Hey @rattrayalex-stripe!  I'll try to take a peek at this today or tomorrow.  Thanks so much for taking this on! \o/

jlomas-stripe:
@rattrayalex-stripe I think the test for "lets you call `next()` to iterate and `next(false)` to break" (and its siblings) should call `next(false)` before you reach `LIMIT` to confirm you can exit; you'll also probably want to have a test to confirm that once it gets to `LIMIT` that `next()` won't be called (and `done()` will).

rattrayalex-stripe:
sweet, thanks @jlomas-stripe !

re; `LIMIT`, that's a great catch – thanks! Will clean that up. 

Separately, re; moving `testUtils/` to be a sibling of `test`, that was because mocha's `--recursive` mode will try to parse _everything_ in `test/`, which breaks on most node versions when you have `for await` syntax in there. `--recursive` doesn't have any way to `exclude` files; the recommended backup is to use a glob, but I found that changing `--recursive` in `mocha.opts` to `test/**/*.spec.js` meant that `npm run mocha -- test/myTest.spec.js` didn't work anymore; it would always run all the tests, since the globs were additive. 😞 

I could have moved the `.node*.js` files into a sibling and left `testUtils.js` in the same place, but felt it'd be a bit confusing to have both. Don't have strong feels though. Thoughts?

rattrayalex-stripe:
By the way, @romain-stripe had suggested `limit:` instead of `max:` for `autoPagingToArray`. I figured we should first decide whether we really want `autoPagingToArray`, but also wanted to raise that bikeshed for discussion (I'm ambivalent).

lib/utils.js
```diff
@@ -227,6 +227,19 @@ var utils = module.exports = {
     }
     return false;
   },
+
+  callbackifyPromise: function(promise, callback) {
```
brandur-stripe: Isn't the old name for this (`wrapTimeout`) better? It basically seems to take a callback and put it in a timeout block.

lib/resources/Accounts.js
```diff
@@ -14,6 +14,7 @@ module.exports = StripeResource.extend({
   list: stripeMethod({
     method: 'GET',
     path: 'accounts',
+    autoPageable: true,
```
brandur-stripe: So the idea is that we still need to add this parameter to a few dozen other list calls elsewhere right?

The `autoPageable` parameter seems a little unusual to me given that (1) only lists are pageable, and (2) all lists should be pageable. What do you think about naming this something like `methodType: 'list'` or something along those lines instead?

lib/autoPagination.js
```diff
@@ -0,0 +1,190 @@
+'use strict';
+
+var makeRequest = require('./makeRequest');
+var utils = require('./utils');
+
+function getItemCallback(args) {
```
brandur-stripe: Do you think that we could make an effort to organize this file a little bit more? Either:

1. Functions listed alphabetically.
2. Functions listed by category with exported on the top (so you drop into the file and see the most important ones first) and unexported below, and alphabetically within each category.

lib/autoPagination.js
```diff
@@ -0,0 +1,190 @@
...
+  // `.autoPagingEach((item, next) => { doSomething(item); next(); });`
+  if (onItem.length === 2) {
+    return onItem;
+  }
+
+  if (onItem.length > 2) {
+    throw Error('The `onItem` callback function passed to autoPagingEach must accept at most two arguments; got ' + onItem);
+  }
+
+  // API compat; turn this:
```
brandur-stripe: Could you add a little more flavor to this comment? What API is this trying to be compatible with?

lib/autoPagination.js
```diff
@@ -0,0 +1,190 @@
...
+      }
+
+      if (i < listResult.data.length) {
+        var value = listResult.data[i];
+        i += 1;
+        return {value: value, done: false};
+      } else if (listResult.has_more) {
+        // Reset counter, request next page, and recurse.
+        i = 0;
+        listPromise = requestNextPage(listResult);
```
brandur-stripe: Is there any potential harm to going to an arbitrary stack depth like this? Like if we iterated a thousand pages, do we have a thousand functions sitting on the stack and the potential to overflow? I know some languages do tail call optimization for this kind of situation, and it looks like newer versions of JS have it, but not sure if we should expect it to kick in here or not.

lib/autoPagination.js
```diff
@@ -0,0 +1,190 @@
...
+          return asyncIteratorNext().then(handleIteration);
+        }
+      });
+    }
+
+    asyncIteratorNext().then(handleIteration).catch(reject);
+  });
+}
+
+function autoPagingEach(self, requestArgs, spec, firstPagePromise) {
```
brandur-stripe: Since these are new methods and fairly non-obvious to figure out how to use based on the function signature, could we try adding some basic documentation blocks for `autoPagingEach` and `autoPagingToArray`?

lib/resources/Accounts.js
rattrayalex-stripe: Yeah, looks like I missed methods like `listTransactions`, etc – looks like there's 13 of those. 

I like `methodType: 'list'` a little better too, let's go with that. My thinking was "what if there's a list that isn't pageable" but that's pretty silly.

lib/autoPagination.js
rattrayalex-stripe: sure, I like (2)!

It's a bit of a smell in JS to call functions before they're defined, imo (it's not possible with arrow functions like `const fn = () =>`, only function declarations like we have here) but given that we're likely to stay with ES5 source code for some time, I'm happy to make a silver lining out of it.

lib/autoPagination.js
rattrayalex-stripe: Hmm, interesting. I wasn't sure about this but it looks like recursive promises are handled by the promise implementation, _not_ the call stack: 

> That recursive call is asynchronous and so it does not abuse the call stack by definition.

Which I got from [this post](https://alexn.org/blog/2017/10/11/javascript-promise-leaks-memory.html), ironically about how the promise implementation leaks memory (fortunately, at rates which don't seem problematic – I think it's not until you get well into the millions at a minimum).

So I think we're good here; thanks for checking

lib/autoPagination.js
rattrayalex-stripe: ✅

lib/autoPagination.js
rattrayalex-stripe: ✅

lib/resources/Accounts.js
rattrayalex-stripe: ✅

lib/autoPagination.js
rattrayalex-stripe: Sure; for users or maintainers? I assume users?

lib/utils.js
rattrayalex-stripe: hmm, it happens to use a timeout to achieve a secondary goal of letting a promise instead call a callback, right?

rattrayalex-stripe:
Actually, in writing docs for this (which I started in docstrings but would like to move to the readme), I'm leaning towards moving from this:

```js
for await (const customer of stripe.customers.list().autoPagingEach()) {}
```
to this:
```js
for await (const customer of stripe.customers.list()) {}
```

which will enable getting rid of a hack or two and should be cleaner for users. 

I have a WIP on that but will need to finish on Monday.

jlomas-stripe:
@rattrayalex-stripe How would that work, exactly?  Is it, like ... `list()` just becomes the auto-paging thing?

lib/utils.js
brandur-stripe: Hm, agreed, but it seems like the timeout piece is kind of important here. `callbackifyPromise` reads like a promise utility function or something to me. Maybe `callbackifyPromiseWithTimeout`?

lib/autoPagination.js
brandur-stripe: K, thanks for looking into that.

brandur-stripe:
> @rattrayalex-stripe How would that work, exactly? Is it, like ... list() just becomes the auto-paging thing?

Was wondering the same thing! :) The invocation certainly looks incredibly clean if we can get there.

lib/utils.js
rattrayalex-stripe: SGTM, will do on Monday

rattrayalex-stripe:
yeah, instead of `return utils.callbackifyPromiseWithTimeout(promise, callback)` we'll do something like: 

```
const ret = utils.callbackifyPromiseWithTimeout(promise, callback);
if (isList) {
  const asyncIterationNext = factoryThing();
  ret.autoPagingEach = foo(asyncIterationNext)
  ret.autoPagingToArray = bar(asyncIterationNext);
  ret.next = asyncIterationNext;
  ret.return = () => ({});
  ret[Symbol.asyncIterator] = () => ret;
}
return ret;
```

rattrayalex-stripe:
FWIW, I ran a back-of-napkin `stripe.charges.list().autoPagingToArray({limit: 10000})` and it completed in ~10 minutes (643306ms) with a resulting object around 20MB. That's not coming very close to anyone's memory limit so I'd feel pretty comfortable raising it to a million or something, but it's also slow enough that if you're doing something that big, you're probably better off parallelizing.

README.md
```diff
@@ -216,6 +216,87 @@ stripe.setAppInfo({
 
 This information is passed along when the library makes calls to the Stripe API.
 
+### Auto-pagination
+
+As of stripe-node 6.11.0, you may auto-paginate list api methods.
```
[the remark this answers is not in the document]
brandur-stripe: Yeah probably. We can always fix it later too!

brandur-stripe: Can you capitalize to "API" given this is user facing?

README.md
```diff
@@ -216,6 +216,87 @@ stripe.setAppInfo({
...
+We provide a few different API's for this to aid with a variety of node versions and styles.
```
brandur-stripe: Sorry the nit, but I think we usually use "APIs" (see [examples from Google](https://www.google.ca/search?q=apis+site%3Astripe.com&oq=apis+site%3Astripe.com&aqs=chrome..69i57.2805j0j7&sourceid=chrome&ie=UTF-8)). There's a little disagreement, but without an apostrophe is usually considered the right way to pluralize an acronym.

README.md
```diff
@@ -216,6 +216,87 @@ stripe.setAppInfo({
...
+
+
+#### Async iterators (`for-await-of`)
+
+If you are in a node environment that has support for [async iteration](https://github.com/tc39/proposal-async-iteration#the-async-iteration-statement-for-await-of),
```
brandur-stripe: Can we pluralize "Node" here like it is below?

brandur-stripe:
Awesome. Left a few minor comments on README wording, but LGTM.

README.md
rattrayalex-stripe: I assume you meant capitalize 😄 but yes, good catch!

README.md
rattrayalex-stripe: nice, yes, thanks!

README.md
rattrayalex-stripe: hmm, actually I think I'll just remove the word since it's used differently than the next sentence: `you may auto-paginate list methods`. sgty?

rattrayalex-stripe:
Awesome, thanks for the eagle-eyes @brandur-stripe ! Updated. Does the content feel good? 

@jlomas-stripe that's a very good point... I think we should really add docs at `/docs/api#pagination` to show off the pattern of spinning up multiple month-at-a-time paginators in parallel. I think @remi-stripe has shown this pattern off a few times. Can you think of a clarification to the this readme too? I'm coming up dry 😕

README.md
brandur-stripe: Haha, oops, yes.

README.md
brandur-stripe: Yep! SG.

brandur-stripe:
> Awesome, thanks for the eagle-eyes @brandur-stripe ! Updated. Does the content feel good?

Yep, looks great! We can always iterate if we find something missing/unclear later.

rattrayalex-stripe:
Great, thanks!

Sounds like this is approved; marking as such. Will merge & release this afternoon.

PR #499, opened 2018-09-18
Override host per method instead of per resource

ob-stripe:
r? @brandur-stripe @jlomas-stripe @remi-stripe 
cc @stripe/api-libraries 

Replaces the `overrideHost` key in `StripeResource` with a `host` key in the `stripeMethod` spec argument. This allows for overriding the host per method instead of per resource.

### DIFF
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -14,8 +14,9 @@ var makeAutoPaginationMethods = require('./autoPagination').makeAutoPaginationMe
  *  must be passed by the consumer of the API. Subsequent optional arguments are
  *  optionally passed through a hash (Object) as the penultimate argument
  *  (preceding the also-optional callback argument
-  * @param [spec.encode] Function for mutating input parameters to a method.
+ * @param [spec.encode] Function for mutating input parameters to a method.
  *  Usefully for applying transforms to data on a per-method basis.
+ * @param [spec.host] Hostname for the request.
  */
 function stripeMethod(spec) {
   return function() {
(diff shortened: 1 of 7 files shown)


lib/StripeMethod.js
```diff
@@ -13,15 +13,17 @@ var OPTIONAL_REGEX = /^optional!/;
...
  *  Usefully for applying transforms to data on a per-method basis.
+ * @param [spec.host] Hostname for the request.
  */
 function stripeMethod(spec) {
   var commandPath = typeof spec.path == 'function' ? spec.path
     : utils.makeURLInterpolator(spec.path || '');
   var requestMethod = (spec.method || 'GET').toUpperCase();
   var urlParams = spec.urlParams || [];
   var encode = spec.encode || function(data) {return data;};
+  var requestHost = spec.host;
```
brandur-stripe: Should we just call this `host` for consistency?

lib/StripeMethod.js
```diff
@@ -13,15 +13,17 @@ var OPTIONAL_REGEX = /^optional!/;
  *  must be passed by the consumer of the API. Subsequent optional arguments are
  *  optionally passed through a hash (Object) as the penultimate argument
  *  (preceding the also-optional callback argument
-  * @param [spec.encode] Function for mutating input parameters to a method.
+ * @param [spec.encode] Function for mutating input parameters to a method.
  *  Usefully for applying transforms to data on a per-method basis.
+ * @param [spec.host] Hostname for the request.
```
brandur-stripe: It's great that we have some docs for the this. Thanks for updating!

ob-stripe:
> I wonder if it might be worth allowing both a 'resource level override' and a 'request level override'? 

I considered doing just that, but decided against it to keep the implementation simple. I don't anticipate that we will need this for anything other than the file create endpoint.

jlomas-stripe:
Makes sense to me. :+1:

PR #500, opened 2018-09-19
Rename `FileUploads` to `Files`

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

- Rename `stripe.fileUploads` to `stripe.files` (but keep `stripe.fileUploads` as an alias for BC)
- Have the retrieve and list requests hit `api.stripe.com` instead of `uploads.stripe.com`
- Have the create request hit `files.stripe.com` instead of `uploads.stripe.com` (this is actually the same host, but `files.stripe.com` is the new canonical name)

### DIFF
--- a/lib/resources/Files.js
+++ b/lib/resources/Files.js
@@ -71,18 +71,16 @@ module.exports = StripeResource.extend({
     headers: {
       'Content-Type': 'multipart/form-data',
     },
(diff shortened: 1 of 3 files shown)


brandur-stripe:
LGTM!

ob-stripe:
Released as 6.12.0.

PR #498, opened 2018-09-17
Bindings for Terminal endpoints

daz-stripe:
Bindings for Terminal endpoints
- Connection Tokens
- Locations
- Readers

PR #502, opened 2018-09-24
Fix tests

ob-stripe:
cc @stripe/api-libraries 

Not quite sure what happened, but it looks like #498 broke the test suite somehow (despite tests in the PR being green). This should fix it.


ob-stripe:
Released as 6.12.1.

PR #507, opened 2018-10-12
Add support for the Person resource

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/test/resources/Account.spec.js
+++ b/test/resources/Account.spec.js
@@ -231,4 +231,129 @@ describe('Account Resource', function() {
       });
     });
   });
+
+  describe('Person methods', function() {
+    describe('retrievePerson', function() {
+      it('Sends the correct request', function() {
+        stripe.account.retrievePerson('acct_123', 'person_123');
+        expect(stripe.LAST_REQUEST).to.deep.equal({
+          method: 'GET',
+          url: '/v1/accounts/acct_123/persons/person_123',
+          headers: {},
+          data: {},
+        });
+      });
+
+      it('Sends the correct request [with specified auth]', function() {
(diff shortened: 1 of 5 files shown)


test/resources/Persons.spec.js
```diff
@@ -0,0 +1,84 @@
...
+
+// Use spy from existing resource:
+person._request = stripe.customers._request;
+
+describe('Person Resource', function() {
+  describe('create', function() {
+    it('Sends the correct request', function() {
+      person.create({
+        first_name: "John",
+      });
```
remi-stripe: This part makes no sense to me. I know it works and why but that feels like a really weird way to approach this test and I'm ~confused why we even allow this.

jlomas-stripe: What part makes no sense?

remi-stripe: The part where to create a brand new object (which means said object does not exist yet) you use a pre-filled fake object.

test/resources/Account.spec.js
```diff
@@ -231,4 +231,116 @@ describe('Account Resource', function() {
...
+        stripe.account.updatePerson('acct_123', 'person_123', {
+          first_name: 'John',
+        });
+        expect(stripe.LAST_REQUEST).to.deep.equal({
+          method: 'POST',
+          url: '/v1/accounts/acct_123/persons/person_123',
+          headers: {},
+          data: {first_name: 'John'},
+        });
+      });
```
ob-stripe: Missing a test for update + specified auth.

remi-stripe:
@ob-stripe PTAL

ob-stripe:
Released as 6.13.0.

Issue #511, opened 2018-10-18
Feature Request: Promise-based autoPagingEach API


tyscorp:
I love the new `autoPagingEach` functionality but it would be cool if the API was a bit more intuitive. The callback would either wait for a promise resolving to continue iteration or for a promise rejection to exit.

e.g.
```js
stripe.customers.list().autoPagingEach(async function onItem(customer) {
  const result = await doSomething(customer);
  if (shouldStop(result)) {
    throw new Error('Stop the thing');
  }
});
```

I've been wrapping the iteratees with a helper function[0] to enable this, but it would be cool to have first-class support. This implementation does swallow all errors, though... Maybe have it rethrow the error? Or maybe have a specific error class to exit but not bubble the error?

[0]
```js
const promisifyAutopager = (iteratee) => async (obj, next) => {
  try {
    await iteratee(obj);
  } catch (e) {
    return next(false);
  }
  next();
};
```

rattrayalex-stripe:
Hey @tyscorp , that's an interesting idea – I'll take a look at it. 

We would want a nice way to break without interfering with good error handling. I'm also a bit hesitant to make the api even more overloaded; the docs are already quite complex. 

That said, I do think the example you provide sounds compelling. 

What do you think about returning `false` from an async function (or, equivalently, returning a promise which resolves to false) to break?

rattrayalex-stripe:
It turns out we already have support for this! See https://github.com/stripe/stripe-node/pull/512

Thanks @tyscorp ! Let me know if you have any issues or further feedback. Closing for now but feel free to re-open.

tyscorp:
Lol that's awesome. Cheers! 👍

PR #512, opened 2018-10-18
Clarify `autoPagingEach` async usage

rattrayalex-stripe:
r? @brandur-stripe 
cc @ob-stripe @remi-stripe @jlomas-stripe 

In https://github.com/stripe/stripe-node/issues/511, @tyscorp asked for support for the following (paraphrased slightly): 

```js
stripe.customers.list().autoPagingEach(async function onItem(customer) {
  const result = await doSomething(customer);
  if (shouldStop(result)) {
    return false;
  }
});
```

It turns out we already have support for this! The only thing better than inventing a new feature is discovering one. 

I have added a test to verify that this works and prevent regressions, and have added docs to clarify. 

Separately, I have also cleaned up the examples on the home page to better reflect current Node usage (that is, a little more modern).

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ callback:
 // Create a new customer and then a new charge for that customer:
 stripe.customers.create({
   email: 'foo-customer@example.com'
-}).then(function(customer){
+}).then(function(customer) {
   return stripe.customers.createSource(customer.id, {
     source: 'tok_visa'
   });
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -74,23 +74,23 @@ const stripe = new Stripe('sk_test_...');
 Every method returns a chainable promise which can be used instead of a regular
 callback:
 
-``` js
+```js
 // Create a new customer and then a new charge for that customer:
 stripe.customers.create({
   email: 'foo-customer@example.com'
-}).then(function(customer){
+}).then((customer) => {
```
brandur-stripe: These are pretty nice :)

brandur-stripe:
LGTM.

PR #519, opened 2018-11-13
mark change as non-breaking

avimar:


remi-stripe:
Hey @avimar! Thanks a lot for the PR and I agree that we could have made this clearer in the changelog. Would you mind tweaking the wording a bit to clarify that we aliased but still recommend moving to the new resource instead as it will likely be removed in the next major release?

Something like this maybe:

> Rename `FileUploads` to `Files`. For backwards compatibility, `Files` is aliased to `FileUploads`. `FileUploads` is deprecated and will be removed from the next major version.

remi-stripe:
Thanks for the quick fix! Approved and merging!

PR #523, opened 2018-11-23
Handle Buffer instances in Webhook.constructEvent

ob-stripe:
r? @jlomas-stripe 
cc @stripe/api-libraries 

Handle `Buffer` instances in `Webhook.constructEvent`, by silently converting them to strings.

Firebase returns the raw body as a `Buffer` instance, so this lets users directly pass `request.rawBody` to the `constructEvent` method, without having to do the string conversion themselves.

(What happens currently is that the buffer is coerced into a string when we compute the signed payload with the timestamp, but since no explicit encoding is applied, this can apparently cause issues in some cases, cf. https://github.com/stripe/stripe-node/issues/341#issuecomment-441278062).

### DIFF
--- a/lib/Webhooks.js
+++ b/lib/Webhooks.js
@@ -1,16 +1,18 @@
+'use strict';
+
+var Buffer = require('safe-buffer').Buffer;
 var crypto = require('crypto');
 
 var utils = require('./utils');
 var Error = require('./Error');
 
 var Webhook = {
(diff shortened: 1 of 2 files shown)


jlomas-stripe:
lgtm

👍

ob-stripe:
Released as 6.15.1.

Issue #521, opened 2018-11-21
Subscription update method mutates items


jperasmus:
Hi there,

Thanks for this awesome library and for Stripe in general. A-grade 👏

### Using:
```
node v10.9.0
stripe v6.15.0
```

When performing an update of a subscription's `items` I've discovered that the `stripe.subscriptions.update()` method mutates the payload provided to it by transforming the `items` from an array to an object with the array indexes as the object keys. This can lead to very unexpected behaviour when reusing the same payload after the update method was invoked.

I've managed to trace the problem to this utility method:

```javascript
/**
   * Encodes a particular param of data, whose value is an array, as an
   * object with integer string attributes. Returns the entirety of data
   * with just that param modified.
   */
  encodeParamWithIntegerIndexes: function(param, data) {
    if (data[param] !== undefined) {
      data[param] = utils.arrayToObject(data[param]);
    }
    return data;
},
```

Ideally a new transformed `data` object should be returned instead of mutating it. Something like:

```javascript
const cloneDeep = require('lodash.cloneDeep');

/**
   * Encodes a particular param of data, whose value is an array, as an
   * object with integer string attributes. Returns the entirety of data
   * with just that param modified.
   */
  encodeParamWithIntegerIndexes: function(param, payload) {
    const data = cloneDeep(payload);
    if (data[param] !== undefined) {
      data[param] = utils.arrayToObject(data[param]);
    }
    return data;
},
```

The problem would be for any of the other methods that use this utility method and not just for `stripe.subscriptions.update()`.

Is this something you would be interested in fixing? I would be happy to create a PR.

jperasmus:
Hi @ob-stripe 

Thanks for the reply. The reason I used the lodash `cloneDeep` as an example is that in the same file I saw the `lodash.isplainobject` package used in the same way.

JavaScript, unfortunately, doesn't currently have natively functionality to deeply clone an object, so you could either create a small utility yourself or use something like lodash's `cloneDeep` that has been battle tested. By using lodash's method you know it would cover most edge cases that might be missed in your own implementation.

In terms of impact on the bundle size, this might give you more insights to help with the decision: https://bundlephobia.com/result?p=lodash.clonedeep@4.5.0

It is definitely not the smallest package. Other smaller options might be https://bundlephobia.com/result?p=clone-deep@4.0.1

[...]

PR #526, opened 2018-11-26
Do not mutate params when encoding arrays as indexed objects

rattrayalex-stripe:
Fixes https://github.com/stripe/stripe-node/issues/521

Also, for good measure, does a shallow copy of request data in general – I don't think there were any other situations that would have mutated this today, and it also doesn't protect against deep mutations, but I don't think it can hurt much (I wouldn't guess that the extra copy operations would have a meaningful perf impact). 

cc @jperasmus
cc @ob-stripe 
cc @enugent-stripe

### DIFF
--- a/lib/makeRequest.js
+++ b/lib/makeRequest.js
@@ -55,7 +55,7 @@ function getRequestOpts(self, requestArgs, spec, overrideData) {
 
   // Pull request data and options (headers, auth) from args.
   var dataFromArgs = utils.getDataFromArgs(args);
-  var data = encode(Object.assign(dataFromArgs, overrideData));
+  var data = encode(Object.assign({}, dataFromArgs, overrideData));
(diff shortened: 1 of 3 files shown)


jlomas-stripe:
lgtm 👍

jperasmus:
Thanks for the quick fix @rattrayalex-stripe 

I see you went for the shallow copy approach. In this situation, it looks to be a good solution because the `encodeParamWithIntegerIndexes` utility method only applies the `arrayToObject` transformation one level deep.

rattrayalex-stripe:
@jperasmus - yep, sorry, should have addressed directly; for now, a shallow copy is fine since we don't (can't, actually, I think) modify deep properties (please correct me if I'm wrong!). 

If/when we add support for "encoding" deeply nested values, it'll require a code change – I'll be sure to check that we're more careful with our inputs then 😄

PR #530, opened 2018-12-05
Export errors on root Stripe object

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

Export errors on root Stripe object, as suggested in https://github.com/stripe/stripe-node/issues/457#issuecomment-441856552.

Also export webhooks because why not.


ob-stripe:
Released as 6.18.0.

PR #534, opened 2018-12-07
Fix iterating on files.list method

### LINKED ISSUE #533: "files" entity missing list (and other) methods
Hello!

First, versions:
**Node version**: 10.13.0
**stripe-node version**: 6.18.0

OK, so I currently work on a [gatsby-source plugin](https://github.com/njosefbeck/gatsby-source-stripe) for Stripe data. When attempting to call `stripe.files.list` as explained in your docs [here](https://stripe.com/docs/api/files/list) I was getting an error that the method wasn't iterable. I found this strange, as all other list methods are iterable (using and LOVING the new for-await-of pagination btw). So I did further digging.

When inspecting the stripe object, I noticed that the list method for "files" seems to be missing:

screen shot 2018-12-07 at 10 46 26 am

And, actually, now that I look at that screenshot more closely, it looks like it's also missing the other methods (like create and retrieve).

Is there any way that this can be fixed? Or, is there a different way I should be trying to access stripe.files.list?

Happy to do further testing if you'd all like! Thanks again for the awesome node library!

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

The `retrieve` and `list` methods were manually defined on `files` because they used to hit a different host (`files.stripe.com` vs. `api.stripe.com`). Now `files.stripe.com` is only used for upload and download requests and `api.stripe.com` is used for retrieving / listing file objects, so we can use the standard methods.

Fixes #533.

### DIFF
--- a/lib/resources/Files.js
+++ b/lib/resources/Files.js
@@ -65,6 +65,7 @@ module.exports = StripeResource.extend({
   },
 
   path: 'files',
+  includeBasic: ['list', 'retrieve'],
 
   create: stripeMethod({
     method: 'POST',
@@ -72,15 +73,5 @@ module.exports = StripeResource.extend({
       'Content-Type': 'multipart/form-data',
     },
     host: 'files.stripe.com',
-  }),
-
-  list: stripeMethod({
-    method: 'GET',
-  }),
-
-  retrieve: stripeMethod({
-    method: 'GET',
-    path: '/{id}',
-    urlParams: ['id'],
   }),
 });


rattrayalex-stripe:
Nice! This, will work for all users, correct? Assuming so, 

LGTM

ob-stripe:
Yep, the retrieve and list endpoints are accessible via both hosts regardless of API version or anything.

Thanks Alex!

ob-stripe:
Released as 6.18.1.

PR #551, opened 2019-01-17
Move CheckoutSession to the Checkout namespace and rename to Session

remi-stripe:
The class `CheckoutSession` is now removed. Instead, we introduce the `Checkout` namespace and the resource `Session`.

Since the resource is in beta and has a limited number of users, we are making this a "non-breaking change" as we will actively warn anyone using it. The old endpoints will return a useful error message recommending to upgrade the library.

cc @stripe/api-libraries @matt-stripe


remi-stripe:
r? @ob-stripe 

This is now ready to be released

PR #552, opened 2019-01-17
Fix Buffer deprecation warnings

ob-stripe:
r? @remi-stripe @rattrayalex-stripe 
cc @stripe/api-libraries 

Using `new Buffer` is deprecated, cf. https://nodejs.org/api/buffer.html#buffer_new_buffer_array. This PR replaces uses of `new Buffer` with `Buffer.alloc(size)` or `Buffer.from(string)` as appropriate.

### DIFF
--- a/lib/MultipartDataGenerator.js
+++ b/lib/MultipartDataGenerator.js
@@ -8,12 +8,12 @@ var Buffer = require('safe-buffer').Buffer;
 function multipartDataGenerator(method, data, headers) {
   var segno = (Math.round(Math.random() * 1e16) + Math.round(Math.random() * 1e16)).toString();
   headers['Content-Type'] = ('multipart/form-data; boundary=' + segno);
-  var buffer = new Buffer(0);
+  var buffer = Buffer.alloc(0);
 
   function push(l) {
     var prevBuffer = buffer;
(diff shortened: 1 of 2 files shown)


remi-stripe:
this LGTM but deferring to @rattrayalex-stripe who is more familiar with node

rattrayalex-stripe:
Sadly it [seems](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array) that both `Buffer.from` and `Buffer.alloc` were added in Node v5.10.0. It seems we [claim](https://github.com/stripe/stripe-node/blob/master/package.json#L25) support for Node 4. 

I'd be in favor of supporting Node 6+ and dropping support for Node 4 and Node 5, which would of course require a major version bump. We could also check whether `Buffer.from` is defined before using it, etc.

ob-stripe:
@rattrayalex-stripe We use the [`safe-buffer`](https://www.npmjs.com/package/safe-buffer) package to backport those methods to early Node versions.

We should probably drop support for Node < 6 at some point, but we don't need to for this specific PR.

rattrayalex-stripe:
Ah, terrific! Sorry I missed that. 

LGTM

ob-stripe:
No worries, thanks Alex!

ob-stripe:
Released as 6.20.1.

PR #555, opened 2019-01-23
Support OAuth flow for Connect accounts

### LINKED ISSUE #422: Authenticate express account tokens documentation?
How would I make this request using the stripe-node library?
https://stripe.com/docs/connect/express-accounts#token-request

I can make an http request using the curl params from the link above but I just wanted to make sure I wasn't missing something.

robz-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Fixes #422


lib/StripeResource.js
```diff
@@ -23,7 +23,7 @@ function StripeResource(stripe, urlData) {
   this._stripe = stripe;
   this._urlData = urlData || {};
 
-  this.basePath = utils.makeURLInterpolator(stripe.getApiField('basePath'));
+  this.basePath = utils.makeURLInterpolator(this.basePath || stripe.getApiField('basePath'));
```
robz-stripe: This enables support for endpoints that don't begin with `/v1`. In this case, that's `/oauth/token`.

lib/StripeResource.js
```diff
@@ -40,6 +40,9 @@ StripeResource.prototype = {
 
   path: '',
 
+  // Methods that don't use the API's default '/v1' path can override it with this setting.
+  basePath: null,
```
robz-stripe: We default to the API's `/v1` prefix unless it's overridden when `extend`-ing `StripeResource`.

robz-stripe:
Tested this via the steps at https://stripe.com/docs/connect/standard-accounts and the access code redeemed correctly for a token.

ob-stripe:
Hey @robz-stripe, thanks for taking a stab at this!

This looks like a great start. Before we merge this however, I think we need some additional features for bring this to parity with our other libraries:
- a utility `authorizeURL` method for creating the `https://connect.stripe.com/oauth/authorize?...` redirection URL ([doc](https://stripe.com/docs/connect/standard-accounts#integrating-oauth))
    - this method should have an option for Express accounts, to change the resulting URL to `https://connect.stripe.com/express/oauth/authorize?...`
- support for deauthorizing accounts via the `/oauth/deauthorize` endpoint ([doc](https://stripe.com/docs/connect/standard-accounts#revoked-access))

If you haven't already, I recommend you take a look at the implementations in other languages, particularly stripe-ruby's: https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/oauth.rb. (If I'm not mistaken, stripe-ruby is the only one that supports Express accounts right now.)

Also, very minor nit, but would you mind changing the capitalization from `Oauth` to `OAuth`?

lib/stripe.js
```diff
@@ -152,6 +153,7 @@ Stripe.errors = require('./Error');
 Stripe.webhooks = require('./Webhooks');
 
 Stripe.prototype = {
+  clientId: '',
```
robz-stripe: This could potentially be moved into an OAuth helper similar to `this._setApiField`, but that may be overkill for a single field.

lib/stripe.js
```diff
@@ -21,6 +21,10 @@ Stripe.USER_AGENT = {
 
 Stripe.USER_AGENT_SERIALIZED = null;
 
+Stripe.ERROR_CLIENT_ID_NOT_SET = 'Please set stripe.clientId or pass client_id as a parameter' +
```
robz-stripe: This seems like a logical place to store this, since `Error.js` is specifically for API errors. Open to opinions.

robz-stripe:
r? @ob-stripe

This implements `authorizeUrl` and `deauthorize` as well, now.

robz-stripe:
Making a note here to update https://github.com/stripe/stripe-node/wiki/Using-Stripe-Connect-with-node.js once the interface is finalized.

lib/stripe.js
ob-stripe: No strong feels, but it seems kind of odd to have a single error message lying here.

In other libraries, this check and error were added to be consistent with the API key check. Since (AFAICT) stripe-node doesn't have a check for the API key, maybe the right thing to do here is to remove the client_id check too and let the server return an error.

lib/stripe.js
ob-stripe: My preference would go towards exposing the client ID similarly to the API key, i.e. via a `setClientId` method.

lib/stripe.js
robz-stripe: I'm OK with leaving it to the server to return the error when the generated URL is used. I'll revert this change unless anyone has a strong feeling that we should keep it.

lib/stripe.js
robz-stripe: Although OAuth exists on a separate host from the API proper, would it make more sense to have the `clientId` stored in `_api` object or put it as a property on its own `_oauth` object?

lib/resources/OAuth.js
```diff
@@ -0,0 +1,56 @@
...
+var stripeMethod = StripeResource.method;
+var utils = require('../utils');
+
+var oAuthHost = 'connect.stripe.com';
+
+module.exports = StripeResource.extend({
+  basePath: '/',
+
+  authorizeUrl: function(params, options) {
+    // As long as we support Node v4, we have to set defaults like this
```
rattrayalex-stripe: Should probably remove comments like this from public repos 😄

lib/resources/OAuth.js
```diff
@@ -0,0 +1,56 @@
...
+  deauthorize: function(spec) {
+    if (!spec.client_id) {
+      spec.client_id = this._stripe.getClientId();
+    }
+
+    return stripeMethod({
+      method: 'POST',
+      path: 'oauth/deauthorize',
+      host: oAuthHost,
+    }).apply(this, arguments);
```
rattrayalex-stripe: I don't love that we have to do this, but don't know of a better way. 👍

test/resources/OAuth.spec.js
```diff
@@ -0,0 +1,177 @@
...
+
+var stripe = require('../../testUtils').getSpyableStripe();
+
+var expect = require('chai').expect;
+
+describe('OAuth', function() {
+  describe('authorize', function() {
+    describe('when a default client_id is not set', function() {
+      beforeEach(function() {
+        stripe.setClientId('');
```
rattrayalex-stripe: I think the default resting state of the world is for it to be `undefined`; should we `stripe.setClientId()` instead?

test/resources/OAuth.spec.js
```diff
@@ -0,0 +1,177 @@
...
+
+          expect(url).to.equal(expectedUrl);
+        });
+
+        it('Generates the correct URL with state provided', function() {
+          var url = stripe.oAuth.authorizeUrl({state: 'some_state'});
+
+          var expectedUrl = 'https://connect.stripe.com/oauth/authorize?state=some_state&response_type=code&client_id=default_client_id&scope=read_write';
+
+          expect(url).to.equal(expectedUrl);
```
rattrayalex-stripe: These tests are a bit hard to read/verify, and I can imagine someone in the future "fixing" the tests in a way that hides a bug. 

Could we get a bit more granular? I'm thinking something like this:

```js
var URL = require('url');
var qs = require('qs');

// in this test
var state = qs.parse(URL.parse(url).query).state;
expect(state).to.equal('some_state');

// in other tests
var pathname = URL.parse(url).pathname;
expect(pathname).to.equal('/express/oauth/authorize');
```

test/resources/OAuth.spec.js
```diff
@@ -0,0 +1,177 @@
...
+              scope: 'read_write',
+            }
+          );
+
+          var expectedUrl = 'https://connect.stripe.com/oauth/authorize?response_type=code&client_id=123abc&scope=read_write';
+
+          expect(url).to.equal(expectedUrl);
+        });
+
+        it('Generates the correct URL with state provided', function() {
```
rattrayalex-stripe: This doesn't seem to be providing state?

lib/resources/OAuth.js
robz-stripe: Ha, fair enough. I see how that could be read as a knock. :)

robz-stripe:
Cleaned up the tests to be more granular and simpler to grok. Let me know what you think!

samheutmaker:
Thanks @robz-stripe!

Nisthar:
@robz-stripe is there an example to do oauth with this lib?

noah-witt:
Any way to get this documented in the API?

remi-stripe:
@Nisthar You can find examples in the repo here: https://github.com/stripe/stripe-node/blob/master/test/resources/OAuth.spec.js

@noah-witt Yes we're working on adding examples to our docs!

PR #557, opened 2019-01-28
Add Stripe client telemetry to request headers

jameshageman-stripe:
r? @brandur-stripe @ob-stripe 
cc @stripe/api-libraries @dcarney @akropp-stripe 

Replicates https://github.com/stripe/stripe-go/pull/766, adding request duration metrics to subsequent requests in the `X-Stripe-Client-Telemetry` header.

Since node.js is concurrent but only runs on one thread, I've used a buffering approach similar to the go implementation but without explicit mutual exclusion. Right now the buffer is unbounded, so it's size is proportional to the maximum number of successful, concurrent requests.

Instead of mocking out the server, I spin up a real http server on a random port for each test, and add assertions on the server side. This may be unconventional, but it allows me to be absolutely sure that the metrics are reaching the server. I used a similar approach in the [Go implementation](https://github.com/jameshageman-stripe/stripe-go/blob/d87ed19be3d9d9f25e26ee112a1f3a9ed6c98415/stripe_test.go#L229).

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -174,7 +174,7 @@ StripeResource.prototype = {
           );
         }
 
-        if (res.requestId) {
+        if (self._stripe.telemetryEnabled() && res.requestId) {
           self._stripe._prevRequestMetrics.push({
             'request_id': res.requestId,
             'request_duration_ms': requestDuration,
@@ -258,7 +258,7 @@ StripeResource.prototype = {
           Object.assign(headers, options.headers);
         }
 
-        if (self._stripe._prevRequestMetrics.length > 0) {
+        if (self._stripe.telemetryEnabled() && self._stripe._prevRequestMetrics.length > 0) {
           var metrics = self._stripe._prevRequestMetrics.pop();
           headers['X-Stripe-Client-Telemetry'] = JSON.stringify({
             'last_request_metrics': metrics
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -125,6 +125,8 @@ StripeResource.prototype = {
         // lastResponse.
         res.requestId = headers['request-id'];
 
+        var requestDuration = Date.now() - req._requestStart;
+
         var responseEvent = utils.removeEmpty({
           api_version: headers['stripe-version'],
           account: headers['stripe-account'],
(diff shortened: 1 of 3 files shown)


lib/StripeResource.js
```diff
@@ -125,6 +125,8 @@ StripeResource.prototype = {
         // lastResponse.
         res.requestId = headers['request-id'];
 
+        var requestDuration = Date.now() - req._requestStart;
```
dcarney: could you make this variable include the units in the name?  (e.g. suffix it with "ms")?

lib/stripe.js
```diff
@@ -299,12 +302,19 @@ Stripe.prototype = {
     return formatted;
   },
 
+  setTelemetryEnabled: function(enableTelemetry) {
+    this._enableTelemetry = enableTelemetry;
+  },
+
+  telemetryEnabled: function() {
```
dcarney: It looks like the convention in the rest of this file is to use get/set prefixes on the function names, so I'd change this to `getTelemetryEnabled`.

dcarney:
Made a couple small comments, but I'll defer to the lib's maintainers for the final approval.  👍

jameshageman-stripe:
@brandur could you take a look at this one?

lib/StripeResource.js
```diff
@@ -171,6 +173,14 @@ StripeResource.prototype = {
             null
           );
         }
+
+        if (self._stripe.getTelemetryEnabled() && res.requestId) {
+          self._stripe._prevRequestMetrics.push({
+            'request_id': res.requestId,
+            'request_duration_ms': requestDurationMs,
+          });
+        }
```
brandur-stripe: Like for stripe-python, can we put these into a `_addTelemetryHeader` and `_recordRequestMetrics` or such?

brandur-stripe: This unbounded buffer worries me a little bit, do you think that we could have a check against some arbitrarily large size (e.g. `> 100`) and if we get there, don't push on a new value and then use `utils.emitWarning` to indicate that we dropped a request result? (So that we can tell it's happening.)

lib/StripeResource.js
```diff
@@ -248,6 +258,13 @@ StripeResource.prototype = {
           Object.assign(headers, options.headers);
         }
 
+        if (self._stripe.getTelemetryEnabled() && self._stripe._prevRequestMetrics.length > 0) {
+          var metrics = self._stripe._prevRequestMetrics.pop();
```
brandur-stripe: Should we use `shift` instead of `pop` so that the buffer becomes FIFO?

jameshageman-stripe:
I've made the buffer both bounded and FIFO
ptal @brandur-stripe

brandur-stripe:
LGTM! Thanks.

brandur-stripe:
Released as 6.23.0.

Issue #558, opened 2019-01-28
Feature request: Idempotent retries on API calls


suchitagarwal:
In our stripe integration, we sometimes get api errors (StripeAPIError or StripeConnectionError), which are most likely just transient issues and should go away if the API call is just retried. Implementing this on top of the nodejs library right now is hard because we have to manually hijack every single call to the stripe client and implement the retry ourselves.

It'd be great if when setting up the stripe client, I could specify a number of retries, and a max timeout for each api call. 
- As a v1, this could be done just for GET requests, so that the retry logic does not have to worry about idempotency
- Bonus feature request: All API requests can be retried, and are guaranteed to be idempotent as appropriate.

remi-stripe:
We should introduce retry logic as we did in [stripe-php](https://github.com/stripe/stripe-php/pull/428) for example. Assigning to @paulasjes-stripe who mentioned he was interested and tagging as future for now

PR #559, opened 2019-01-30
Network retries

paulasjes-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

Fixes https://github.com/stripe/stripe-node/issues/558

Adds network retry functionality. The default retry amount is 0 but can be changed via `stripe.setMaxNetworkRetries(n)`. I modeled this after a similar change in `stripe-php`: https://github.com/stripe/stripe-php/pull/428/files

In the case of a `POST` request, an idempotency key is generated if not already provided.

This also adds the `uuid` and `nock` libraries. The former is for generating idempotency keys, there already is a similar function used in testing however according to our docs \[0\] `uuid/v4` should ideally be used for real requests. I use `nock` to mock API responses and failures.

\[0\]: https://stripe.com/docs/api/idempotent_requests

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ const stripe = Stripe('sk_test_...');
 
 ### Usage with TypeScript
 
-Stripe does not currently maintain typings for this package, but there are 
-community typings available from DefinitelyTyped. 
+Stripe does not currently maintain typings for this package, but there are
+community typings available from DefinitelyTyped.
 
 To install:
 
@@ -144,6 +144,17 @@ if (process.env.http_proxy) {
   stripe.setHttpAgent(new ProxyAgent(process.env.http_proxy));
 }
 ```
+
+### Network retries
+
+Network retry amount can be added with `setMaxNetworkRetries`. This will automatically retry a request `n` amount of times if it failed due to a connection, conflict or rate limiting error.
+
+```js
+// Retry a request once before giving up
+stripe.setMaxNetworkRetries(1);
+```
+
+For POST requests, this will automatically add a [idempotency key](https://stripe.com/docs/api/idempotent_requests) if one is not already set to prevent duplication.
 
 ### Examining Responses
 
--- a/README.md
+++ b/README.md
@@ -147,14 +147,12 @@ if (process.env.http_proxy) {
 
 ### Network retries
 
-Network retry amount can be added with `setMaxNetworkRetries`. This will automatically retry a request `n` amount of times if it failed due to a connection, conflict or rate limiting error.
+Automatic network retries can be enabled with `setMaxNetworkRetries`. This will retry requests `n` times with exponential backoff if they fail due to connection, conflict or rate limiting errors. [Idempotency keys](https://stripe.com/docs/api/idempotent_requests) are added where appropriate to prevent duplication.
 
 ```js
 // Retry a request once before giving up
 stripe.setMaxNetworkRetries(1);
 ```
-
-For POST requests, this will automatically add a [idempotency key](https://stripe.com/docs/api/idempotent_requests) if one is not already set to prevent duplication.
 
 ### Examining Responses
 
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -206,14 +206,25 @@ StripeResource.prototype = {
   },
 
   _shouldRetry: function(res, numRetries) {
-    // Retry on connection error, if we have retries left.
-    if (!res) {
-      return numRetries < this._stripe.getMaxNetworkRetries();
+    // Do not retry if we are out of retries.
+    if (numRetries >= this._stripe.getMaxNetworkRetries()) {
+      return false;
     }
 
-    // Retry on conflict or rate limit error, if we have retries left.
-    if (res.statusCode === 409 || res.statusCode === 429) {
-      return numRetries < this._stripe.getMaxNetworkRetries();
+    // Retry on connection error.
+    if (!res) {
+      return true;
+    }
+
+    // Retry on conflict, rate limit, and availability errors.
+    if (res.statusCode === 409 || res.statusCode === 429 || res.statusCode === 503) {
+      return true;
+    }
+
+    // Retry on 5xx's, except POST's, which our idempotency framework
+    // would just replay as 500's again anyway.
+    if (res.statusCode >= 500 && res.req._requestEvent.method !== 'POST') {
+      return true;
     }
 
     return false;
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -200,13 +200,18 @@ StripeResource.prototype = {
     }
   },
 
-  _shouldRetry: function(err, numRetries) {
+  _shouldRetry: function(error, numRetries) {
     if (numRetries >= this._stripe.getMaxNetworkRetries()) {
       return false;
     }
 
-    if (err.type === 'StripeConnectionError' || err.type === 'StripeApiError') {
+    if (error.type === 'StripeConnectionError' || error.type === 'StripeApiError') {
       return true;
+    }
+
+    // conflict or rate limiting
+    if (error.code === 409 || error.code === 429) {
+      return false;
     }
 
     return false;
@@ -328,6 +333,13 @@ StripeResource.prototype = {
       req.on('error', function(error) {
         if (self._shouldRetry(error, requestRetries)) {
           requestRetries += 1;
+
+          // if it's a POST request, add an idempotency key
+          if (method === 'POST') {
+            if (!headers.hasOwnProperty('Idempotency-Key')) {
+              headers['Idempotency-Key'] = utils.getUUID();
+            }
+          }
 
           return setTimeout(
             makeRequest,
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -206,12 +206,12 @@ StripeResource.prototype = {
   },
 
   _shouldRetry: function(res, numRetries) {
-    // retry on conflict or rate limit error, if we have retries left
+    // Retry on conflict or rate limit error, if we have retries left.
     if (res && (res.statusCode === 409 || res.statusCode === 429)) {
       return numRetries < this._stripe.getMaxNetworkRetries();
     }
 
-    // retry on connection error, if we have retries left
+    // Retry on connection error, if we have retries left.
     if (!res && (numRetries < this._stripe.getMaxNetworkRetries())) {
       return true;
     }
@@ -231,10 +231,10 @@ StripeResource.prototype = {
       maxNetworkRetryDelay
     );
 
-    // Apply some jitter
+    // Apply some jitter.
     sleepSeconds *= 0.5 * (1 + Math.random());
 
-    // Never sleep more than the max allowed
+    // Never sleep more than the max allowed.
     sleepSeconds = Math.max(initialNetworkRetryDelay, sleepSeconds);
 
     return sleepSeconds;
@@ -301,13 +301,6 @@ StripeResource.prototype = {
     function retryRequest(requestFn, apiVersion, headers, requestRetries) {
       requestRetries += 1;
 
-      // if it's a POST request, add an idempotency key
-      if (method === 'POST') {
-        if (!headers.hasOwnProperty('Idempotency-Key')) {
-          headers['Idempotency-Key'] = uuid();
-        }
-      }
-
       return setTimeout(
         requestFn,
         self._getSleepTime(requestRetries) * 1000,
@@ -332,6 +325,14 @@ StripeResource.prototype = {
         headers: headers,
         ciphers: 'DEFAULT:!aNULL:!eNULL:!LOW:!EXPORT:!SSLv2:!MD5',
       });
+
+      // If this is a POST and we allow multiple retries, set a idempotency key if one is not
+      // already provided.
+      if (method === 'POST' && self._stripe.getMaxNetworkRetries() > 0) {
+        if (!headers.hasOwnProperty('Idempotency-Key')) {
+          headers['Idempotency-Key'] = uuid();
+        }
+      }
 
       var requestEvent = utils.removeEmpty({
         api_version: apiVersion,
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -182,19 +182,6 @@ StripeResource.prototype = {
     };
   },
 
-  _parseResponse: function(res, cb) {
-    var response = '';
-
-    res.setEncoding('utf8');
-    res.on('data', function(chunk) {
-      response += chunk;
-    });
-
-    res.on('end', function() {
-      cb(response);
-    });
-  },
-
   _errorHandler: function(req, requestRetries, callback) {
     var self = this;
     return function(error) {
@@ -213,13 +200,13 @@ StripeResource.prototype = {
     }
   },
 
-  _shouldRetry: function(error, numRetries) {
+  _shouldRetry: function(res, numRetries) {
     if (numRetries >= this._stripe.getMaxNetworkRetries()) {
       return false;
     }
 
     // conflict or rate limiting
-    if (error.code === 409 || error.code === 429) {
+    if (res && (res.statusCode === 409 || res.statusCode === 429)) {
       return false;
     }
 
@@ -305,6 +292,14 @@ StripeResource.prototype = {
       makeRequestWithData(null, utils.stringifyRequestData(data || {}));
     }
 
+    function addIdempotentKey(headerObj) {
+      if (!headerObj.hasOwnProperty('Idempotency-Key')) {
+        headerObj['Idempotency-Key'] = utils.getUUID();
+      }
+
+      return headerObj;
+    }
+
     function makeRequest(apiVersion, headers, numRetries) {
       var timeout = self._stripe.getApiField('timeout');
       var isInsecureConnection = self._stripe.getApiField('protocol') == 'http';
@@ -338,41 +333,35 @@ StripeResource.prototype = {
       self._stripe._emitter.emit('request', requestEvent);
 
       req.setTimeout(timeout, self._timeoutHandler(timeout, req, callback));
-      // req.on('response', self._responseHandler(req, callback));
-      req.on('reponse', function(res) {
-        self._parseResponse(res, function(parsedResponse) {
-          if (self._shouldRetry(parsedResponse, requestRetries)) {
-            requestRetries += 1;
 
-            // if it's a POST request, add an idempotency key
-            if (method === 'POST') {
-              if (!headers.hasOwnProperty('Idempotency-Key')) {
-                headers['Idempotency-Key'] = utils.getUUID();
-              }
-            }
-
-            return setTimeout(
-              makeRequest,
-              self._getSleepTime(requestRetries) * 1000,
-              apiVersion,
-              headers,
-              requestRetries
-            );
-          } else {
-            return self._responseHandler(req, callback)(res);
-          }
-        });
-      });
-
-      req.on('error', function(error) {
-        if (self._shouldRetry(error, requestRetries)) {
+      req.on('response', function(res) {
+        if (self._shouldRetry(res, requestRetries)) {
           requestRetries += 1;
 
           // if it's a POST request, add an idempotency key
           if (method === 'POST') {
-            if (!headers.hasOwnProperty('Idempotency-Key')) {
-              headers['Idempotency-Key'] = utils.getUUID();
-            }
+            headers = addIdempotentKey(headers);
+          }
+
+          return setTimeout(
+            makeRequest,
+            self._getSleepTime(requestRetries) * 1000,
+            apiVersion,
+            headers,
+            requestRetries
+          );
+        } else {
+          return self._responseHandler(req, callback)(res);
+        }
+      });
+
+      req.on('error', function(error) {
+        if (self._shouldRetry(null, requestRetries)) {
+          requestRetries += 1;
+
+          // if it's a POST request, add an idempotency key
+          if (method === 'POST') {
+            headers = addIdempotentKey(headers);
           }
 
           return setTimeout(
@@ -402,7 +391,6 @@ StripeResource.prototype = {
       });
     }
   },
-
 };
 
 module.exports = StripeResource;
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -206,20 +206,20 @@ StripeResource.prototype = {
   },
 
   _shouldRetry: function(res, numRetries) {
-    // Retry on conflict or rate limit error, if we have retries left.
-    if (res && (res.statusCode === 409 || res.statusCode === 429)) {
+    // Retry on connection error, if we have retries left.
+    if (!res) {
       return numRetries < this._stripe.getMaxNetworkRetries();
     }
 
-    // Retry on connection error, if we have retries left.
-    if (!res && (numRetries < this._stripe.getMaxNetworkRetries())) {
-      return true;
+    // Retry on conflict or rate limit error, if we have retries left.
+    if (res.statusCode === 409 || res.statusCode === 429) {
+      return numRetries < this._stripe.getMaxNetworkRetries();
     }
 
     return false;
   },
 
-  _getSleepTime: function(numRetries) {
+  _getSleepTimeInMS: function(numRetries) {
     var initialNetworkRetryDelay = this._stripe.getInitialNetworkRetryDelay();
     var maxNetworkRetryDelay = this._stripe.getMaxNetworkRetryDelay();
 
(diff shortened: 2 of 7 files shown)


lib/StripeResource.js
```diff
@@ -258,7 +297,16 @@ StripeResource.prototype = {
       makeRequestWithData(null, utils.stringifyRequestData(data || {}));
     }
 
-    function makeRequest(apiVersion, headers) {
+    // add idempotency key if we're retrying the request
```
remi-stripe: I think this comment is confusing because it describes the behaviour when the function is called, not what the function does. It also implies we only add the header on the retry and not on the first attempt (which we should not do since it could cause 2 separate requests to be sent)

lib/StripeResource.js
```diff
@@ -192,14 +192,53 @@ StripeResource.prototype = {
...
+    // conflict or rate limiting
+    if (res && (res.statusCode === 409 || res.statusCode === 429)) {
+      return false;
+    }
+
+    // don't retry if everything succeeded
+    if (res && res.statusCode === 200) {
+      return false;
+    }
+
```
remi-stripe: Does this mean we retry 400 and 500 errors too? We should not, a 400 is a bad parameter error or similar and retrying won't help (+ it will also add delay). I think we should whitelist what makes a retry instead of always retry unless a specific behaviour happened

lib/StripeResource.js
```diff
@@ -282,15 +330,57 @@ StripeResource.prototype = {
...
-      req.on('response', self._responseHandler(req, callback));
-      req.on('error', self._errorHandler(req, callback));
+
+      req.on('response', function(res) {
+        if (self._shouldRetry(res, requestRetries)) {
+          requestRetries += 1;
+
+          // if it's a POST request, add an idempotency key
+          if (method === 'POST') {
+            headers = addIdempotentKey(headers);
```
remi-stripe: Okay this goes with my other comment, if you add the idempotency key on retry, you cause 2 requests to be sent.

lib/StripeResource.js
```diff
@@ -282,15 +330,57 @@ StripeResource.prototype = {
...
 
       req._requestStart = Date.now();
 
       self._stripe._emitter.emit('request', requestEvent);
 
       req.setTimeout(timeout, self._timeoutHandler(timeout, req, callback));
-      req.on('response', self._responseHandler(req, callback));
-      req.on('error', self._errorHandler(req, callback));
+
+      req.on('response', function(res) {
```
robz-stripe: This function and the one for `req.on('error' ...` share a lot of code. Might want to combine them into one functions that's reusable.

lib/StripeResource.js
rattrayalex-stripe: Further comment nits, not really related: 

1) function docstrings should use block comment format (just `/**`, not like `@param` stuff). 
2) Comments should be in sentence case (first letter caps) with a period at the end (many comments in this PR... to be fair, I bet a lot of our comments don't follow this).

lib/StripeResource.js
```diff
@@ -192,14 +192,53 @@ StripeResource.prototype = {
...
     }
   },
 
+  _shouldRetry: function(res, numRetries) {
+    if (numRetries >= this._stripe.getMaxNetworkRetries()) {
+      return false;
+    }
+
+    // conflict or rate limiting
+    if (res && (res.statusCode === 409 || res.statusCode === 429)) {
```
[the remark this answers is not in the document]
paulasjes-stripe: This would retry on 429 with exponential backoff. PHP doesn't have it yet but [will soon I believe](https://github.com/stripe/stripe-php/pull/447).

lib/StripeResource.js
paulasjes-stripe: I misunderstood how idempotency works. Should be fixed now though!

lib/StripeResource.js
```diff
@@ -192,14 +197,49 @@ StripeResource.prototype = {
       callback.call(
         self,
         new Error.StripeConnectionError({
-          message: 'An error occurred with our connection to Stripe',
+          message: self._generateConnectionErrorMessage(requestRetries),
```
rattrayalex-stripe: nit, let's just inline this since it's a single expr anyway.

lib/StripeResource.js
```diff
@@ -192,14 +197,49 @@ StripeResource.prototype = {
...
 
+  _shouldRetry: function(res, numRetries) {
+    // Retry on conflict or rate limit error, if we have retries left.
+    if (res && (res.statusCode === 409 || res.statusCode === 429)) {
+      return numRetries < this._stripe.getMaxNetworkRetries();
+    }
+
+    // Retry on connection error, if we have retries left.
+    if (!res && (numRetries < this._stripe.getMaxNetworkRetries())) {
+      return true;
```
rattrayalex-stripe: nit, I would move this to the top as:
```js
if (!res) {
  return numRetries < this._stripe.getMaxNetworkRetries();
}
```
and can then remove `res &&` above.

lib/StripeResource.js
```diff
@@ -192,14 +197,49 @@ StripeResource.prototype = {
...
+    var sleepSeconds = Math.min(
+      initialNetworkRetryDelay * Math.pow(numRetries - 1, 2),
+      maxNetworkRetryDelay
+    );
+
+    // Apply some jitter.
+    sleepSeconds *= 0.5 * (1 + Math.random());
+
+    // Never sleep more than the max allowed.
+    sleepSeconds = Math.max(initialNetworkRetryDelay, sleepSeconds);
```
rattrayalex-stripe: The comment seems to say something quite different from what this code seems to do?

lib/StripeResource.js
```diff
@@ -192,14 +197,49 @@ StripeResource.prototype = {
...
+    // Apply exponential backoff with initialNetworkRetryDelay on the
+    // number of numRetries so far as inputs. Do not allow the number to exceed
+    // maxNetworkRetryDelay.
+    var sleepSeconds = Math.min(
+      initialNetworkRetryDelay * Math.pow(numRetries - 1, 2),
+      maxNetworkRetryDelay
+    );
+
+    // Apply some jitter.
+    sleepSeconds *= 0.5 * (1 + Math.random());
```
rattrayalex-stripe: this is a range of roughly 0.5-1, correct? Seems too wide a range to me... also odd to have it not centered at a modification of 1... Maybe +/- 10%?

Ah, I see that this is what we did in stripe-php. That's fine; let's just include the range in the comment, as stripe-php does.

lib/StripeResource.js
```diff
@@ -258,7 +298,19 @@ StripeResource.prototype = {
       makeRequestWithData(null, utils.stringifyRequestData(data || {}));
     }
 
-    function makeRequest(apiVersion, headers) {
+    function retryRequest(requestFn, apiVersion, headers, requestRetries) {
+      requestRetries += 1;
+
+      return setTimeout(
+        requestFn,
+        self._getSleepTime(requestRetries) * 1000,
```
rattrayalex-stripe: nit, I would expect `_getSleepTime` to return ms, not s; mind doing the `* 1000` in its return statement?

lib/StripeResource.js
```diff
@@ -282,15 +342,31 @@ StripeResource.prototype = {
...
+        if (self._shouldRetry(res, requestRetries)) {
+          retryRequest(makeRequest, apiVersion, headers, requestRetries);
+        } else {
+          return self._responseHandler(req, callback)(res);
+        }
+      });
+
+      req.on('error', function(error) {
+        if (self._shouldRetry(null, requestRetries)) {
+          retryRequest(makeRequest, apiVersion, headers, requestRetries);
```
rattrayalex-stripe: we don't need to `return` this? (I'm not that familiar with `req.on('request'` et al)

lib/stripe.js
```diff
@@ -21,6 +21,9 @@ Stripe.USER_AGENT = {
 
 Stripe.USER_AGENT_SERIALIZED = null;
 
+Stripe.MAX_NETWORK_RETRY_DELAY = 2;
+Stripe.INITIAL_NETWORK_RETRY_DELAY = 0.5;
```
rattrayalex-stripe: can we suffix this with `_SEC`?

lib/StripeResource.js
rattrayalex-stripe: Or adopt the control flow that stripe-php uses

lib/StripeResource.js
paulasjes-stripe: This makes for easier testing though as @robz-stripe mentioned in an earlier comment.

rattrayalex-stripe: Ah righto, sgtm

lib/StripeResource.js
paulasjes-stripe: Well the good news is the code is right, but that comment is totally wrong. It should be "Never sleep _less_ than the max allowed".

rattrayalex-stripe: err, you mean, "Never sleep less than the initial delay" ?

lib/StripeResource.js
```diff
@@ -192,14 +197,50 @@ StripeResource.prototype = {
...
+
+    // Retry on conflict or rate limit error, if we have retries left.
+    if (res.statusCode === 409 || res.statusCode === 429) {
+      return numRetries < this._stripe.getMaxNetworkRetries();
+    }
+
+    return false;
+  },
+
+  _getSleepTimeInMS: function(numRetries) {
```
rattrayalex-stripe: FWIW in JS "time" always means ms to me, so I don't think is necessary, but fine to leave.

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+    var options = {
+      host: stripe.getConstant('DEFAULT_HOST'),
+      path: '/v1/charges',
+      data: {
+        amount: 1000,
+        currency: 'usd',
+        source: 'tok_visa',
+        description: 'test'
+      },
+      params: 'amount=1000&currency=usd&source=tok_visa&description=test'
```
rattrayalex-stripe: why are we passing both data and params?

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+    };
+
+    afterEach(function() {
```
rattrayalex-stripe: should this be `beforeEach`?

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+      realStripe.setMaxNetworkRetries(0);
+      stripe.setMaxNetworkRetries(0);
+    });
+
+    after(function() {
+      nock.cleanAll();
+    })
+
+    describe('_request', function() {
+      it('throws an error on connection failure', function(done) {
```
rattrayalex-stripe: is this a connection error, or an api error?

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+          .post(options.path, options.params)
+          .replyWithError('bad stuff');
+
+        realStripe.charges.create(options.data, function(err) {
+          expect(err.detail.message).to.deep.equal('bad stuff');
+          done();
+        });
+      });
+
+      it('should retry the connection if max retries are set', function(done) {
```
rattrayalex-stripe: "should retry the request" ?

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+          .post(options.path, options.params)
+          .replyWithError('bad stuff')
+          .post(options.path, options.params)
+          .replyWithError('worse stuff');
+
+        realStripe.setMaxNetworkRetries(1);
+
+        realStripe.charges.create(options.data, function(err) {
+          var errorMessage = realStripe.invoices._generateConnectionErrorMessage(1);
+          expect(err.message).to.equal(errorMessage);
```
rattrayalex-stripe: should we also check that
```
expect(err.detail.message).to.deep.equal('worse stuff');
```

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+          .post(options.path, options.params)
+          .replyWithError('bad stuff')
+          .post(options.path, options.params)
+          .reply(200, {
+            id: 'ch_123',
+            object: 'charge',
+            amount: 1000,
+          });
+
+        realStripe.setMaxNetworkRetries(2);
```
rattrayalex-stripe: apologies if I don't know `nock` well enough, but is there a guarantee in this test that it did not try to send another request?

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+          .reply(400, {
+            error: {
+              message: 'You goofed'
+            }
+          });
+
+        realStripe.setMaxNetworkRetries(1);
+
+        realStripe.charges.create(options.data, function(err) {
+          expect(err).to.not.be.null;
```
rattrayalex-stripe: Hmm... I think this test could pass erroneously (in theory). Can we explicitly check that it generates the expected error message, instead of some other error? Eg; check that `you goofed` appears in the right place. 

Same with the 500 test below, so we know we're passing along the right message, not blowing up internally

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+          done();
+        });
+      });
+    });
+
+    describe('_shouldRetry', function() {
+      it('should return false if we\'ve reached maximum retries', function() {
+        var res = stripe.invoices._shouldRetry({
+          statusCode: 429
+        }, 0);
```
rattrayalex-stripe: Using the zero case here makes me a bit nervous; let's `stripe.setMaxNetworkRetries(1);` and use `1` here? This _could_ just be testing the case where `stripe.setMaxNetworkRetries` is unset (in theory)

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+        var res = stripe.invoices._shouldRetry({
+          statusCode: 200
+        }, 1);
+
+        expect(res).to.equal(false);
+      });
+    });
+
+    describe('_getSleepTimeInMS', function() {
+      it('should not exceed the maximum or minimum values', function() {
```
rattrayalex-stripe: Call me a cranky fellow, but having randomness/fuzzing in tests like this makes me uncomfortable (even though I expect it to pass in all cases). Imagine we had an exceedingly rare edge-case bug, such as failing when `Math.random()` produces exactly `0.0` - it'd occasionally fail in CI, which is frustrating. Not the end of the world in stripe-node (we can just hit rebuild, and it doesn't block other things from getting released).

To be honest, I'm not sure this test is that important; thoughts on just deleting it?

test/StripeResource.spec.js
rattrayalex-stripe: Ah... This is fine, though might be nice to use `qs` to derive params from data...

test/StripeResource.spec.js
```diff
@@ -32,4 +35,245 @@ describe('StripeResource', function() {
...
+            }
+          });
+
+        realStripe.charges.create(options.data, function(err) {
+          expect(err).to.not.be.null;
+          done();
+        });
+      })
+
+      it('should add an idempotency key for retries using the POST method', function(done) {
```
rattrayalex-stripe: Should we have a test that it doesn't do this for GET's?

lib/StripeResource.js
paulasjes-stripe: I like the idea of being explicit here, might help prevent a bug in the future.

test/StripeResource.spec.js
paulasjes-stripe: So in newer versions of nock (v10 and up) you can pass in an object which the library will automatically parse as a querystring. Unfortunately after downgrading to v9 (for Node v4 compatibility) it looks like I need to provide both a querystring and a data object here.

test/StripeResource.spec.js
paulasjes-stripe: This is more of a clean up step rather than set up, so I opted for `afterEach`.

test/StripeResource.spec.js
paulasjes-stripe: Connection error, which is simulated via nock's `replyWithError` method.

test/StripeResource.spec.js
paulasjes-stripe: Well I think it's still useful to ensure that the sleep time is a value we'd expect. 

I'd prefer a test that very rarely fails over a edge-case bug over never finding the edge-case bug at all. Especially since the `Math.random()` call is in our production code and not in the test. Also it's nice to have more code coverage.

test/StripeResource.spec.js
paulasjes-stripe: Once a 200 response is received it should stop there, if it didn't we'd probably see errors like a charge with an idempotency key being called twice.

brunocascio:
I am facing some issues with duplicated invoices line items and customers creations. This PR will solve it? Now, I am using idempotency_key in all my stripe mutations to avoid the problem.

test/StripeResource.spec.js
rattrayalex-stripe: Hmm, I'm not sure nock would give us that functionality... I wanted to double-check that this throws an error: 
```js
// test
nock(url)
  .get(oneThing)
  .get(anotherThing)

// app
get(url); 
get(url); 
get(url); // Nock should raise an UnmockedRequest error, since only two were mocked.
```
My read of [the docs](https://github.com/nock/nock#read-this---about-interceptors) is that this is the behavior, so I think we're good here 😄

README.md
```diff
@@ -145,6 +145,17 @@ if (process.env.http_proxy) {
 }
 ```
 
+### Network retries
+
+Network retry amount can be added with `setMaxNetworkRetries`. This will automatically retry a request `n` amount of times if it failed due to a connection, conflict or rate limiting error.
```
rattrayalex-stripe: `Automatic network retries can be enabled with` *

I think I would move the bit about idempotency keys to this paragraph, with a few edits:

```
Retries use [idempotency keys](...) where appropriate to prevent duplication and exponential backoff with jitter to help in cases of rate limiting.
```

rattrayalex-stripe: Ah, Paul will be out of the office for a week so I'll just commit these changes.

rattrayalex-stripe:
I'm leaning towards retrying GET requests that fail on 500 errors, too - thoughts, @ob-stripe and @remi-stripe ?

remi-stripe:
I think it'd make sense to do it!

lib/StripeResource.js
```diff
@@ -192,14 +197,50 @@ StripeResource.prototype = {
...
 
+  _shouldRetry: function(res, numRetries) {
+    // Retry on connection error, if we have retries left.
+    if (!res) {
+      return numRetries < this._stripe.getMaxNetworkRetries();
+    }
+
+    // Retry on conflict or rate limit error, if we have retries left.
+    if (res.statusCode === 409 || res.statusCode === 429) {
+      return numRetries < this._stripe.getMaxNetworkRetries();
```
suchitagarwal: quick question:
the two errors we encounter most often when calling stripe that would be helped by using retries are 
- StripeAPIError
- StripeConnectionError

I cant find the http codes for those two, but wondering if this list would cover these two errors. Might be nice to let the client set the statuscodes to retry? i wonder if that'd open up the gates to people starting to retry 400s etc

lib/StripeResource.js
```diff
@@ -274,6 +327,14 @@ StripeResource.prototype = {
         ciphers: 'DEFAULT:!aNULL:!eNULL:!LOW:!EXPORT:!SSLv2:!MD5',
       });
 
+      // If this is a POST and we allow multiple retries, set a idempotency key if one is not
+      // already provided.
```
suchitagarwal: this is wonderful, thanks!

lib/StripeResource.js
rattrayalex-stripe: Hi @suchitagarwal , thanks, that's very helpful to hear!

As it is, the code will _not_ handle `StripeAPIError`, which is status code 500, but _will_ handle `StripeConnectionError`. I'm going to change that to handle 500's too, but only when they are GET's or DELETE's (not POST). 

The reason for this is that there isn't really much point to retrying a 500'd POST request with an idempotency key, because our idempotency machinery will simply give you a 500 again.

I don't think we'll open up the logic for when to automatically retry, pretty much based for reasons like you mentioned (retrying a 400 isn't a great idea), but users can always implement retries themselves. 

Thoughts?

suchitagarwal: makes perfect sense. i think handling 500s for GET/DELETE should cover the cases we encounter most often, and not retrying POSTs makes sense. thanks a lot for the quick turnaround on this

rattrayalex-stripe: Awesome, thanks for the feedback!

Speaking of, I'm working on an unrelated feature and looking for some early beta users... if you're willing to try something out, drop me a line? email is `githubUsername.replace('-', '@') + '.com'` 😉

rattrayalex-stripe:
r? @ob-stripe 

cc @remi-stripe I'm adding 503's and non-POST 500's.

remi-stripe:
I have a little bit of FUD adding the 503 retry logic on non GET. I know we talked about internally, but I'd rather not assume that there are never any side effect to a 503

rattrayalex-stripe:
As discussed IRL, I feel fairly confident that retrying on 503's is the right thing to do - but we'll need to add that logic to all the other libs, so we should discuss as a team in the morning.

lib/StripeResource.js
```diff
@@ -192,14 +197,61 @@ StripeResource.prototype = {
...
+    }
+
+    // Retry on conflict, rate limit, and availability errors.
+    if (res.statusCode === 409 || res.statusCode === 429 || res.statusCode === 503) {
+      return true;
+    }
+
+    // Retry on 5xx's, except POST's, which our idempotency framework
+    // would just replay as 500's again anyway.
+    if (res.statusCode >= 500 && res.req._requestEvent.method !== 'POST') {
```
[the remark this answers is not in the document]
rattrayalex-stripe: From [the docs](https://stripe.com/docs/api/idempotent_requests):

> sending idempotency keys in GET and DELETE requests has no effect and should be avoided, as these requests are idempotent by definition.

retrying DELETE's _should_ be inherently safe/idempotent.

brunocascio:
> Hi @brunocascio - If you were experiencing those issues because you were implementing automatic retries _without_ idempotency keys, then yes, this PR will make it easier to fix that in the future by adding the idempotency keys for you. (Though do note that you will need to configure stripe-node to make use of the automatic retries by setting eg; `stripe.setMaxRetries(5)`).

Hi @rattrayalex-stripe, actually I don't have any retry logic. I am using [serverless framework](https://serverless.com/) and I notice that customers are created multiple times. In order to avoid that, I set an idempotency_key and it works.

Just to clarify, Is it not dangerous to perform retries? What about HTTP timeouts? For example, you could get an HTTP Timeout response but your node function will continue executing retries. I mean, you send an error to the client because about but the request to stripe was ok after some retries (?)

rattrayalex-stripe:
If your node application (express, lambda, etc) times out a request before the stripe promise has resolved (eg; before `stripe.customers.create().then()`), then yes, your application may end up in a bad state, especially if your application's client interprets a request timeout to mean "did not create customer", since the customer may be created. 

Reasons like this are why we won't be enabling this feature by default. To be on the safe side, you would have to keep your server's timeout greater than the number of max retries times 30 seconds, or alternatively build your application to properly handle such timeouts (eg; allow the client to retry with the same idempotency key).

suchitagarwal:
@rattrayalex-stripe is it possible to specify a timeout for the stripe library? what is the default?

brunocascio:
@rattrayalex-stripe Got it, but there are other challenges to resolve with this feature. For example, in order to generate a one-off invoice, you have to create an invoice line and then an invoice since it is not transactional. So, you should have N idempotency keys? How do you know which request needs retry?

rattrayalex-stripe:
@suchitagarwal you can - https://github.com/stripe/stripe-node#configuring-timeout - though you cannot configure the Stripe API's timeout, so this can be dangerous. See discussion also at https://github.com/stripe/stripe-ruby/issues/46 . 

@brunocascio If you are only using this new, built-in retry logic, it handles that for you. If you want to do client-created idempotency keys, things get complicated and you need to think carefully about your application's behavior. One thing you could consider is generate and store a uuid on the client, and then for each request on the backend to stripe, append a consistent string (eg; `_invoice` and `_invoiceitem`). However, we're getting off-topic here 😄 the docs are at https://stripe.com/docs/api/idempotent_requests, and you can ask for help at support.stripe.com or on IRC at #stripe on freenode.

paulasjes-stripe:
@ob-stripe Done! We now mock `_getSleepTimeInMS` so the new tests are pretty much instant.

ob-stripe:
> We now mock `_getSleepTimeInMS` so the new tests are pretty much instant.

Nice! Thanks @paulasjes-stripe.

Reassigning to @brandur-stripe and @rattrayalex-stripe for the final decision regarding the exact retry conditions, but this looks nearly ready to 🚢

brandur-stripe:
Okay, Alex and I had a long talk over lunch following a long talk yesterday, and we've decided to not retry on 429s for the time being, but allowing for a system that eventually does pending a few server-side improvements that we'd like to make.

I just pushed a commit to take 429s out of the retry equation.

r? @rattrayalex-stripe Mind taking a quick look at that? Thanks!

brandur-stripe:
(Also, just wanted to note that I'm really happy to see this come into stripe-node. Thanks for the implementation @paulasjes-stripe!)

remi-stripe:
@rattrayalex-stripe re-assigning to you to release!

ob-stripe:
Can we squash the commits before merging please? :)

rattrayalex-stripe:
Thanks for the reminder - will squash and merge before releasing sometime later today.

PR #560, opened 2019-02-01
Enable persistent connections by default

ob-stripe:
r? @rattrayalex-stripe 

This PR enables persistent connection by default. From my laptop in Europe, running `test/flows.spec.js` takes 33 seconds with this PR, vs. 1 minute with the master branch.

Tagged as WIP, let me know what you think!

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -292,6 +292,7 @@ StripeResource.prototype = {
     function makeRequest(apiVersion, headers) {
       var timeout = self._stripe.getApiField('timeout');
       var isInsecureConnection = self._stripe.getApiField('protocol') == 'http';
+      var agent = isInsecureConnection ? self._stripe.getApiField('http_agent') : self._stripe.getApiField('https_agent');
 
       var req = (
         isInsecureConnection ? http : https
@@ -300,7 +301,7 @@ StripeResource.prototype = {
         port: self._stripe.getApiField('port'),
         path: path,
         method: method,
-        agent: self._stripe.getApiField('agent'),
+        agent: agent,
         headers: headers,
         ciphers: 'DEFAULT:!aNULL:!eNULL:!LOW:!EXPORT:!SSLv2:!MD5',
       });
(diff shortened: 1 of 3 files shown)


lib/stripe.js
```diff
@@ -227,7 +231,11 @@ Stripe.prototype = {
   },
 
   setHttpAgent: function(agent) {
-    this._setApiField('agent', agent);
+    if (agent instanceof https.Agent) {
+      this._setApiField('https_agent', agent);
+    } else {
+      this._setApiField('http_agent', agent);
```
rattrayalex-stripe: I like this approach. Glad to see the docs already tell you to pass an https (not http) agent to this.

lib/stripe.js
```diff
@@ -310,6 +318,12 @@ Stripe.prototype = {
     return this._enableTelemetry;
   },
 
+  buildDefaultAgent: function(protocol) {
```
rattrayalex-stripe: Should this be a prefixed with an underscore? I don't _think_ we want/expect users to call this (instead just `setHttpAgent`)...

lib/StripeResource.js
```diff
@@ -292,6 +292,7 @@ StripeResource.prototype = {
     function makeRequest(apiVersion, headers) {
       var timeout = self._stripe.getApiField('timeout');
       var isInsecureConnection = self._stripe.getApiField('protocol') == 'http';
+      var agent = isInsecureConnection ? self._stripe.getApiField('http_agent') : self._stripe.getApiField('https_agent');
```
rattrayalex-stripe: I assume we have to continue to support `http` for things like dev proxies?

lib/stripe.js
ob-stripe: Yeah, I wasn't sure about this. Let's prefix it.

lib/StripeResource.js
ob-stripe: That, and I'd like to update the test suite to use stripe-mock instead of local stubbing + real API requests at some point. (Although stripe-mock does support HTTPS, if you disable cert verification.)

ob-stripe:
@rattrayalex-stripe I added a test for `setHttpAgent`, but I'm unsure how to go about testing `makeRequest` and that connections are actually reused. Any suggestions?

ob-stripe:
Thanks Alex!

ob-stripe:
Released as v6.32.1.

PR #565, opened 2019-02-13
Always encode arrays as integer-indexed hashes

### LINKED ISSUE #561: GET requests with arrays in data not handled correctly by stripe API
I noticed (and found a workaround to solve) a bug occurring in GET requests with data containing arrays.

For example: `stripe.invoices.retrieve('upcoming', {..., subscription_items: [..., ...] })`

This will try to make a GET request with url-encoded data (in the body of the request apparently(?)), but the converted querystring doesn't* have array indexes.
* purposely, according to the commit found via blame ; and also according to the 'Stripe-specific cases'  test in utils.spec.js)

Thing is: Stripe API seems to need them to understand the query.
Current code in utils.js:
```javascript
  /**
   * Stringifies an Object, accommodating nested objects
   * (forming the conventional key 'parent[child]=value')
   */
  stringifyRequestData: function(data) {
    // console.log('stringifyRequestData', data, qs.stringify(data, {arrayFormat: 'brackets'}))
    return qs.stringify(data, {arrayFormat: 'brackets'});
  },
```
outputs:
```
subscription=sub_ES6MM9DZExqcN0&subscription_prorate=true&subscription_items%5B%5D%5Bid%5D=si_ES6MwymN93s0Fh&subscription_items%5B%5D%5Bdeleted%5D=true&subscription_items%5B%5D%5Bplan%5D=business_monthly&subscription_items%5B%5D%5Bquantity%5D=1&coupon=redacted
```
urldecoded:
```
subscription=sub_ES6MM9DZExqcN0&subscription_prorate=true&subscription_items[][id]=si_ES6MwymN93s0Fh&subscription_items[][deleted]=true&subscription_items[][plan]=business_monthly&subscription_items[][quantity]=1&coupon=redacted
```
Log trace in Stripe Dashboard: 
screenshot at 02-05-39

workaround:
```javascript
  /**
   * Stringifies an Object, accommodating nested objects
   * (forming the conventional key 'parent[child]=value')
   */
  stringifyRequestData: function(data) {
    // console.log('stringifyRequestData', data, qs.stringify(data, {arrayFormat: 'brackets'}))
    return qs.stringify(data);
  },
```
outputs:
```
subscription=sub_ES6MM9DZExqcN0&subscription_prorate=true&subscription_items%5B0%5D%5Bid%5D=si_ES6MwymN93s0Fh&subscription_items%5B0%5D%5Bdeleted%5D=true&subscription_items%5B1%5D%5Bplan%5D=business_monthly&subscription_items%5B1%5D%5Bquantity%5D=1&coupon=redacted
```
urldecoded:
```
subscription=sub_ES6MM9DZExqcN0&subscription_prorate=true&subscription_items[0][id]=si_ES6MwymN93s0Fh&subscription_items[0][deleted]=true&subscription_items[1][plan]=business_monthly&subscription_items[1][quantity]=1&coupon=redacted
```
Log trace in Stripe Dashboard: 
screenshot at 02-06-32

So, either you have to fix it in your API backend (to accept unindexed array backets), or fix it here (to provide API compatible format).

Thanks.

ob-stripe:
r? @brandur-stripe @rattrayalex-stripe 
cc @stripe/api-libraries 

Always encode arrays as integer-indexed hashes. This is the default behavior of `qs.stringify()`. Stripe's API can now accept integer-indexed hashes for all array parameters (including `expand`) so there is no longer a need to special-case arrays of hashes parameters like `subscription_items`.

I've also updated the encoding method to change square brackets back to their literals to make the strings easier to read. This is consistent with most of our other libraries.

Fixes #561.

### DIFF
--- a/test/StripeResource.spec.js
+++ b/test/StripeResource.spec.js
@@ -36,6 +36,64 @@ describe('StripeResource', function() {
     });
   });
 
+  describe('Parameter encoding', function() {
+    // Use a real instance of stripe as we're mocking the http.request responses.
+    var realStripe = require('../lib/stripe')(utils.getUserStripeKey());
+
+    after(function() {
+      nock.cleanAll();
+    })
+
+    describe('_request', function() {
+      it('encodes the query string in GET requests', function(done) {
+        var options = {
+          host: stripe.getConstant('DEFAULT_HOST'),
+          path: '/v1/invoices/upcoming',
+          data: {
+            subscription_items: [
+              {plan: 'foo', quantity: 2},
+              {id: 'si_123', deleted: true},
+            ],
+          },
+        };
+
+        const scope = nock('https://' + options.host)
+          .get(options.path)
+          .query(Object.assign({customer: 'cus_123'}, options.data))
+          .reply(200, '{}');
+
+        realStripe.invoices.retrieveUpcoming('cus_123', options.data, function(err, response) {
+          done();
+          scope.done();
+        });
+      });
+
+      it('encodes the body in POST requests', function(done) {
+        var options = {
+          host: stripe.getConstant('DEFAULT_HOST'),
+          path: '/v1/subscriptions/sub_123',
+          data: {
+            customer: 'cus_123',
+            items: [
+              {plan: 'foo', quantity: 2},
+              {id: 'si_123', deleted: true},
+            ],
+          },
+          body: 'customer=cus_123&items[0][plan]=foo&items[0][quantity]=2&items[1][id]=si_123&items[1][deleted]=true',
+        };
+
+        const scope = nock('https://' + options.host)
+          .post(options.path, options.body)
+          .reply(200, '{}');
+
+        realStripe.subscriptions.update('sub_123', options.data, function(err, response) {
+          done();
+          scope.done();
+        });
+      });
+    });
+  });
+
   describe('Retry Network Requests', function() {
     // Use a real instance of stripe as we're mocking the http.request responses.
     var realStripe = require('../lib/stripe')(utils.getUserStripeKey());
--- a/test/resources/Invoices.spec.js
+++ b/test/resources/Invoices.spec.js
@@ -113,7 +113,7 @@ describe('Invoices Resource', function() {
         expect(stripe.LAST_REQUEST).to.deep.equal({
           method: 'GET',
           url: '/v1/invoices/upcoming?customer=cus_123&' +
-            'subscription_items%5B0%5D%5Bplan%5D=potato&subscription_items%5B1%5D%5Bplan%5D=rutabaga',
+            'subscription_items[0][plan]=potato&subscription_items[1][plan]=rutabaga',
           headers: {},
           data: {},
         });
@@ -137,18 +137,11 @@ describe('Invoices Resource', function() {
           url: '/v1/invoices/upcoming?customer=cus_123&subscription=sub_123',
           headers: {},
           data: {
-            subscription_items: {
-              0: {
-                plan: 'potato',
-              },
-              1: {
-                plan: 'rutabaga',
-              },
-              2: {
-                deleted: true,
-                id: 'SOME_ID',
-              },
-            },
+            subscription_items: [
+              {plan: 'potato'},
+              {plan: 'rutabaga'},
+              {id: 'SOME_ID', deleted: true},
+            ],
             subscription_prorate: true,
           },
         });
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -34,7 +34,7 @@ describe('utils', function() {
     });
 
     it('Handles deeply nested object', function() {
-      expect(decodeURI(utils.stringifyRequestData({
+      expect(utils.stringifyRequestData({
         a: {
           b: {
             c: {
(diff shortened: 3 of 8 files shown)


test/resources/Invoices.spec.js
```diff
@@ -137,18 +137,11 @@ describe('Invoices Resource', function() {
...
-              2: {
-                deleted: true,
-                id: 'SOME_ID',
-              },
-            },
+            subscription_items: [
+              {plan: 'potato'},
+              {plan: 'rutabaga'},
+              {id: 'SOME_ID', deleted: true},
+            ],
```
rattrayalex-stripe: I'm somewhat worried about our test coverage reduction here... is there anything in our test suite to verify that these POST bodies are form-encoded with `subscription_items[0][plan]` instead of `subscription_items[][plan]` ?

test/utils.spec.js
```diff
@@ -34,33 +34,33 @@ describe('utils', function() {
...
 
     it('Handles arrays of objects', function() {
-      expect(decodeURI(utils.stringifyRequestData({
+      expect(utils.stringifyRequestData({
         a: [
           {b: 'c'},
           {b: 'd'},
         ],
-      }))).to.equal('a[][b]=c&a[][b]=d');
+      })).to.equal('a[0][b]=c&a[1][b]=d');
```
rattrayalex-stripe: \o/

test/utils.spec.js
```diff
@@ -34,33 +34,33 @@ describe('utils', function() {
...
-      expect(decodeURI(utils.stringifyRequestData({
+      expect(utils.stringifyRequestData({
         a: {
-          0: 'c',
-          1: 'd',
+          0: {b: 'c'},
+          1: {b: 'd'},
         },
-      }))).to.equal('a[0]=c&a[1]=d');
+      })).to.equal('a[0][b]=c&a[1][b]=d');
```
rattrayalex-stripe: curious, why this change?

test/utils.spec.js
```diff
@@ -76,18 +76,10 @@ describe('utils', function() {
...
       ].join('&'));
     });
-
-    describe('Stripe-specific cases', function() {
-      it('Handles the `expand` array correctly (producing the form `expand[]=_` for each item', function() {
-        expect(decodeURI(utils.stringifyRequestData({
-          expand: ['a', 'foo', 'a.b.c'],
-        }))).to.equal('expand[]=a&expand[]=foo&expand[]=a.b.c');
-      });
-    });
```
rattrayalex-stripe: why remove this test?

rattrayalex-stripe: on that note... we probably _do_ ideally want want `expand` to continue to be `[]` instead of `[0]`, right? Not a big deal, but it is prettier 😄

test/utils.spec.js
ob-stripe: Mainly to illustrate that a list of hashes and an integer-indexed hash of hashes have the same encoded representation.

test/utils.spec.js
ob-stripe: My preference would be to avoid unnecessary special-casing and use the same encoding scheme everywhere. Which is why I removed this test, there's no real need to test "Stripe-specific" stuff since there are no special cases anymore.

test/resources/Invoices.spec.js
ob-stripe: Fair! Looking into it.

ob-stripe: Okay, so this is difficult to test in the current state. The URL-encoding of the request parameters happens in the original `_request` method, but `getSpyableStripe()` returns an instance of `stripe` where `_request` is patched to memorize the method parameters instead of actually sending the request.

We've recently added `nock` for testing the telemetry stuff, so we could probably write some new tests that use it to check the actual body of our HTTP requests.

test/utils.spec.js
rattrayalex-stripe: Fair enough - I assume we have this behavior in our other libs now too, so best to be consistent. 👍

rattrayalex-stripe:
LGTM

test/resources/Invoices.spec.js
rattrayalex-stripe: Yeah, I think nock sounds like the right tool for this job, should be pretty straightforward - fft DM me or paul if you need help with it!

ob-stripe: So I lied, we didn't add nock at all. No idea why I thought that was the case!

Nevertheless, I tried writing a test with nock to test `StripeResource._request`, but I can't seem to have nock intercept the request -- for some reason the request is always sent to the real API server. Mind giving it a try @rattrayalex-stripe?

rattrayalex-stripe: We add nock in https://github.com/stripe/stripe-node/pull/559 😄

ob-stripe:
@rattrayalex-stripe I added a couple of encoding tests in `StripeResource.spec.js`, ptal.

Unfortunately, for GET requests, it seems nock doesn't let you check the raw query string -- instead, it uses `qs` itself to decode the query string into an object. I'd have preferred to check the encoded string directly just in case `qs`' logic differs from Stripe's in subtle ways, but that doesn't seem to be possible.

test/StripeResource.spec.js
```diff
@@ -36,6 +36,64 @@ describe('StripeResource', function() {
...
+          host: stripe.getConstant('DEFAULT_HOST'),
+          path: '/v1/subscriptions/sub_123',
+          data: {
+            customer: 'cus_123',
+            items: [
+              {plan: 'foo', quantity: 2},
+              {id: 'si_123', deleted: true},
+            ],
+          },
+          body: 'customer=cus_123&items[0][plan]=foo&items[0][quantity]=2&items[1][id]=si_123&items[1][deleted]=true',
```
rattrayalex-stripe: just be crystal-clear because I don't see an `expect` and I'm not super-familiar with `nock`, but if we wrote `[]` instead of `[1]` here, the test fails, right?

ob-stripe: Right! The call to `scope.done()` asserts that all specified calls on that scope were performed.

ob-stripe:
Released as 6.25.1.

PR #566, opened 2019-02-14
Remove rate limiting language

paulasjes-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

We removed network retries for rate limiting errors in https://github.com/stripe/stripe-node/pull/559 for now, but accidentally left references to it in the README.md and comments. This fixes that.

PR #567, opened 2019-02-14
Clarify that automatic network retries occur for most api errors

rattrayalex-stripe:
r? @remi-stripe 
cc @paulasjes-stripe 
cc @brandur-stripe 
cc @stripe/api-libraries 
Follows https://github.com/stripe/stripe-node/pull/566 which follows https://github.com/stripe/stripe-node/pull/559

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ if (process.env.http_proxy) {
 
 ### Network retries
 
-Automatic network retries can be enabled with `setMaxNetworkRetries`. This will retry requests `n` times with exponential backoff if they fail due to connection errors or most api errors. [Idempotency keys](https://stripe.com/docs/api/idempotent_requests) are added where appropriate to prevent duplication.
+Automatic network retries can be enabled with `setMaxNetworkRetries`. This will retry requests `n` times with exponential backoff if they fail due to an intermittent network problem. [Idempotency keys](https://stripe.com/docs/api/idempotent_requests) are added where appropriate to prevent duplication.
 
 ```js
 // Retry a request once before giving up
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ if (process.env.http_proxy) {
 
 ### Network retries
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,7 +147,7 @@ if (process.env.http_proxy) {
 
 ### Network retries
 
-Automatic network retries can be enabled with `setMaxNetworkRetries`. This will retry requests `n` times with exponential backoff if they fail due to connection or conflict errors. [Idempotency keys](https://stripe.com/docs/api/idempotent_requests) are added where appropriate to prevent duplication.
+Automatic network retries can be enabled with `setMaxNetworkRetries`. This will retry requests `n` times with exponential backoff if they fail due to connection errors or most api errors. [Idempotency keys](https://stripe.com/docs/api/idempotent_requests) are added where appropriate to prevent duplication.
```
remi-stripe: What does "most api errors" imply? It seems wrong. The most important API errors are `card_error` and `invalid_request_error` which we would explicitly never retry right?

paulasjes-stripe: I'd be more explicit here and list the exact errors we do retries on to reduce any confusion.

rattrayalex-stripe: to wit: 

```
Automatic network retries can be enabled with `setMaxNetworkRetries`. This will retry requests `n` times with exponential backoff if they fail due to connection errors, 409 (conflict) errors, 503 (availability) errors, and some 500 (internal) errors. [Idempotency keys](https://stripe.com/docs/api/idempotent_requests) are added where appropriate to prevent duplication. 400 (user) errors and 429 (rate limit exceeded) errors are not retried, nor are 500 errors which result from POST requests, as our idempotency framework would simply no-op and return a replayed 500 error.
```

eh?

remi-stripe: I would personally remove the last sentence, it seems quite confusing. Or we need to get into the weeds of what is retry-able and what isn't which seems out of scope from this small section in the readme

rattrayalex-stripe:
updated to match php/python, which I think expresses the intent better. 

r? @remi-stripe

PR #513, opened 2018-10-24
Clarify TypeScript usage

rattrayalex-stripe:
Coming from https://github.com/stripe/stripe-node/issues/296#issuecomment-432718356, I have adapted the [various](https://github.com/stripe/stripe-node/issues/296#issuecomment-335308245) [recommendations](https://github.com/stripe/stripe-node/issues/296#issuecomment-425801351) of @sampsonjoliver and others (thanks!).

Thanks also to everyone who has [contributed](https://github.com/DefinitelyTyped/DefinitelyTyped/commits/master/types/stripe/index.d.ts) to the Stripe DefinitelyTyped package. 

r? @brandur-stripe

PR #568, opened 2019-02-19
Update readme (to fix typescript usage)

hideokamoto:
## ENV

```
$ node -v
v10.5.0

$ npm -v
6.1.0

$ cat package.json  | jq .devDependencies
{
  "stripe": "^6.25.1",
  "@types/stripe": "^6.19.7",
  "typescript": "^3.3.3"
}

$ cat tsconfig.json
{
  "compilerOptions": {
    "target": "es6",
    "module": "commonjs",
    "outDir": "./dist",
    "strict": true,
    "esModuleInterop": true
  },
  "include": [
    "libs"
  ]
}

```

## Expected

I want to write code by TypeScript and the README.md told me by the following code.

```
import * as Stripe from 'stripe';
const stripe = new Stripe('sk_test_...');

const customer: Promise<Stripe.customers.ICustomer> = stripe.customers.create(/* ... */);
```

## Actual
When I try to build the code, I got the following error.

```
$ tsc
> tsc

libs/index.ts:4:16 - error TS2351: Cannot use 'new' with an expression whose type lacks a call or construct signature.

4 const stripe = new Stripe(sk)
                 ~~~~~~~~~~~~~~

  libs/shiftet.ts:1:1
    1 import * as Stripe from 'stripe';
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.

```

## Issue
I think the example should be the following code.

```
import Stripe from 'stripe';
const stripe = new Stripe('sk_test_...');

const customer: Promise<Stripe.customers.ICustomer> = stripe.customers.create(/* ... */);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ To use:
 
 ```ts
 // Note `* as` and `new Stripe` for TypeScript:
-import * as Stripe from 'stripe';
+import Stripe from 'stripe';
 const stripe = new Stripe('sk_test_...');
 
 const customer: Promise<Stripe.customers.ICustomer> = stripe.customers.create(/* ... */);


rattrayalex-stripe:
Hi @hideokamoto , 

Taking a look at this. In https://github.com/stripe/stripe-node/pull/513 , the existing pattern was introduced (shoutout to @sampsonjoliver for the help). I'm not sure if a more recent version of TypeScript caused a change.

rattrayalex-stripe:
Ah, yes, this is due to `"esModuleInterop": true`. If you can turn that off, this goes away (I assume that you can't). 

If `* as` is removed, I see TS passing but a runtime error, since there is no `.default`. I may be able to add that, but until then, I would recommend using a `// @ts-ignore` comment if you can't change your `esModuleInterop` setting.

rattrayalex-stripe:
The following code worked for me with your settings:

```
import {
  // @ts-ignore
  Stripe, 
  customers,
} from 'stripe';

const stripe = new Stripe('sk_test_***');

const hello =  async () => {
  const customer: customers.ICustomer = await stripe.customers.create({description: 'hello'});
  console.log(customer.id);
}

hello();
```

However, `@types/stripe` is not currently aware that you can `import {Stripe} from 'stripe'` – I would recommend filing a ticket with the DefinitelyTyped project to fix that.

hideokamoto:
Oh my!
I update my `tsconf.json` to be false the esModuleInterop, and it works well.

Thanks a lot!

rattrayalex-stripe:
Great to hear! Thanks again for such complete debugging info, and for offering a PR!

PR #574, opened 2019-03-01
Encode Dates as Unix timestamps

### LINKED ISSUE #569: Format `Date` instances as UNIX timestamps in parameters
Node version: v8.15.0
Stripe version : 6.21.0

This code does not fetch any charge
```
await stripe.charges.list({ created: { gt: fromDateTimestamp }, limit: 100 }).autoPagingEach(charge => {
         console.log(charge); // is never called
    });
```

whereas this code does fetch charges that were created after `fromDateTimestamp`
```
await stripe.charges.list({ limit: 100 }).autoPagingEach(charge => {
         console.log(charge); // is called
    });
```

I would expect the first code to return filtered values or at least log a warning saying that the created option is not supported.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Encode `Date`s as Unix timestamps, using [`qs`](https://github.com/ljharb/qs)' `serializeDate` option.

Fixes #569.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -27,7 +27,7 @@ var utils = module.exports = {
    * (forming the conventional key 'parent[child]=value')
    */
   stringifyRequestData: function(data) {
-    return qs.stringify(data)
+    return qs.stringify(data, {serializeDate: function (d) { return Math.floor(d.getTime() / 1000); }})
       // Don't use strict form encoding by changing the square bracket control
       // characters back to their literals. This is fine by the server, and
       // makes these parameter strings easier to read.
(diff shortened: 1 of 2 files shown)


remi-stripe:
That LGTM but re-assigning to @rattrayalex-stripe to be safe

rattrayalex-stripe:
Double-checking that we don't expect another datetime format anywhere in the API,...

rattrayalex-stripe:
LGTM

Thanks for this ob!

ob-stripe:
Released as 6.26.0.

PR #578, opened 2019-03-06
Add support for retrieving a Checkout Session

remi-stripe:
This adds support for retrieving a Checkout Session via the API and also changes the shape of the resource to have more details.

r? @rattrayalex-stripe 
cc @stripe/api-libraries


rattrayalex-stripe:
> and also changes the shape of the resource to have more details.

I don't see that? Assuming it's intended to involve no code, 

LGTM

remi-stripe:
Ah yeah it's because I started with PHP that has the list of fields. Thanks for the quick review. I'll merge once the endpoint goes live!

PR #598, opened 2019-04-23
Rework error handling around exec

rattrayalex-stripe:
This is a follow-up to https://github.com/stripe/stripe-node/pull/597 that incorporates [feedback](https://github.com/stripe/stripe-node/pull/597/files#r277846768) from @brandur-stripe 

It also fixes a (small) longstanding bug coming from brandur's hunch; in the `err` case, [this line](https://github.com/stripe/stripe-node/blob/140fd420fbc5407973afff71a578de69e1e1bd3d/lib/stripe.js#L306) would produce `"null"` instead of `"UNKNOWN"`, as intended.

@brennentsmith 's PR introduced no regressions and was minimally-invasive, just taking this opportunity to add some tests for confidence and rework the API a bit (I needed to move things into `utils` anyway for mockability, and figured it was silly _not_ to expose the safer api if I'm already moving things into a utils file – that's what they're for, anyway!). 

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -27,7 +27,6 @@ Stripe.INITIAL_NETWORK_RETRY_DELAY_SEC = 0.5;
 var APP_INFO_PROPERTIES = ['name', 'version', 'url', 'partner_id'];
 
 var EventEmitter = require('events').EventEmitter;
-var exec = require('child_process').exec;
 var utils = require('./utils');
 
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Wow, above and beyond Alex! LGTM. Thanks!

PR #600, opened 2019-04-26
Flatten parameters in multipart requests

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

Allows for nested parameters in multipart requests (i.e. file creation requests).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -250,6 +250,41 @@ var utils = module.exports = {
 
   // For mocking in tests.
   _exec: exec,
+
+  isObject: function isObject(obj) {
+    var type = typeof obj;
+    return (type === 'function' || type === 'object') && !!obj;
+  },
+
+  // For use in multipart requests
+  flattenAndStringify: function flattenAndStringify(data) {
+    var result = {};
+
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -250,6 +250,41 @@ var utils = module.exports = {
 
   // For mocking in tests.
   _exec: exec,
+
+  // For use in multipart requests
+  flattenAndStringify: function flattenAndStringify(data) {
+    var result = {};
+
+    function isObject(obj) {
```
rattrayalex-stripe: this should be a member of `utils`

ob-stripe:
@rattrayalex-stripe TBC, you should only pass params objects to `flattenAndStringify`, it's not meant to work with arrays. But I can add a check and an explicit error if you think that's preferable.

rattrayalex-stripe:
Ah okay. 

LGTM

lib/utils.js
ob-stripe: 👍

ob-stripe:
Thanks Alex!

PR #602, opened 2019-05-01
Handle errors from the oauth/token endpoint

mdlavin:
Before this change, failures from the OAuth API would result in `Invalid JSON received from the Stripe API` errors.

The cause was not a parse error, but an error thrown at https://github.com/stripe/stripe-node/compare/master...mdlavin:handle-oauth-errors?expand=1#diff-404ddfdf655bbca4b088b8c7242d0a3aL155 when the `headers` value was assigned to a string value.

The response body coming back from the oauth/token endpoint is:
```
{
  "error":"invalid_grant",
  "error_description":"This authorization code has already been used. All tokens issued with this code have been revoked."
}
```

As described in https://stripe.com/docs/connect/oauth-reference#post-token-errors

I also added a nicely typed Error for invalid_token responses, and the other OAuth failures will be reported as Generic errors instead of Invalid JSON errors

### DIFF
--- a/lib/Error.js
+++ b/lib/Error.js
@@ -59,6 +59,8 @@ StripeError.generate = function(rawStripeError) {
     return new _Error.StripeAPIError(rawStripeError);
   case 'idempotency_error':
     return new _Error.StripeIdempotencyError(rawStripeError);
+  case 'invalid_grant':
+    return new _Error.StripeInvalidGrantError(rawStripeError);
   }
   return new _Error('Generic', 'Unknown Error');
 };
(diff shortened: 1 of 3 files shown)


ob-stripe:
Thanks for the contribution @mdlavin!

r? @rattrayalex-stripe

mdlavin:
Yes, the test fails without the changes in `lib/StripeResource.js`. I reverted to master locally and here is the failure:

```
  18 passing (110ms)
  1 failing

  1) StripeResource
       Retry Network Requests
         _request
           should handle OAuth errors gracefully:

      Uncaught AssertionError: expected 'StripeAPIError' to equal 'StripeInvalidGrantError'
      + expected - actual

      -StripeAPIError
      +StripeInvalidGrantError
      
      at /Users/mattlavin/Projects/stripe-node/test/StripeResource.spec.js:244:31
      at Timeout._onTimeout (lib/utils.js:39:295)

```

The `StripeAPIError` has the message `Invalid JSON received from the Stripe API`

rattrayalex-stripe:
Awesome, thanks so much Matt!

rattrayalex-stripe:
Released! 

We generally try to release ~immediately after merging master; apologies for the delay.

PR #604, opened 2019-05-03
Prettier formatting

irace-stripe:
Added [Prettier](https://prettier.io/) formatting.


rattrayalex-stripe:
cc @paulasjes-stripe @rob-stripe

lib/StripeMethod.basic.js
```diff
@@ -54,39 +53,60 @@ module.exports = {
...
+            metadata[key] = value;
+            sendMetadata(metadata, auth);
+          } else {
+            // Set entire metadata object after resetting it:
+            this._request(
+              'POST',
+              null,
+              path,
+              {
+                metadata: null,
```
rattrayalex-stripe: minor, would you mind removing the newline before `metadata` here such that the output is `{metadata: null}`? this is the ~one place that prettier respects original formatting, annoyingly

rattrayalex-stripe: (same below)

README.md
```diff
@@ -88,30 +79,53 @@ callback:
...
+
+```js
+var stripe = require('stripe')('sk_test_...');
+
+stripe.customers.create(
+  {
+    email: 'customer@example.com',
+  },
+  function(err, customer) {
+    err; // null if no error occurred
```
paulasjes-stripe: Nit: Can we do something with these values? It looks a little odd by themselves. Perhaps something like:

```
if (err) {
  // Deal with an error
}

// Do something with created customer object
console.log(customer.id) 
```

irace-stripe:
@rattrayalex-stripe @paulasjes-stripe Feedback addressed. Since this was already approved and my changes were trivial, I’m going to merge this.

irace-stripe:
LGTM

PR #607, opened 2019-05-04
Modernize ES5 to ES6 with lebab

rattrayalex-stripe:
Builds on https://github.com/stripe/stripe-node/pull/604, targeting the v7.0.0 release

Uses https://github.com/lebab/lebab, running each transform in its own commit. 

See the script I added in the first commit (and removed in the last) for the process.

The primary benefits are, IMO, `var`->`const`, `function`->`=>` and `{foo: foo}`->`{foo}`.

cc @stripe/api-libraries 
cc @paulasjes-stripe @robz-stripe @ob-stripe

I'm putting a lot of faith in the test suite here, and my confidence isn't 100% that this is bug-free. I'm not totally sure how to proceed with confidence, or who should review. 

EDIT: looking through these commit-by-commit, a few I'm reverting because I think they're bad/dangerous. 

These I'm not sure about: 

1. c1934deb3307e242e8d7eea1ff9a91a06f22cc8e I think I should just carefully check that there's code coverage for any possible bugs introduced here. I like the change and would like to keep it.
2. 08e4905237ea6eb7b4e6d049969edf0e9048b102 Relatedly, I love this change but do think there's some possibility it could introduce some subtle edge-case bugs. This one however is way too big to audit. I think any bugs that the test suite wouldn't have caught would be of the shape of "a user customizes their stripe-node setup in some way that we support but don't test and there was some implicit reliance on `this`". I'm not sure how large the surface area for this is...

Note that I made a few changes by hand (which I probably should have done in separate commits in retrospect). They were: 

1. This needed to stay `function` instead of arrow functions:
https://github.com/stripe/stripe-node/blob/d80c8f5e906c8fabfa4993ad9365ffe71d9e0912/lib/ResourceNamespace.js#L16-L20
2. These needed to move to `const` (lebab moved `var` to `let`, then eslint complained that they were not reassigned): 
https://github.com/stripe/stripe-node/blob/d80c8f5e906c8fabfa4993ad9365ffe71d9e0912/lib/StripeResource.js#L341-L343
3. This needed to stay a `function`:
https://github.com/stripe/stripe-node/blob/d80c8f5e906c8fabfa4993ad9365ffe71d9e0912/test/utils.spec.js#L114


rattrayalex-stripe:
cc @irace-stripe FYI – looks like it should mostly affect you in that it translates `var` to `const` at the top of the files.

paulasjes-stripe:
I ran the tests on node v6.0.0 and v8.0.0. Works as expected in v6, broke on autopagination tests in v8. Turns out it's a node error, upgrading to v8.1 fixed the issue.

Perhaps we should exclude v8 in favour of v8.1.0?

irace-stripe:
> cc @irace-stripe FYI – looks like it should mostly affect you in that it translates var to const at the top of the files.

Noted – a change I’m happy to make!

lib/StripeResource.js
```diff
@@ -50,7 +50,7 @@ StripeResource.prototype = {
   // Methods that don't use the API's default '/v1' path can override it with this setting.
   basePath: null,
 
-  initialize: function() {},
+  initialize() {},
```
paulasjes-stripe: I'd either revert this or change it to `initialize: () => {},`

lib/resources/Files.js
```diff
@@ -18,43 +18,43 @@ module.exports = StripeResource.extend({
...
           errorHandler(err);
         });
     }
 
     function streamError(callback) {
-      var StreamProcessingError = Error.extend({
+      const StreamProcessingError = Error.extend({
         type: 'StreamProcessingError',
-        populate: function(raw) {
+        populate(raw) {
```
paulasjes-stripe: Nit like before, but I prefer `populate: (raw) => {`

lib/StripeResource.js
rattrayalex-stripe: Yeah, I prefer that style too... Doesn't seem lebab supports it, I think. Will try to see if I can find an eslint autofix...

rattrayalex-stripe:
@paulasjes-stripe in https://github.com/stripe/stripe-node/pull/607/commits/6f2dcb6aadad78239d684d5f9f98ec33566e4a4d I used arrow syntax for more methods; there weren't many places I could safely do this. 

I also tried to migrate from `self` to `this`, but tests started failing and I aborted before rabbit-holing too deeply. Would appreciate a follow-on PR (no urgency, can happen post-release) finishing this if you're willing to submit one! 

r? @ob-stripe

rattrayalex-stripe:
Per IRL, 

LGTM

Issue #601, opened 2019-04-28
Export generateHeaderString or similar testing utility


tatethurston:
When writing an automated test case to test processing a Stripe webhook, I needed to copy the `generateHeaderString` function (found here: https://github.com/stripe/stripe-node/blob/57d5f09b13973f2134326dabd99762bd66de3bf2/test/Webhook.spec.js#L140).

It would be great to provide this as an exported module, and include a brief example in the readme using it to write tests against a webhook endpoint.

paulasjes-stripe:
Hi there, this should be exposed and ready for use in the next release!

PR #609, opened 2019-05-07
Exposing generateHeaderString as generateWebhookHeaderString

paulasjes-stripe:
Fixes https://github.com/stripe/stripe-node/issues/601

Exposes `generateWebhookHeaderString` so users can mock Stripe webhook events. This feature already existed within the repo, this just lets others play with it.

I moved some constants around and changed `generateHeaderString` to also accept an instance of `stripe`, in case you already have a spyable instance you want to reuse.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 *.node*.js
+node_modules
--- a/README.md
+++ b/README.md
@@ -223,6 +223,29 @@ const event = stripe.webhooks.constructEvent(
 );
 ```
 
+#### Testing Webhook signing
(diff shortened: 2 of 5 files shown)


.eslintignore
```diff
@@ -1 +1,2 @@
 *.node*.js
+node_modules
```
paulasjes-stripe: I added this cause the linter would yell about `node_modules` files, which you could have in the `/examples` directory.

rattrayalex-stripe: Thank you!

PR #613, opened 2019-05-08
🔥Remove deprecated methods 🔥

ob-stripe:
r? @rattrayalex-stripe @remi-stripe 
cc @stripe/api-libraries @irace-stripe 

Remove all deprecated methods, as documented in https://github.com/stripe/stripe-node/wiki/%5BWIP%5D-Migration-guide-for-v7.


lib/resources/Balance.js
```diff
@@ -9,16 +9,4 @@ module.exports = StripeResource.extend({
   retrieve: stripeMethod({
```
ob-stripe: The `balance` object is a "singleton", it doesn't have an ID. You just send a GET request to `/v1/balance` and get the singleton object back.

lib/resources/Issuing/Cards.js
```diff
@@ -6,14 +6,7 @@ const stripeMethod = StripeResource.method;
...
-    'create',
-    'list',
-    'retrieve',
-    'update',
-    'setMetadata',
-    'getMetadata',
-  ],
+  includeBasic: ['create', 'list', 'retrieve', 'update'],
 
   retrieveDetails: stripeMethod({
```
ob-stripe: Consistency with other nested resources. `/v1/issuing/cards/{card}/details` only has a single method, but if it had multiple ones they'd be called e.g. `createDetails`.

Arguably the method should be called `listDetails` and not `retrieveDetails` since it doesn't take an ID for the nested resource, but the endpoint returns a single object, not a list object. `card_details` is a bit similar to `balance`, it's sort of a singleton (in the context of a card).

lib/resources/Sources.js
```diff
@@ -6,7 +6,7 @@ const stripeMethod = StripeResource.method;
 module.exports = StripeResource.extend({
   path: 'sources',
 
-  includeBasic: ['create', 'retrieve', 'update', 'setMetadata', 'getMetadata'],
+  includeBasic: ['create', 'retrieve', 'update'],
 
   listSourceTransactions: stripeMethod({
```
ob-stripe: Same as above, it's called `listSourceTransactions` to be consistent with other nested resources. If there were other methods they'd be called `createSourceTransaction` etc. (In this case the endpoint does return a list object containing a list of `source_transaction` objects, so `listSourceTransactions` is the correct name.)

For `detach`, you're correct that it's missing, but you can call `customers.deleteSource()` which does exactly the same thing (although the name is a bit misleading since in the case of a source, it won't actually be deleted, just detached from the customer).

lib/resources/SubscriptionSchedules.js
```diff
@@ -6,14 +6,7 @@ const stripeMethod = StripeResource.method;
...
 
-  includeBasic: [
-    'create',
-    'list',
-    'retrieve',
-    'update',
-    'setMetadata',
-    'getMetadata',
-  ],
+  includeBasic: ['create', 'list', 'retrieve', 'update'],
```
ob-stripe: Yeah, fairly sure this is what we want!

lib/resources/Issuing/Cards.js
remi-stripe: In a way, this is not really a nested resource though right? A nested resource would be a POST or a LIST on that endpoint but it's a GET so it's more like `/pay` on Invoice no?

lib/resources/Balance.js
rattrayalex-stripe: yes, it's a singleton retrieve, not a standard retrieve.

rattrayalex-stripe: (that is, it takes no `id`)

lib/resources/Recipients.js
```diff
@@ -1,48 +1,8 @@
...
-    path: '/{recipientId}/cards/{cardId}',
-    urlParams: ['recipientId', 'cardId'],
-  }),
-
-  deleteCard: stripeMethod({
-    method: 'DELETE',
-    path: '/{recipientId}/cards/{cardId}',
-    urlParams: ['recipientId', 'cardId'],
-  }),
+  includeBasic: ['create', 'del', 'list', 'retrieve', 'update'],
```
remi-stripe: so we're not removing `create` and all that?

lib/resources/Recipients.js
```diff
@@ -1,48 +1,8 @@
...
-    'getMetadata',
-  ],
-
-  createCard: stripeMethod({
-    method: 'POST',
-    path: '/{id}/cards',
-    urlParams: ['id'],
-  }),
-
-  listCards: stripeMethod({
```
remi-stripe: How do you list cards if not through this? Sorry I don't remember what we discussed already :goldfish:

lib/resources/SubscriptionSchedules.js
rattrayalex-stripe: I think in this case and in the `listSourceTransactions` case, what you see in x-stripe fields is java-specific, not canonically true.

lib/resources/Transfers.js
```diff
@@ -6,33 +6,7 @@ const stripeMethod = StripeResource.method;
...
-    'getMetadata',
-  ],
-
-  reverse: stripeMethod({
-    method: 'POST',
-    path: '/{id}/reversals',
-    urlParams: ['id'],
-  }),
-
-  cancel: stripeMethod({
```
remi-stripe: are you sure `cancel` should go?

lib/stripe.js
```diff
@@ -121,9 +117,6 @@ const resources = {
   }),
 };
 
-// For backwards compatibility, `Files` is aliased to `FileUploads`
-resources.FileUploads = resources.Files;
```
remi-stripe: So we're sure we want to remove this? I think that's what we gavel-ed but wanted to reflag so that the IRL discussion is tracked in text

test/flows.spec.js
```diff
@@ -104,8 +102,9 @@ describe('Flows', function() {
         stripe.customers.create(CUSTOMER_DETAILS).then((customer) => {
           cleanup.deleteCustomer(customer.id);
 
-          return stripe.customers
-            .updateSubscription(customer.id, {
+          return stripe.subscriptions
```
remi-stripe: Some of those tests should likely just be removed right? You're basically just re-testing subscriptions here (though we stopped adding anything to this flow test suite a while ago)

lib/resources/Sources.js
rattrayalex-stripe: That `detach` thing looks pretty weird to me (and by weird, I mean, quite probably a bug? cc @mickjermsurawong-stripe )

lib/stripe.js
rattrayalex-stripe: I'm +1 to removing `FileUploads`.

lib/resources/Balance.js
```diff
@@ -9,16 +9,4 @@ module.exports = StripeResource.extend({
...
     method: 'GET',
   }),
-
-  listTransactions: stripeMethod({
-    method: 'GET',
-    path: 'history',
-    methodType: 'list',
-  }),
-
-  retrieveTransaction: stripeMethod({
```
rattrayalex-stripe: Wait, I thought we gaveled on keeping things like this?

lib/resources/Transfers.js
rattrayalex-stripe: (in the wiki, it says it's moved to `payouts.cancel()`)

test/flows.spec.js
rattrayalex-stripe: +1

ob-stripe: I wanted to keep the scope for this PR contained to just removing deprecated methods. But yes, we should get rid of `flows.spec.js` entirely and ideally move the rest of the test suite to use stripe-mock instead of local mocks.

lib/resources/Balance.js
ob-stripe: In the general case of nested resources, yes, we're keeping both `stripe.parents.retrieveNested()` and `stripe.nested.retrieve()`. But balance transactions are not a nested resource, they were just historically implemented as one but I think that was an inconsistency we should fix here.

lib/resources/Issuing/Cards.js
ob-stripe: @remi-stripe I think it qualifies as a nested resource because the endpoint returns a different resource.

lib/resources/Recipients.js
ob-stripe: It's still in the spec (although marked as deprecated), so no.

lib/resources/Recipients.js
ob-stripe: It's no longer in the spec, so... you don't?

lib/stripe.js
ob-stripe: 👍

lib/resources/Transfers.js
ob-stripe: @remi-stripe You can use `/v1/payouts/{payout}/cancel` with any API version, right? If so I think removing this is the right call.

lib/resources/Balance.js
irace-stripe: Any reason this is isn’t `includeBasic: ['retrieve']`?

lib/resources/Sources.js
irace-stripe: Any reason that this is `listSourceTransactions` as opposed to `sourceTransactions`? Differs from OpenAPI in this regard.

irace-stripe: FWIW, OpenAPI also includes the following, which is not included here:

```js
detach: stripeMethod({
  method: 'DELETE',
  path: '/{id}/{id}',
  urlParams: ['id'],
}),
```

lib/resources/SubscriptionSchedules.js
irace-stripe: The `listRevisions` method in this file is listed in OpenAPI as just `revisions` (I can user overrides to keep it as `listRevisions`, just making sure this is what we want).

lib/resources/Issuing/Cards.js
irace-stripe: Any idea why this is `retrieveDetails` when OpenAPI has it as just `details`?

lib/resources/Balance.js
rattrayalex-stripe: Ah, makes sense! Thanks. 

cc @irace-stripe

test/flows.spec.js
rattrayalex-stripe: Can you create a follow-on jira?

lib/resources/Recipients.js
ob-stripe: After consideration, these methods should be in the spec, so I'll re-add them.

rattrayalex-stripe:
(per IRL, we'll add recipients.listCards back in and then merge)

PR #614, opened 2019-05-08
Drop support for Node 9 and bump dependencies to latest versions

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

Bump all dependencies to their latest versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,22 +22,22 @@
   },
   "bugs:": "https://github.com/stripe/stripe-node/issues",
   "engines": {
-    "node": "6 || >=8.1"
+    "node": "^6 || ^8.1 || >=10.*"
   },
   "main": "lib/stripe.js",
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


rattrayalex-stripe:
Sorry, hope to look at this ~tomorrow... (per IRL, +1 to also dropping Node 9 as part of this)

package.json
```diff
@@ -22,22 +22,22 @@
   },
   "bugs:": "https://github.com/stripe/stripe-node/issues",
   "engines": {
-    "node": "6 || >=8.1"
+    "node": "^6 || ^8.1 || >=10.*"
```
ob-stripe: Is this correct? OOC, why are we excluding 8.0, is there a known issue with it?

rattrayalex-stripe: looks correct to me, and travis seems to agree. 

yes, paul found an issue with it 😄

PR #615, opened 2019-05-08
Removed most instances of 'self'

paulasjes-stripe:
Removed almost all cases of `const self = this;` since we're (mostly) using arrow functions now.

Notable exceptions are functions that are getting removed in https://github.com/stripe/stripe-node/pull/613 (e.g. `setMetadata`) and functions which require the scope to be untouched (most of the Mocha test suite functions).

Ran the tests in Node 6 and Node 8, leaning quite heavily on them here.


rattrayalex-stripe:
Sorry, hope to look at this ~tomorrow...

lib/StripeMethod.js
```diff
@@ -21,19 +21,18 @@ const makeAutoPaginationMethods = require('./autoPagination')
  */
 function stripeMethod(spec) {
   return function() {
-    const self = this;
-    const args = [].slice.call(arguments);
+    const args = [...arguments];
```
rattrayalex-stripe: why not `return function (...args)` ?

lib/StripeResource.js
```diff
@@ -370,30 +366,30 @@ StripeResource.prototype = {
...
     }
 
-    function retryRequest(requestFn, apiVersion, headers, requestRetries) {
+    const retryRequest = (requestFn, apiVersion, headers, requestRetries) => {
       requestRetries += 1;
 
       return setTimeout(
-        requestFn,
-        self._getSleepTimeInMS(requestRetries),
+        requestFn.bind(this),
```
rattrayalex-stripe: ahhh I think this was what I was missing earlier

rattrayalex-stripe: err, why do we not translate `makeRequest` to be an arrow function?

rattrayalex-stripe:
LGTM

though if we can get that last one as an arrow function that'd be nice too

paulasjes-stripe:
@rattrayalex-stripe I was getting some weird errors with Nock when `makeRequest` was an arrow function rather than a regular function. I dug into them and found that it was simply a case of temporal dead zone where `makeRequest` was being called before it was defined 🤦‍♂. 

Should be ready for re-review now.

rattrayalex-stripe:
@paulasjes-stripe Still seeing a `ReferenceError: self is not defined` in Travis. 

What do you think about holding off on this until after 7.0.0 lands? Feels like it may be a bit riskier, might be good to isolate it a bit from the other changes.

rattrayalex-stripe:
Per DM, we're going to hold off on this until 7.0.0 has been released and live for a few days.

paulasjes-stripe:
Closing this as will be merging into master rather than the old `7.0.0` branch.

Issue #616, opened 2019-05-08
Cannot find module '../testUtils'


schester44:
Related to PR #609.

I believe the testUtils folder needs whitelisted in .npmignore.

remi-stripe:
@schester44 Thanks for the heads up, we're working on a fix.

rattrayalex-stripe:
We've pulled the 6.33.0 release from npm. We expect to have a fixed version up shortly, but until then, if you encounter this, please reinstall to get 6.32.0.

rattrayalex-stripe:
6.34.0 has been released, addressing this issue. 

Thanks so much for reporting, and apologies for the trouble caused!

PR #617, opened 2019-05-08
Add testUtils to npmignore

paulasjes-stripe:
Should fix https://github.com/stripe/stripe-node/issues/616

Let me know if we want to only specify `index` here or if the whole folder is okay.

r? @remi-stripe @rattrayalex-stripe

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -10,3 +10,4 @@
 !/VERSION
 !/package.json
 !/lib/**/*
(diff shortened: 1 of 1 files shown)


rattrayalex-stripe:
Hmm, I don't think this is what we want – it'd defeat the point of https://github.com/stripe/stripe-node/pull/609, I think.

PR #619, opened 2019-05-08
Move generateTestHeaderString to stripe.webhooks

paulasjes-stripe:
Attempt 2!

Moved the header generator to a more reasonable place. This version is a little more strict with what you pass in, this is mainly because we don't want test constants in non-test code.

Also no longer requiring things that aren't whitelisted in `.npmignore`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,8 +236,9 @@ const payload = {
 const payloadString = JSON.stringify(payload, null, 2);
 const secret = 'whsec_test_secret';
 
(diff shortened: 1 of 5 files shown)


rattrayalex-stripe:
For easy linking, this is a followup to https://github.com/stripe/stripe-node/pull/609 which fixes a bug raised here: https://github.com/stripe/stripe-node/issues/616

PR #620, opened 2019-05-08
Add VSCode and EditorConfig files

ob-stripe:
r? @rattrayalex-stripe

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,11 @@
+{
+  "recommendations": [
+    // JS Language plugins (~required)
+    "dbaeumer.vscode-eslint",
+    "joshpeng.sublime-babel-vscode",
+    "esbenp.prettier-vscode",
+
+    // Highly recommended
+    "EditorConfig.editorconfig"
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,13 @@
+{
+  // Show the path in the top window bar.
+  "window.title": "${rootName}${separator}${activeEditorMedium}",
+
+  // Formatting
+  "editor.formatOnSave": true,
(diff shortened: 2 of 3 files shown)


rattrayalex-stripe:
I thought I added this in my branch? they are slightly different, and should probably be merged...

.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
+    "recommendations": [
+        // JS Language plugins (~required)
+        "dbaeumer.vscode-eslint",
+        "joshpeng.sublime-babel-vscode",
+        "esbenp.prettier-vscode",
+
+        // Highly recommended
+        "eamodio.gitlens",
+        "ziyasal.vscode-open-in-github",
```
rattrayalex-stripe: let's remove these two, so that contributors don't get cluttered.

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+  // Show the path in the top window bar.
+  "window.title": "${rootName}${separator}${activeEditorMedium}",
+
+  // Formatting
+  "editor.formatOnSave": true,
```
rattrayalex-stripe: there is some beautify excludes that we have in others like prettier-poet, can you copy those over here?

ob-stripe: Hm, I can't find any `beautify.ignore` setting in VSCode with my current configuration. I assume it's a setting for a different extension is not in the list of recommended extensions, but that, if installed separately, would interfere with ESLint auto-formatting? Is that correct?

rattrayalex-stripe: yeah, with the prettier auto-formatting. 

We want this:
```
"beautify.ignore": ["**/*.js", "**/*.md"],
```

rattrayalex-stripe: Eh, I'll just add that to your PR.

PR #621, opened 2019-05-08
Remove legacy parameter support in invoices.retrieveUpcoming()

ob-stripe:
r? @rattrayalex-stripe @remi-stripe 
cc @stripe/api-libraries @irace-stripe 

- Removes support for legacy customer and subscription parameters in `invoices.retrieveUpcoming()`. All parameters must now be provided in a hash.

- While I was at it, I also noticed that `retrieveLines()` should really be called `listLineItems()` because it returns a list object containing a list of `line_item`s.

- I also added `listUpcomingLineItems()` because I think that's much cleaner than telling users to use `listLines()` with `"upcoming"` as the ID.

### DIFF
--- a/lib/resources/Invoices.js
+++ b/lib/resources/Invoices.js
@@ -2,7 +2,6 @@
 
 const StripeResource = require('../StripeResource');
 const stripeMethod = StripeResource.method;
-const utils = require('../utils');
 
 module.exports = StripeResource.extend({
   path: 'invoices',
@@ -12,6 +11,17 @@ module.exports = StripeResource.extend({
     method: 'POST',
     path: '{id}/finalize',
     urlParams: ['id'],
+  }),
+
+  listLineItems: stripeMethod({
+    method: 'GET',
+    path: '{id}/lines',
+    urlParams: ['id'],
+  }),
+
+  listUpcomingLineItems: stripeMethod({
+    method: 'GET',
+    path: 'upcoming/lines',
   }),
 
   markUncollectible: stripeMethod({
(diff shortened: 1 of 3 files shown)


lib/resources/Invoices.js
```diff
@@ -14,6 +13,17 @@ module.exports = StripeResource.extend({
     urlParams: ['id'],
   }),
 
+  listLines: stripeMethod({
+    method: 'GET',
+    path: '{id}/lines',
+    urlParams: ['id'],
+  }),
+
+  listLinesUpcoming: stripeMethod({
```
remi-stripe: In other libraries we do this by passing `upcoming` as the id.

test/StripeResource.spec.js
```diff
@@ -49,11 +49,12 @@ describe('StripeResource', () => {
     });
 
     describe('_request', () => {
-      it('encodes the query string in GET requests', (done) => {
+      it('encodes the body in GET requests', (done) => {
```
ob-stripe: Fun fact, stripe-node always sends parameters in request bodies regardless of the HTTP method. The only exception was `retrieveUpcoming()` because it had a custom implementation.

Ideally, the library should send parameters in query strings for GET and DELETE requests and in request bodies for POST requests, but it doesn't look like a simple fix and the backend doesn't care.

rattrayalex-stripe: 😮

lib/resources/Invoices.js
rattrayalex-stripe: I would actually think this'd be `listUpcomingLines`...

rattrayalex-stripe: I'm also uncomfortable-but-peaceful with an `'upcoming'` id if that's what we do in other libs (and if it has not caused user confusion/sadness)

rattrayalex-stripe:
Thanks for doing this!
Looks great. Would like to settle the question around `listLinesUpcoming`, then we can ship. 

ptal @ob-stripe

lib/resources/Invoices.js
ob-stripe: I renamed it `listUpcomingLineItems`, which is consistent with what we do in .NET: https://stripe.com/docs/api/invoices/upcoming_invoice_lines?lang=dotnet.

We should also add this to other languages because having users pass a magic ID as a string is terrible.

ob-stripe:
@remi-stripe @rattrayalex-stripe Are you both okay with `listUpcomingLineItems()`?

rattrayalex-stripe:
I am okay with that!

ob-stripe:
Cool. I updated the WIP migration guide with the changes here. Pulling this in!

PR #606, opened 2019-05-03
stripe-node v7.0.0

rattrayalex-stripe:
This is a major release of stripe-node, dropping support for several old versions of Node and deprecated methods.

[The migration guide](https://github.com/stripe/stripe-node/wiki/Migration-guide-for-v7) contains a detailed list of backwards-incompatible changes with upgrade instructions.

cc @stripe/api-libraries 
cc @ob-stripe @paulasjes-stripe @robz-stripe 

⚠️ = breaking changes

- [x] ⚠️ Drop support for Node 4, 5 and 7 (#606)
- [x] Prettier formatting (#604)
- [x] Alphabetize “basic” methods (#610)
- [x] Use `id` for single positional arguments (#611)
- [x] Modernize ES5 to ES6 with lebab (#607)
- [x] ⚠️ 🔥Remove deprecated methods 🔥 (#613)
- [x] Add VSCode and EditorConfig files (#620)
- [x] ⚠️ Drop support for Node 9 and bump dependencies to latest versions (#614)
- [x] Misc. manual formatting (#623)
- [x] ⚠️ Remove legacy parameter support in `invoices.retrieveUpcoming()` (#621)
- [x] ⚠️ Remove curried urlData and manually specified urlParams (#625)
- [x] Extract resources file (to aid with code generation) (#626)

### follow-on 

- [ ] Removed most instances of `self` (#615)

PR #623, opened 2019-05-09
Misc. manual formatting

irace-stripe:
Some misc. manual formatting to get us closer to the code style that we’re striving for:

- No chained method calls on `require` statements
- Paths start with `/`
- Line break between `path` and `includeBasic` entries in object literals
- Alphabetical method ordering
- Remove misc. comments

r? @rattrayalex-stripe 
r? @ob-stripe

### DIFF
--- a/lib/resources/CreditNotes.js
+++ b/lib/resources/CreditNotes.js
@@ -5,11 +5,12 @@ const stripeMethod = StripeResource.method;
 
 module.exports = StripeResource.extend({
   path: 'credit_notes',
+
   includeBasic: ['create', 'list', 'retrieve', 'update'],
 
   voidCreditNote: stripeMethod({
     method: 'POST',
-    path: '{id}/void',
+    path: '/{id}/void',
     urlParams: ['id'],
   }),
 });
--- a/lib/resources/LoginLinks.js
+++ b/lib/resources/LoginLinks.js
@@ -4,5 +4,6 @@ const StripeResource = require('../StripeResource');
 
 module.exports = StripeResource.extend({
   path: 'accounts/{accountId}/login_links',
+
   includeBasic: ['create'],
 });
--- a/lib/resources/SubscriptionScheduleRevisions.js
+++ b/lib/resources/SubscriptionScheduleRevisions.js
@@ -2,19 +2,8 @@
 
 const StripeResource = require('../StripeResource');
 
-/**
- * SubscriptionScheduleRevisions is a unique resource in that, upon instantiation,
- * requires a subscription schedule id, and therefore each of its methods only
- * require the scheduleId argument.
- *
- * This streamlines the API specifically for the case of accessing a revision
- * on a returned subscription schedule object.
- *
- * E.g. scheduleObject.revisions.retrieve(revisionId)
- * (As opposed to the also-supported stripe.subscriptionSchedules.retrieveRevision(scheduleId,
- * revisionId))
- */
 module.exports = StripeResource.extend({
   path: 'subscription_schedules/{scheduleId}/revisions',
+
   includeBasic: ['list', 'retrieve'],
 });
--- a/lib/resources/TransferReversals.js
+++ b/lib/resources/TransferReversals.js
@@ -2,19 +2,8 @@
 
 const StripeResource = require('../StripeResource');
 
-/**
- * TransferReversals is a unique resource in that, upon instantiation,
- * requires a transferId, and therefore each of its methods only
- * require the reversalId argument.
- *
- * This streamlines the API specifically for the case of accessing reversals
- * on a returned transfer object.
- *
- * E.g. transferObject.reversals.retrieve(reversalId)
- * (As opposed to the also-supported stripe.transfers.retrieveReversal(transferId,
- * reversalId))
- */
 module.exports = StripeResource.extend({
   path: 'transfers/{transferId}/reversals',
+
   includeBasic: ['create', 'list', 'retrieve', 'update'],
 });
--- a/lib/resources/AccountLinks.js
+++ b/lib/resources/AccountLinks.js
@@ -1,6 +1,9 @@
 'use strict';
 
-module.exports = require('../StripeResource').extend({
+const StripeResource = require('../StripeResource');
+
+module.exports = StripeResource.extend({
   path: 'account_links',
+
   includeBasic: ['create'],
 });
--- a/lib/resources/ApplePayDomains.js
+++ b/lib/resources/ApplePayDomains.js
@@ -1,6 +1,9 @@
 'use strict';
 
-module.exports = require('../StripeResource').extend({
+const StripeResource = require('../StripeResource');
+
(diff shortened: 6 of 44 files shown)


lib/resources/CreditNotes.js
```diff
@@ -5,11 +5,12 @@ const stripeMethod = StripeResource.method;
 
 module.exports = StripeResource.extend({
   path: 'credit_notes',
+
   includeBasic: ['create', 'list', 'retrieve', 'update'],
 
   voidCreditNote: stripeMethod({
     method: 'POST',
-    path: '{id}/void',
+    path: '/{id}/void',
```
ob-stripe: Is that correct? Other paths do not have a leading `/`.

irace-stripe: @ob-stripe Some do, some don’t. I’m assuming they should all be consistent at the very least and unless I’m misremembering, @rattrayalex-stripe suggested leading `/`

rattrayalex-stripe: I did. I'm pretty sure we have something that normalizes this. just in case, @irace-stripe can you check that there is test coverage for credit notes?

lib/resources/LoginLinks.js
```diff
@@ -3,6 +3,7 @@
 const StripeResource = require('../StripeResource');
 
 module.exports = StripeResource.extend({
-  path: 'accounts/{accountId}/login_links',
+  path: 'accounts/{account}/login_links',
```
rattrayalex-stripe: this should be `accountId`, no?

lib/resources/TransferReversals.js
```diff
@@ -2,19 +2,8 @@
...
- * This streamlines the API specifically for the case of accessing reversals
- * on a returned transfer object.
- *
- * E.g. transferObject.reversals.retrieve(reversalId)
- * (As opposed to the also-supported stripe.transfers.retrieveReversal(transferId,
- * reversalId))
- */
 module.exports = StripeResource.extend({
-  path: 'transfers/{transferId}/reversals',
+  path: 'transfers/{id}/reversals',
```
rattrayalex-stripe: this should be `transferId` right?

rattrayalex-stripe:
few nits on nested resource url things, otherwise this looks magnifico!

r? @irace-stripe

lib/resources/TransferReversals.js
irace-stripe: @rattrayalex-stripe Not as per OpenAPI, but we can override if necessary?

lib/resources/LoginLinks.js
irace-stripe: @rattrayalex-stripe Not as per OpenAPI, but we can override if necessary?

lib/resources/TransferReversals.js
rattrayalex-stripe: ahhh interesting

rattrayalex-stripe: I think passing openapi's view of the world through is fine, carry on

rattrayalex-stripe:
LGTM

pending build failure

irace-stripe:
@rattrayalex-stripe I had to make a bunch of unit test changes in response to having changed the URL parameter names: https://github.com/stripe/stripe-node/pull/623/commits/d8067a8e5d465f78d9779e1ae41d85946dd5bcde#diff-b0d90832c1cb39522f0108600be90655. This scares me a little bit since how would users of this library going to know that they would have to do this?

There’s still one failing test that I’m unsure how to fix, unfortunately.

lib/resources/CreditNotes.js
irace-stripe: There seems to be, though I’m not 100% sure if these tests cover the difference between leading `/` and no leading `/`? https://github.com/stripe/stripe-node/blob/master/test/resources/CreditNotes.spec.js

rattrayalex-stripe:
ahhhh shoot. We weren't sure if the url params could be set with a hash; I was under the impression they couldn't be. 

This definitely implies that the url params _are_ part of our public API, and need to be preserved, which is super sad since we're so inconsistent about it today. 

Actually, on that note, should we just make a backwards-incompatible change as part of Node 7, @ob-stripe ?

Either way, I would say that we should _not_ use openapi's url param names (which are not consistent) and should instead rename the url params with a consistent format. 

For something not named, `{id}` would be good, but I think probably `{resourceId}` is best given that users can pass this as a hash. 

Thoughts?

rattrayalex-stripe:
(stepping back, I think let's pull the url params into a separate PR, and merge everything else here – what do you say, @irace-stripe ?)

irace-stripe:
@rattrayalex-stripe Yes, that sounds right to me. Will work on that.

Question: does that _also_ need we need to roll back changes like https://github.com/stripe/stripe-node/pull/606/files#diff-82d7692da4c2b29e9c7c59e40e28531bL16?

lib/resources/SubscriptionScheduleRevisions.js
```diff
@@ -2,19 +2,8 @@
...
- *
- * This streamlines the API specifically for the case of accessing a revision
- * on a returned subscription schedule object.
- *
- * E.g. scheduleObject.revisions.retrieve(revisionId)
- * (As opposed to the also-supported stripe.subscriptionSchedules.retrieveRevision(scheduleId,
- * revisionId))
- */
 module.exports = StripeResource.extend({
   path: 'subscription_schedules/{scheduleId}/revisions',
```
irace-stripe: @rattrayalex-stripe Is it weird that we’re standardizing on leading `/` for methods paths but not for resource paths?

rattrayalex-stripe: IMO we should have leading slashes on both, yes

rattrayalex-stripe:
> Question: does that also need we need to roll back changes like https://github.com/stripe/stripe-node/pull/606/files#diff-82d7692da4c2b29e9c7c59e40e28531bL16?

If that's the direction we want to go, yes.

irace-stripe:
> If that's the direction we want to go, yes.

Noted. Those method paths might be far more painful to override, unfortunately

lib/resources/CreditNotes.js
rattrayalex-stripe: https://github.com/stripe/stripe-node/blob/master/test/resources/CreditNotes.spec.js#L62-L65 seems to give us coverage – if the leading slash was problematic, it would fail since the request would include a `//`.

rattrayalex-stripe:
Yeah, I'd rather not override, just want to decide whether, for non-nested resources, we want `/foo/{id}` or `/foo/{fooId}`

irace-stripe:
> Yeah, I'd rather not override, just want to decide whether, for non-nested resources, we want `/foo/{id}` or `/foo/{fooId}`

Unifying on either one would mean breaking public API for the other 😐

rattrayalex-stripe:
Yep! Now is the time to do it.

PR #625, opened 2019-05-10
Remove "curried" nested resources and manually specified urlParams

rattrayalex-stripe:
r? @ob-stripe 
cc @irace-stripe @paulasjes-stripe @robz-stripe @remi-stripe 
cc @stripe/api-libraries 

This removes support for "curried" url parameters on nested resources, like this:

```js
const transferReversals = new resources.TransferReversal(stripe, {transferId: 'tr_123'});
transferReversals.update('trr_123', params);
```
and along with it, removing the associated resource objects (ie; `stripe.transferReversals` no longer exists) on the grounds that they were not usable outside of the curried case. 

You can still do this instead:
```js
stripe.transters.updateReversal('tr_123', 'tr_123', params);
```

This applies to all nested resources, namely: 
- `Persons`
- `LoginLinks`
- `SubscriptionScheduleRevisions`
- `ApplicationFeeRefunds`
- `TaxIds`
- `TransferReversals`

This means the names of url parameters is no longer meaningful, and can be changed without compatibility concerns. We can even have two url parameters named `id` if we like (though I think that would be less clear).

Because of this, and because we no longer have any `path`s specified as a function, I was also able to remove `urlParams` from the stripeMethods, inferring them with a regex instead (though their name values are not actually used; only the number of url params is meaningful).

Note that we briefly experimented with support for this pattern:
```js
stripe.transferReversals.update('tr_123', 'trr_123', params);
```
and ultimately decided against it for now; we may reconsider in the future. Thoughts from the community welcome!


lib/makeRequest.js
```diff
@@ -1,14 +1,10 @@
 'use strict';
 
 const utils = require('./utils');
-const OPTIONAL_REGEX = /^optional!/;
 
 function getRequestOpts(self, requestArgs, spec, overrideData) {
   // Extract spec values with defaults.
-  const commandPath =
-    typeof spec.path == 'function'
```
rattrayalex-stripe: we had no `path`s specified as functions anymore

lib/makeRequest.js
```diff
@@ -1,14 +1,10 @@
 'use strict';
 
 const utils = require('./utils');
-const OPTIONAL_REGEX = /^optional!/;
```
rattrayalex-stripe: we had no url params specified as optional anymore

lib/makeRequest.js
```diff
@@ -1,59 +1,31 @@
...
-      if (isOptional) {
-        urlData[param] = '';
-        continue;
-      }
-
-      path = self.createResourcePathWithSymbols(spec.path);
-      throw new Error(
-        `Stripe: Argument "${
-          urlParams[i]
-        }" required, but got: ${arg} (on API request to \`${requestMethod} ${path}\`)`
```
rattrayalex-stripe: this error wasn't as good as the other, combined it.

lib/makeRequest.js
```diff
@@ -1,59 +1,31 @@
...
-    const arg = args[0];
-
-    let param = urlParams[i];
-
-    const isOptional = OPTIONAL_REGEX.test(param);
-    param = param.replace(OPTIONAL_REGEX, '');
-
-    if (param == 'id' && typeof arg !== 'string') {
-      path = self.createResourcePathWithSymbols(spec.path);
+  const urlData = urlParams.reduce((urlData, param) => {
```
rattrayalex-stripe: Now that we're on Node 6+, we can use "modern" features like `.reduce`

lib/resources/Disputes.js
```diff
@@ -8,5 +8,5 @@ module.exports = StripeResource.extend({
 
   includeBasic: ['list', 'retrieve', 'update'],
 
-  close: stripeMethod({method: 'POST', path: '/{id}/close', urlParams: ['id']}),
+  close: stripeMethod({method: 'POST', path: '/{id}/close'}),
```
robz-stripe: Should we break this out onto multiple lines like we do with most of the `stripeMethod()` calls? Purely a style thing.

lib/resources/Orders.js
```diff
@@ -8,11 +8,10 @@ module.exports = StripeResource.extend({
 
   includeBasic: ['create', 'list', 'retrieve', 'update'],
 
-  pay: stripeMethod({method: 'POST', path: '/{id}/pay', urlParams: ['id']}),
+  pay: stripeMethod({method: 'POST', path: '/{id}/pay'}),
```
robz-stripe: Same question here on one vs. multiple lines.

lib/resources/Disputes.js
rattrayalex-stripe: Good call! prettier collapsed them but we can uncollapse.

rattrayalex-stripe: cc @irace-stripe we may want to try to indicate to prettier-poet that these _should_ break.

rattrayalex-stripe: I think `<Node.Object entries={} break />` would be a useful API, since prettier generally allows objects to be configured to be broken

ob-stripe:
Let's wait until we reach a decision re. nested resources, but this looks pretty great to me!

rattrayalex-stripe:
Per internal discussion document, I've reverted my change adding support for `stripe.transferReversals.update('tr_123', 'trr_123', params);` which I feel good about.

rattrayalex-stripe:
Per IRL with @ob-stripe merging this.

PR #626, opened 2019-05-10
Extract resources file (to aid with code generation)

irace-stripe:
In the process, removed a few comments which I didn’t think were necessary (but please let me know if you feel otherwise)


rattrayalex-stripe:
Merged with my branch which just landed

rattrayalex-stripe:
cc @stripe/api-libraries

PR #627, opened 2019-05-14
Add listLineItems and listUpcomingLineItems methods

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries 

We added these two methods in the v7 branch. I'd like to add them to the last 6.x version to make the upgrade path a bit easier, and so that people who need or want to stay on the 6.x branch for a bit can still use these methods (which will be the officially documented ones in the API reference).

### DIFF
--- a/lib/resources/Invoices.js
+++ b/lib/resources/Invoices.js
@@ -27,6 +27,18 @@ module.exports = StripeResource.extend({
     urlParams: ['invoiceId'],
   }),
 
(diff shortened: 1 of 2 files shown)


ob-stripe:
Self-approving 🙈

ob-stripe:
Released as 6.35.0.

rattrayalex-stripe:
post-hoc

LGTM

PR #629, opened 2019-05-18
Several changes to reflect generated version

rattrayalex-stripe:
There are a few categories of changes here: 

1. Renamed url parameters to match openapi (eg; `/{accountId}/foo` -> `/{account}/foo`) 
2. Reordering methods to be alphabetical
3. Removing some comments

The one possibly-meaningful change is the addition of `path: '',` to the `Account` resource, basically for expediency. I believe it should not have any runtime effect, and the test suite seems to confirm this. 

r? @ob-stripe


lib/resources/Accounts.js
```diff
@@ -3,8 +3,19 @@
 const StripeResource = require('../StripeResource');
 const stripeMethod = StripeResource.method;
 
+// Since path can either be `account` or `accounts`, support both through stripeMethod path;
 module.exports = StripeResource.extend({
-  // Since path can either be `account` or `accounts`, support both through stripeMethod path
+  path: '',
+
+  capabilities: stripeMethod({
```
rattrayalex-stripe: oops, need to pull this out

lib/resources.js
```diff
@@ -3,9 +3,9 @@
 const resourceNamespace = require('./ResourceNamespace');
 
 module.exports = {
+  Accounts: require('./resources/Accounts'),
   // Support Accounts for consistency, Account for backwards compatibility
   Account: require('./resources/Accounts'),
-  Accounts: require('./resources/Accounts'),
```
ob-stripe: Why does `Accounts` appear before `Account` now?

rattrayalex-stripe: That was just the simplest way to do the sorting, I think. I don't think it's quite as good, hopefully we can change it back later (was gonna be a bit annoying I think)

PR #634, opened 2019-05-27
Remove most instances of 'self' in favor of arrow functions

paulasjes-stripe:
Attempt 2, created a new branch as the old one was based off of `v7.0.0` and would have been a pain to detangle.

Leaning pretty hard on the tests here, which all pass in Node v6.


ob-stripe:
Thanks Paul! Looks great to me, but I'll defer to Alex.

PR #649, opened 2019-06-21
Enable request latency telemetry by default

ob-stripe:
r? @brandur-stripe @rattrayalex-stripe
cc @stripe/api-libraries

Cf. https://github.com/stripe/stripe-ruby/pull/800.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ See the [Node API docs](https://stripe.com/docs/api/node#intro).
 
 Install the package with:
 
-    npm install stripe --save
+```sh
+npm install stripe --save
+```
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


paulasjes-stripe:
LGTM

rattrayalex-stripe:
Can we add more detail about what sorts of telemetry are taking place?
Otherwise,

LGTM

brandur-stripe:
> Can we add more detail about what sorts of telemetry are taking place?

IMO the call out in the README for "request latency telemetry" is probably good enough, but I guess it can't hurt to be as explicit anywhere we can.

LGTM.

ob-stripe:
Released as 7.3.0.

PR #658, opened 2019-07-12
Update README retry code sample to use two retries

brandur:
While looking at retry code samples across client libraries recently, we
happened to notice that all the other ones use two retries as an example
while stripe-node uses one.

Definitely not adverse to changing this number everywhere at some point
if we think it's right move, but for now just doing the incremental
improvement of making stripe-node consistent with the other libraries.

r? @brandur-stripe (Tiny change. Going to self-approve this one.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ This will retry requests `n` times with exponential backoff if they fail due to 
 
 ```js
 // Retry a request once before giving up
-stripe.setMaxNetworkRetries(1);
+stripe.setMaxNetworkRetries(2);
 ```
 
 ### Examining Responses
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -176,7 +176,7 @@ This will retry requests `n` times with exponential backoff if they fail due to
 
 ```js
 // Retry a request once before giving up
```
ob-stripe:
```suggestion
// Retry a request twice before giving up
```

brandur-stripe:
Welp, I guess that teaches me a lesson about why self approvals are a bad idea haha. Nothing gets by you Olivier ;) Thanks!

ob-stripe:
Haha, you're welcome :)

PR #662, opened 2019-07-25
Remove `safe-buffer` dependency

tinovyatkin:
`safe-buffer` does nothing when native `Buffer.from` API is available and it's available since Nodes.JS 5.10 (https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array), so, there is no sense to bring it as production dependency (one of four) when your declared minimum supported Node.JS version is 6.x.

### DIFF
--- a/lib/MultipartDataGenerator.js
+++ b/lib/MultipartDataGenerator.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const Buffer = require('safe-buffer').Buffer;
 const utils = require('./utils');
 
 // Method for formatting HTTP body for the multipart/form-data specification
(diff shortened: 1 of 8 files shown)


brandur-stripe:
Thanks @tinovyatkin!

@rattrayalex-stripe Want to take a quick look to confirm this one? The [safe-buffer](https://www.npmjs.com/package/safe-buffer) project page seems to confirm that it's indeed safe to remove.

rattrayalex-stripe:
Gave everything a quick look and it does indeed seem this is safe. 

Thanks so much @tinovyatkin !

PR #663, opened 2019-07-25
Handle HTTP headers as case insensitive 

tinovyatkin:
By HTTP standard headers are case insensitive, so checking for existing `Idempotency-Key` like `headers.hasOwnProperty('Idempotency-Key')` is wrong. Should we fix that?

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -341,16 +341,25 @@ StripeResource.prototype = {
 
       // If this is a POST and we allow multiple retries, set a idempotency key if one is not
       // already provided.
+      let idempotencyKey;
       if (method === 'POST' && this._stripe.getMaxNetworkRetries() > 0) {
-        if (!headers.hasOwnProperty('Idempotency-Key')) {
-          headers['Idempotency-Key'] = uuid();
+        if (
+          !Object.keys(headers).find((header) => {
+            if (header.toLowerCase() === 'idempotency-key') {
+              idempotencyKey = header[header];
+              return true;
+            }
+            return false;
+          })
+        ) {
+          headers['Idempotency-Key'] = idempotencyKey = uuid();
         }
       }
 
       const requestEvent = utils.removeEmpty({
         api_version: apiVersion,
         account: headers['Stripe-Account'],
-        idempotency_key: headers['Idempotency-Key'],
+        idempotency_key: idempotencyKey,
         method,
         path,
       });
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -7,6 +7,7 @@ const uuid = require('uuid/v4');
 
 const utils = require('./utils');
 const Error = require('./Error');
+const Headers = require('./http-headers');
 
 const defaultHttpAgent = new http.Agent({keepAlive: true});
 const defaultHttpsAgent = new https.Agent({keepAlive: true});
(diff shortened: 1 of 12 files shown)


paulasjes-stripe:
We always set the header name to use the format `Idempotency-Keys` [when a key is provided](https://github.com/stripe/stripe-node/blob/a5885d7447c57eddee2195286b1fd70218feb43e/lib/utils.js#L138-L140) in a request, however it can't hurt to be extra safe.

Could you add tests for this change? They should go under the Network Retry section [here](https://github.com/stripe/stripe-node/blob/a5885d7447c57eddee2195286b1fd70218feb43e/test/StripeResource.spec.js#L103).

lib/StripeResource.js
```diff
@@ -341,16 +341,25 @@ StripeResource.prototype = {
...
+        if (
+          !Object.keys(headers).find((header) => {
+            if (header.toLowerCase() === 'idempotency-key') {
+              idempotencyKey = headers[header];
+              return true;
+            }
+            return false;
+          })
+        ) {
+          headers['Idempotency-Key'] = idempotencyKey = uuid();
```
paulasjes-stripe: If we're doing this for `Idempotency-Key`, then we should do the same for `Stripe-Version` and `Stripe-Account` https://github.com/stripe/stripe-node/blob/a5885d7447c57eddee2195286b1fd70218feb43e/lib/utils.js#L141-L146

Can you move this functionality to a helper function in `utils` and reuse it for the version and account headers?

tinovyatkin: Here we are checking `params`, not `headers`, that's not the same and may stay as it.

tinovyatkin: There is no problem with setter of headers, only getter of headers must be case insensitive.

tinovyatkin: What I see weird here, is that you are using hardcoded string like this at setters and then lowecased hardcoded strings at getters:
https://github.com/stripe/stripe-node/blob/a5885d7447c57eddee2195286b1fd70218feb43e/lib/StripeResource.js#L127
It looks like a bad style and human errors prone. I would prefer move all private headers to a `http-headers.js` file as lowercase constants and use constants everywhere. Are you OK with that?

rattrayalex-stripe: While I don't think a separate constants file is necessary, I think that's fine!

rattrayalex-stripe: I do think we should have a `getHeader(header, 'Header-Name')` function in utils, and use that to check whether idempotency-key is already defined, and also to get the stripe-version/stripe-account values.

rattrayalex-stripe: (Alternatively, we could normalize the headers hash to be all lower-case keys prior to this processing)

tinovyatkin: > While I don't think a separate constants file is necessary, I think that's fine!

You already have around 10 private HTTP headers sprinkled all around SDK. Some of them are prefixed with `X-Stripe-Client-Telemetry`, some doesn't like `Stripe-Signature` or `Stripe-Version`. Having them all in the same file will set good conventions how to add new one and enforce constants use to prevent any errors in future version regarding them. If you are OK, I'll try to do that today in this PR.

rattrayalex-stripe: Sure, if you think it would help readability to users, please do!

rattrayalex-stripe:
Hey Tino, Just wanted to say thanks again for all your work here. I'm pretty busy today but hope to give this and #661 a thorough review Monday or Tuesday; appreciate your patience!

rattrayalex-stripe:
Trying to give this a thorough, careful review made my head spin because our headers were just all over the place (our fault, not yours) so I threw together #667 which ended up obviating this, so closing. 

I know you were looking forward to exporting constants for the headers, but on second thought I'd rather not treat this as a public API, and with #667 it's at least easy to see them in one place. I hope you understand.

Thanks for all your terrific work here @tinovyatkin !

PR #664, opened 2019-07-26
Expose and use `once`

tinovyatkin:
Fixes a misspelling in EventEmitter property declaration, exposes `once` method and uses `once` where it's more appropriate than `on`

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -113,7 +113,7 @@ StripeResource.prototype = {
       res.on('data', (chunk) => {
         response += chunk;
       });
-      res.on('end', () => {
+      res.once('end', () => {
         const headers = res.headers || {};
         // NOTE: Stripe responds with lowercase header names/keys.
 
@@ -365,7 +365,7 @@ StripeResource.prototype = {
 
       req.setTimeout(timeout, this._timeoutHandler(timeout, req, callback));
 
-      req.on('response', (res) => {
+      req.once('response', (res) => {
         if (this._shouldRetry(res, requestRetries)) {
           return retryRequest(makeRequest, apiVersion, headers, requestRetries);
         } else {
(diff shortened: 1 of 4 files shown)


lib/stripe.js
```diff
@@ -43,10 +43,11 @@ function Stripe(key, version) {
     value: new EventEmitter(),
     enumerable: false,
     configurable: false,
-    writeable: false,
+    writable: false,
   });
```
tinovyatkin: 🧐

lib/StripeResource.js
```diff
@@ -381,13 +381,16 @@ StripeResource.prototype = {
         }
       });
 
-      req.on('socket', (socket) => {
+      req.once('socket', (socket) => {
```
rattrayalex-stripe: I'm not 100% sure that `'socket'` event can only be sent once, and the body of the function seems to imply it might be more than once. Do you know of a guarantee that this is at-most-once?

lib/stripe.js
rattrayalex-stripe: heh, oops! Good catch.

lib/StripeResource.js
tinovyatkin: Me neither :-) You never can be 100% sure in something in software. But look at the code - it checks a boolean flag `socket.connecting` (https://nodejs.org/api/net.html#net_socket_connecting) and in case of `true` waits for connection to be established by setting other event listener (`socket.once`). 
In both cases this function are finishing with `res.end()` - so, I'm pretty sure it's designed to be running by `once`.
 I can keep an `on`, as you wish, but see no error in this refactoring.

tinovyatkin: In fact, not, keeping `on` is an error. Just trace what will happen in this event handler will be called more than once - it will attempt to write to the `req` after calling `req.end` in previous call or in `socket.once` event handler.

rattrayalex-stripe: Okay great, I think that looks right. Thanks for digging here.

PR #665, opened 2019-07-27
remove `lodash.isplainobject`

tinovyatkin:
`lodash.isplainobject` is not a huge dependency ([install size](https://packagephobia.now.sh/result?p=lodash.isplainobject)), but it's one of 3 productions dependencies and real overkill at `utils` - what do you really want check there is that an object can be safely serialised to JSON, not to check prototypes, etc. I think 8 lines function will do a better and faster job there while saving installation and startup time.


lib/utils.js
```diff
@@ -5,19 +5,38 @@ const exec = require('child_process').exec;
...
+    } catch (_) {
+      return false;
+    }
+  },
+
+  isOptionsHash(o) {
+    return (
+      this.isSerializableObject(o) &&
+      Object.keys(o).filter(
+        (p) => OPTIONS_KEYS.has(p) && typeof o[p] === 'string'
```
tinovyatkin: We are doing here better job that original check did, because we also make sure the property value is `string`

lib/utils.js
```diff
@@ -5,19 +5,38 @@ const exec = require('child_process').exec;
 const qs = require('qs');
 const crypto = require('crypto');
 
-const hasOwn = {}.hasOwnProperty;
-const isPlainObject = require('lodash.isplainobject');
+const hasOwn = Function.bind.call(
+  Function.call,
+  Object.prototype.hasOwnProperty
+);
```
tinovyatkin: Just an extra gift, to avoid calling `hasOwn.call` and just do `hasOwn(`

lib/utils.js
```diff
@@ -5,19 +5,38 @@ const exec = require('child_process').exec;
...
 const utils = (module.exports = {
-  isOptionsHash: (o) => {
-    return isPlainObject(o) && OPTIONS_KEYS.some((key) => hasOwn.call(o, key));
+  isSerializableObject(obj) {
+    if (typeof obj !== 'object') {
+      return false;
+    }
+    try {
+      const str = JSON.stringify(obj);
+      return str === JSON.stringify(JSON.parse(str));
```
tinovyatkin: Looks brute force, but it's actually faster than `lodash.isPlainObject` loading

lib/utils.js
rattrayalex-stripe: Hmm, I'd rather not do that – if someone passes `{idempotency_key: null}` I'd rather it continue to get picked up as an option, so they can understand what's happening.

lib/utils.js
```diff
@@ -5,19 +5,38 @@ const exec = require('child_process').exec;
...
+    } catch (_) {
+      return false;
+    }
+  },
+
+  isOptionsHash(o) {
+    return (
+      this.isSerializableObject(o) &&
```
rattrayalex-stripe: I actually think we don't need to be this careful – it's fine if people pass us a class instance with the right properties, or whatever. 

I think we can get away with this:

```
isOptionsHash(o) {
  return o && typeof o === 'object' && OPTIONS_KEYS.some(key => hasOwn(o, key))
}
```
(accordingly, probably best to put `OPTIONS_KEYS` back to an array, since it's better to iterate over a known-short array than a possibly-long one).

lib/utils.js
```diff
@@ -72,9 +91,16 @@ const utils = (module.exports = {
...
+   * @param {object[]} args
+   * @returns {object}
    */
-  getDataFromArgs: (args) => {
-    if (args.length < 1 || !isPlainObject(args[0])) {
+  getDataFromArgs(args) {
+    if (
+      !Array.isArray(args) ||
+      args.length < 1 ||
+      !this.isSerializableObject(args[0])
```
rattrayalex-stripe: Again, here, I think just `!(args[0] && typeof args[0] === 'object')` is fine.

lib/utils.js
rattrayalex-stripe: Now that we're in ES6, mind clarifying to `const hasOwn = (obj, key) => {}.hasOwnProperty.call(ob, key)` ? Just a bit clearer what's going on that way.

PR #669, opened 2019-07-30
Updated utils.getDataFromArgs to support mixed options/data hash

509dave16:
Mixed options/data hashes may occur when a client is making a request on behalf of a Connect account(aka `stripe_account: "acct_<some account id>"`) and wishing to expand a particular field of the resource(e.g. `expand: ["charges.data.balance_transaction"]`. Meaning the hash would look like this in our figurative example:
```
{
  stripe_account: "acct_<some account id>",
  expand: ["charges.data.balance_transaction"]
}
```
Currently `utils.getDataFromArgs` does not support a mixed options/data hash. Meaning the above hash would result in the `expand` data arg being discarded instead of included. So in order to support mixed options/data hash, we instead look for keys that are not in `OPTIONS_KEYS` and pluck those from the hash. As a part of this plucking, we delete the data properties from the hash object, because they are not options that should be plucked in `utils.getOptionsFromArgs`. This way misleading warnings will not be emitted. We also change a previous `getDataFromArgs` test from warning about a mixed options/data hash to instead finding the data in a mixed options/data hash.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -84,26 +84,16 @@ const utils = (module.exports = {
 
     const argKeys = Object.keys(args[0]);
 
-    const optionKeysInArgs = argKeys.filter((key) =>
-      OPTIONS_KEYS.includes(key)
-    );
-
-    // In some cases options may be the provided as the first argument.
-    // Here we're detecting a case where there are two distinct arguments
-    // (the first being args and the second options) and with known
-    // option keys in the first so that we can warn the user about it.
-    if (
-      optionKeysInArgs.length > 0 &&
-      optionKeysInArgs.length !== argKeys.length
-    ) {
-      emitWarning(
-        `Options found in arguments (${optionKeysInArgs.join(
-          ', '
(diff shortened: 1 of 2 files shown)


remi-stripe:
@509dave16 The `expand` options and the `stripe_account` should never be in the same hash, they should always be in separate hashes as `expand` is an API parameter and `stripe_account` is a request configuration option, sent as a custom header.

509dave16:
@remi-stripe Thank you for the clarification on the `expand` data arg! I misinterpreted the docs reference on the `expand` data arg and the data hash, which is an optional hash that can be passed preceding the options hash argument to a stripe resource method. I'll go ahead close this pull requested. I tested the `expand` data arg on my backend based on your example and it worked. Thanks again for your help!

remi-stripe:
Glad I was able to clarify! This is definitely something we could make clearer in our docs as you're not the first one confused by the "two hashes" approach :)

PR #671, opened 2019-07-31
Remove uuid dep, include a prefix in generated idempotency keys

rattrayalex-stripe:
r? @richardm-stripe
cc @stripe/api-libraries @paulasjes-stripe @ob-stripe @brandur-stripe 

I checked and this should be well within our idempotency key length limits.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -341,6 +341,17 @@ const utils = (module.exports = {
 
     return result;
   },
+
+  /**
+   * https://stackoverflow.com/a/2117523
+   */
+  uuid4: () => {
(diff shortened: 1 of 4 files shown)


brandur-stripe:
You have my vote! Was surprised to find out that the `uuid` package was that big.

lib/utils.js
```diff
@@ -341,6 +341,17 @@ const utils = (module.exports = {
 
     return result;
   },
+
+  /**
+   * https://stackoverflow.com/a/2117523
+   */
+  uuid4: () => {
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
```
richardm-stripe: This is cute, love it

PR #672, opened 2019-08-02
Separate super-deprecated errors from base errors, fix type override, and remove name override

rattrayalex-stripe:
This is a follow-on to https://github.com/stripe/stripe-node/pull/661

It makes several small changes but the most consequential are:

1) Adding `get type()` support on `StripeError`, so subclasses have the property `new StripeFooClass({}).type === 'StripeFooClass'`, which is an important property to uphold. We missed this in #661 because our `StripeError` tests had actually been testing `GenericError` on accident, oops!
2) Remove support for magical `.name`, which turned out not to be important for backcompat (previously `.name` was just `'Error'`) – was a mistake for me to ask for that.
3) Have `StripeError` inherit directly from `Error`, and have the default export of `Error.js` be an "old-school class" to maximize backcompat without any complexity leaking into our proper base class, and to emphasize that it is totally deprecated and will be going completely away. Users should not be relying on the stripe library to extend Error.

r? @richardm-stripe
cc @stripe/api-libraries

### DIFF
--- a/lib/Error.js
+++ b/lib/Error.js
@@ -1,82 +1,33 @@
 'use strict';
 
-/**
- *  @typedef {{ message?: string, type?: string, code?: number, param?: string, detail?: string, headers?: Record<string, string>, requestId?: string, statusCode?: number }} ErrorParams
- */
-/**
- * Generic Error class to wrap any errors returned by stripe-node
- */
-class GenericError extends Error {
-  /**
-   *
-   * @param  {string} type
-   * @param {string} message
-   */
-  constructor(type, message) {
-    super(message);
-    this.type = type || this.constructor.name;
-    // Saving class name in the property of our custom error as a shortcut.
-    this.name = this.constructor.name;
+const utils = require('./utils');
 
-    // Capturing stack trace, excluding constructor call from it.
-    Error.captureStackTrace(this, this.constructor);
+/**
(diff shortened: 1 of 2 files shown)


remi-stripe:
Lurk: I don;t fully understand the changes, but we should make sure we're trying to be aligned across languages, and when we're not that we know why. We're doing something similar to make access errors better in ruby https://github.com/stripe/stripe-ruby/pull/811 and have decided that it's cleaner to do `e.error.code` instead of `e.code`. It might be what you're doing I just wanted to flag.
cc @ob-stripe @brandur-stripe

rattrayalex-stripe:
Ah, thanks! That doesn't actually affect this PR but good to know about. We don't have support for `decline_code` in stripe-node yet, so we should add that. As for `e.error.decline_code` vs `e.decline_code`, IMO we should make that decision on a per-language basis, depending on its idioms (and I'm not sure what the right answer is for Ruby or Node).

rattrayalex-stripe:
My pretty strong hunch, though, looking at this a bit more, is that we should add these properties directly to the error in stripe-node.

remi-stripe:
Fair, that's what I meant! Either we document why they are different (more idiomatic or not) or we are consistent but we should make sure to get aligned since today the 4 dynamic libs have no direct access and we're trying to fix that

PR #673, opened 2019-08-07
Add request start and end time to request and response events

Just-Sieb:
I am working on setting up distributed tracing for work and I am wanting to create spans on our Stripe calls. This change makes it easier to get accurate start and end times for the spans. Plus I think this addition should be beneficial for other people as well.

### DIFF
--- a/test/flows.spec.js
+++ b/test/flows.spec.js
@@ -324,15 +324,18 @@ describe('Flows', function() {
       const idempotencyKey = Math.random()
         .toString(36)
         .slice(2);
+      const lowerBoundStartTime = Date.now();
 
       function onRequest(request) {
-        expect(request).to.eql({
-          api_version: 'latest',
-          idempotency_key: idempotencyKey,
-          method: 'POST',
-          path: '/v1/charges',
-          account: connectedAccountId,
-        });
+        expect(request.api_version).to.equal('latest');
+        expect(request.idempotency_key).to.equal(idempotencyKey);
+        expect(request.account).to.equal(connectedAccountId);
+        expect(request.method).to.equal('POST');
+        expect(request.path).to.equal('/v1/charges');
+        expect(request.request_start_time).to.be.within(
+          lowerBoundStartTime,
+          Date.now()
+        );
 
         done();
       }
@@ -379,7 +382,13 @@ describe('Flows', function() {
         expect(response.path).to.equal('/v1/charges');
         expect(response.request_id).to.match(/req_[\w\d]/);
         expect(response.status).to.equal(402);
-        expect(response.elapsed).to.be.within(50, 30000);
+        expect(response.elapsed).to.equal(
+          response.request_end_time - response.request_start_time
+        );
+        expect(response.request_end_time).to.be.within(
+          response.request_end_time,
+          Date.now()
+        );
 
         done();
       }
--- a/README.md
+++ b/README.md
@@ -212,10 +212,11 @@ stripe.off('request', onRequest);
 ```js
 {
   api_version: 'latest',
-  account: 'acct_TEST',       // Only present if provided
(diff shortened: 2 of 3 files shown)


ob-stripe:
Hi @Just-Sieb, thanks for the PR!

This looks peaceful to me, but assigning to @rattrayalex-stripe for review.

test/flows.spec.js
```diff
@@ -326,13 +326,15 @@ describe('Flows', function() {
...
-          account: connectedAccountId,
-        });
+        expect(request.api_version).to.equal('latest');
+        expect(request.idempotency_key).to.equal(idempotencyKey);
+        expect(request.account).to.equal(connectedAccountId);
+        expect(request.method).to.equal('POST');
+        expect(request.path).to.equal('/v1/charges');
+        expect(request.request_start_time).to.be.within(
+          Date.now() - 30000,
+          Date.now()
```
ob-stripe: Just a note that in general, I don't like tests that rely on a real clock and prefer using dependency injection or even mocks/stubs to make the test more deterministic... Not sure if we have an established pattern for doing this in stripe-node. We can probably keep this as-is and revisit if the test ever becomes flaky.

Just-Sieb: I concur with you on your sentiments on tests like this. I went with this approach because I saw this pattern used for testing the elapsed time so I figured it was what Stripe expects/prefers. Perhaps just a type check of these values to ensure they are a number would be a better option?

rattrayalex-stripe: While I generally also dislike relying on the real clock, I don't think we have delightful mocking available at this time.

That said, we can probably set the starting bound to be a `Date.now()` set at the top of the test, which should be both more accurate and robust. Similarly, below, in an ideal world we'd check that `response.request_end_time` is greater than `request.request_start_time` (which would require adding an `onRequest` handler).

Just-Sieb: Good suggestions @rattrayalex-stripe, I went ahead and added the request_start_time to the response event as well as that lets me more easily implement your request_end_time greater that request_start_time test as well as improve the elapsed test to not be vulnerable to a bizarre long request. Plus I suspect for some people's usage it will be beneficial for adding logging of the start and end time of any Stripe request. I also made the set lower bound at start of test change. This should be ready for you to review again. Thanks!

test/flows.spec.js
```diff
@@ -379,7 +383,16 @@ describe('Flows', function() {
         expect(response.path).to.equal('/v1/charges');
         expect(response.request_id).to.match(/req_[\w\d]/);
         expect(response.status).to.equal(402);
-        expect(response.elapsed).to.be.within(50, 30000);
+        expect(response.elapsed).to.equal(
+          response.request_end_time - response.request_start_time
+        );
+        expect(response.request_end_time).to.be.within(
+          lowerBoundStartTime,
```
rattrayalex-stripe: sorry, tiny nit: probably a bit cleaner to use `response.request_start_time` here and then remove the `expect` below.

test/flows.spec.js
```diff
@@ -379,7 +383,16 @@ describe('Flows', function() {
         expect(response.path).to.equal('/v1/charges');
         expect(response.request_id).to.match(/req_[\w\d]/);
         expect(response.status).to.equal(402);
-        expect(response.elapsed).to.be.within(50, 30000);
+        expect(response.elapsed).to.equal(
```
rattrayalex-stripe: nice!!

test/flows.spec.js
rattrayalex-stripe: Great thinking! Thanks!

test/flows.spec.js
Just-Sieb: I went ahead and updated. This should be good for you to merge now.

PR #674, opened 2019-08-09
Refactor requestDataProcessor for File out into its own file

rattrayalex-stripe:
r? @richardm-stripe 
cc @ob-stripe @stripe/api-libraries 

This was in a codegen'd file, which lead to big/ugly diffs; we should ~never have substantial application logic in a codegen'd file.

While I was looking at the code, I refactored to make it a bit more simple/modern/unified.

### DIFF
--- a/lib/multipart.js
+++ b/lib/multipart.js
@@ -1,11 +1,14 @@
 'use strict';
 
 const utils = require('./utils');
+const {StripeError} = require('./Error');
+
+class StreamProcessingError extends StripeError {}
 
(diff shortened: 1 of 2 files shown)


richardm-stripe:
LGTM
Seems much better to me and, as far as I can reason, does the same thing. I'd be even more happy if there were a test.

rattrayalex-stripe:
FWIW there are some high-level integration tests in `test/resources/Files.spec.js` but that's certainly fair, coverage here is probably not what it should be.

Issue #690, opened 2019-09-06
GET requests use request body instead of query parameters


ajbouh:
Setting a body for GET requests appears to be a gray area in the HTTP spec.

This can cause many subtle and hard to debug issues. For example, when the request body is not properly submitted to the server, auto pagination logic goes into an infinite loop, since the starting_after parameter is lost/ignored and it will just return the first 10 values over and over again.

A possible fix would be to just append the data to the request path as query parameters. (This appears to be what `curl` does automatically.)

ob-stripe:
Thanks for the report @ajbouh. This was already a known issue but we've been ignoring it for too long. We're working on a fix and will release a new version soon.

ob-stripe:
This was fixed in 7.9.0, courtesy of @richardm-stripe!

PR #691, opened 2019-09-06
GET and DELETE requests data: body->queryParams

richardm-stripe:
r? @ob-stripe 
cc? @rattrayalex-stripe @remi-stripe 

Addresses #690 -- whenever an ApiResource method is called that corresponds to a GET or DELETE, those properties are attached to the end of the path.

I think this should be safe, assuming:
1. "Path" never already contains a query string.
2. No users are sending bodies large enough to risk hitting some sort of URL length limit.

I also changed `mocha.opts` not to bail by default. Rationale: I couldn't find a way to override this from the command-line (but it's easy to add `-b` if that's what you want), and I prefer not-bailing in cases where I need to make simple changes in many tests.

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,2 +1,1 @@
---bail
 --recursive
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -47,3 +47,5 @@ function stripeMethod(spec) {
 }
 
 module.exports = stripeMethod;
+
+module.exports = stripeMethod;
--- a/lib/makeRequest.js
+++ b/lib/makeRequest.js
@@ -46,10 +46,15 @@ function getRequestOpts(self, requestArgs, spec, overrideData) {
     spec.validator(data, {headers});
   }
 
+  const dataInQuery = spec.method === 'GET' || spec.method === 'DELETE';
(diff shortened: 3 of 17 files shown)


test/mocha.opts
```diff
@@ -1,2 +1 @@
---bail
```
rattrayalex-stripe: I know this is obnoxious during development, but we might want it during CI? not sure. I've removed it locally myself in the past as well. Maybe we should move it to the script in package.json or whatever?

richardm-stripe: My sentiment is bailing is even less desirable on CI. If 10 tests were failing only in Node 6.x or whatever, I would want to know all of them the first time around rather than discovering one by one.

richardm-stripe: Let's wait until that happens though before changing it. Adding --bail to the package.json.

rattrayalex-stripe: Ah, yeah, I think you're right. Let's just bail on bail (ie; not even include in package.json) for now, unless you think there's good reason to keep it, and can add it back granularly if we find there are times it's useful.

ob-stripe:
Released as 7.9.0.

PR #692, opened 2019-09-07
Retry based on `Stripe-Should-Retry` and `Retry-After` headers

rattrayalex-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @paulasjes-stripe @jlomas-stripe

This adds support for the `Stripe-Should-Retry` and `Retry-After` headers. 

This allows the API to specify specific cases when an automatic retry is recommended (such as `lock_timeout` errors) or not recommended (such as when our idempotency framework would replay the request anyway). 

It also allows the API to encourage slower retries.

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -228,25 +228,24 @@ StripeResource.prototype = {
       return true;
     }
 
-    if (res.headers['should-retry'] === 'false') {
+    if (res.headers['stripe-should-retry'] === 'false') {
       return false;
     }
-    if (res.headers['should-retry'] === 'true') {
+    if (res.headers['stripe-should-retry'] === 'true') {
       return true;
     }
 
-    // Retry on conflict, rate-limit, and availability errors.
-    if (
-      res.statusCode === 409 ||
-      res.statusCode === 429 ||
-      res.statusCode === 503
-    ) {
+    // Retry on conflict errors.
+    if (res.statusCode === 409) {
       return true;
     }
 
-    // Retry on 5xx's, except POST's, which our idempotency framework
-    // would just replay as 500's again anyway.
-    if (res.statusCode >= 500 && res.req._requestEvent.method !== 'POST') {
+    // Retry on 500, 503, and other internal errors.
+    //
+    // Note that we expect the stripe-should-retry header to be false
+    // in most cases when a 500 is returned, since our idempotency framework
+    // would typically replay it anyway.
+    if (res.statusCode >= 500) {
       return true;
     }
 
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -253,7 +253,7 @@ StripeResource.prototype = {
     return false;
   },
 
-  _getSleepTimeInMS(numRetries) {
+  _getSleepTimeInMS(numRetries, retryAfter) {
     const initialNetworkRetryDelay = this._stripe.getInitialNetworkRetryDelay();
     const maxNetworkRetryDelay = this._stripe.getMaxNetworkRetryDelay();
 
@@ -272,7 +272,7 @@ StripeResource.prototype = {
     // But never sleep less than the base sleep seconds.
     sleepSeconds = Math.max(initialNetworkRetryDelay, sleepSeconds);
 
-    return sleepSeconds * 1000;
+    return Math.max(sleepSeconds * 1000, retryAfter || 0);
   },
 
   _defaultIdempotencyKey(method) {
@@ -353,10 +353,16 @@ StripeResource.prototype = {
   _request(method, host, path, data, auth, options, callback) {
     let requestData;
 
-    const retryRequest = (requestFn, apiVersion, headers, requestRetries) => {
+    const retryRequest = (
+      requestFn,
+      apiVersion,
+      headers,
+      requestRetries,
+      retryAfter
+    ) => {
       return setTimeout(
         requestFn,
-        this._getSleepTimeInMS(requestRetries),
+        this._getSleepTimeInMS(requestRetries, retryAfter),
         apiVersion,
         headers,
         requestRetries + 1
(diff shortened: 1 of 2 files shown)


lib/StripeResource.js
```diff
@@ -228,21 +228,31 @@ StripeResource.prototype = {
...
+    // would typically replay it anyway.
+    if (res.statusCode >= 500) {
       return true;
     }
 
     return false;
   },
 
-  _getSleepTimeInMS(numRetries) {
+  _getSleepTimeInMS(numRetries, retryAfter) {
```
jlomas-stripe: Is it worth making sure, at some point, that `retryAfter` is actually a (reasonably-sized) positive integer?

paulasjes-stripe: Since we're doing a `Math.max` on this value a few lines later, I don't think we need to explicitly check this value. We should however set a default value here `retryAfter = 0`

lib/StripeResource.js
```diff
@@ -261,6 +271,9 @@ StripeResource.prototype = {
     // But never sleep less than the base sleep seconds.
     sleepSeconds = Math.max(initialNetworkRetryDelay, sleepSeconds);
 
+    // And never sleep less than the time the API asks us to wait
+    sleepSeconds = Math.max(sleepSeconds, retryAfter || 0);
```
paulasjes-stripe: Instead of doing `retryAfter || 0` can we just set a default parameter for `retryAfter` in the function declaration?

lib/StripeResource.js
```diff
@@ -394,15 +413,27 @@ StripeResource.prototype = {
 
       req.once('response', (res) => {
         if (this._shouldRetry(res, requestRetries)) {
-          return retryRequest(makeRequest, apiVersion, headers, requestRetries);
+          return retryRequest(
+            makeRequest,
+            apiVersion,
+            headers,
+            requestRetries,
+            ((res || {}).headers || {})['retry-after']
```
paulasjes-stripe: This makes my head hurt a bit. Also won't it throw an error if  `headers` doesn't exist? Maybe something like:

`const retryAfter = res.hasOwnProperty('headers') ? res.headers['retry-after'] || null : null;`

Wish optional chaining was already in JS 😭

jlomas-stripe: That's funny.  I love this and it validates so many of my life decisions. 😂 

It won't throw because it takes either res or an empty object, and then either that thing or an empty object, and so eventually gets `retry-after` from either `res['headers']` or an empty object.

It's Javascript at is finest. 🥂

lib/StripeResource.js
rattrayalex-stripe: I'd rather not, since default params don't work with `null`. `Math.max` _does_ work with `null`, though it behaves unexpectedly (to me at least) with `NaN` (`Math.max(2, NaN) === NaN`). Using `|| 0` is more robust there, even if it isn't an expected program state.

lib/StripeResource.js
rattrayalex-stripe: heh, yeah, this is basically how I do optional chaining until it's a Real Feature.

lib/StripeResource.js
rattrayalex-stripe: Eh, I think this was a good point in retrospect, and also helped resolve the `|| 0` debate below. Added an integer check and <60, since we should definitely never ask for something longer than that.

lib/StripeResource.js
```diff
@@ -228,21 +228,31 @@ StripeResource.prototype = {
       return true;
     }
 
-    // Retry on conflict and availability errors.
-    if (res.statusCode === 409 || res.statusCode === 503) {
+    if (res.headers && res.headers['stripe-should-retry'] === 'false') {
```
brandur-stripe: Can we put some docs on this to describe generally what `Stripe-Should-Retry` is?

lib/StripeResource.js
```diff
@@ -228,21 +228,31 @@ StripeResource.prototype = {
...
+      return false;
+    }
+    if (res.headers && res.headers['stripe-should-retry'] === 'true') {
```
brandur-stripe: How does the interaction between `Stripe-Should-Retry` and `Retry-After` work?

I guess based on this condition being placed so early, if we're going to send `Retry-After`, then `Stripe-Should-Retry` is set to `false` or not included? If so, doesn't that mean we have to fall back on other heuristics to decide whether to retry (like status code)?

I wonder if we should have it so that if there's a `Retry-After`, `Stripe-Should-Retry` is still set to an appropriate value that indicates whether a retry is appropriate. If we went down that route, we'd have to include a `&& !res.headers['stripe-retry-after']` in this check.

lib/StripeResource.js
```diff
@@ -261,6 +271,11 @@ StripeResource.prototype = {
     // But never sleep less than the base sleep seconds.
     sleepSeconds = Math.max(initialNetworkRetryDelay, sleepSeconds);
 
+    // And never sleep less than the time the API asks us to wait, assuming it's a reasonable ask.
+    if (Number.isInteger(retryAfter) && retryAfter <= 60) {
```
brandur-stripe: Can we constant-ize `60` here to make it more literate? Maybe `MAX_RETRY_AFTER_WAIT`?

lib/StripeResource.js
rattrayalex-stripe: Good idea, will do

lib/StripeResource.js
rattrayalex-stripe: > if we're going to send Retry-After, then Stripe-Should-Retry is set to false or not included?

I assume you meant "true" not "false, and yes: I can share the internal doc, but our API should hold the invariant that if `Retry-After` is set, `Stripe-Should-Retry` should not be set to `false`. 

Were that to happen, though, I think `Stripe-Should-Retry: false` should take precedence and block a `Retry-After: X`, given that the likely intent of the latter is to ask clients to wait a little longer before retrying, and the intent of the former is to ask them not to retry at all (eg; because it will be futile or damaging).

In general, they're pretty orthogonal; technically, these should be separate PR's (I'm just lumping them together for convenience, since they came to mind at the same time). 

Note that I think we'll ~always want to fall back to status code checks for at least some cases (eg; 500, 503, 504) in case the parts of our api stack that set `Stripe-Should-Retry` are not hit. 

I could see a case for dropping the code for `409` handling here and moving that into the API...

However, in general, we should expect the client to retry in some cases that the server does not make a recommendation, for example, 429's, where we do not retry but the client may choose to (but would hopefully respect Retry-After).

rattrayalex-stripe:
Self-approving after adopting recommended changes

PR #699, opened 2019-09-25
Add request-specific fields from raw error to top level error

richardm-stripe:
Per request of @remi-stripe (https://jira.corp.stripe.com/secure/RapidBoard.jspa?rapidView=1346&projectKey=DX&view=planning&selectedIssue=DX-4753&quickFilter=3988)
It is sometimes important to access properties such as `error.raw.payment_intent`, `error.raw.payment_method`, etc. But "raw" means "here be dragons, don't rely on anything here". This PR copies these properties from `raw` to the top level error object, to indicate that they are properly part of our interface and remove the need for users to access "raw".

### DIFF
--- a/test/Error.spec.js
+++ b/test/Error.spec.js
@@ -60,6 +60,15 @@ describe('Error', () => {
       ).to.be.instanceOf(Error.StripeIdempotencyError);
     });
 
+    it('copies whitelisted properties', () => {
+      const e = new Error.StripeError({
+        charge: 'foo',
+        unknown_prop: 'bar',
+      });
+      expect(e).to.have.property('charge', 'foo');
+      expect(e).not.to.have.property('unknown_prop', 'bar');
+    });
+
     it('Pulls in headers', () => {
       const headers = {'Request-Id': '123'};
       const e = Error.StripeError.generate({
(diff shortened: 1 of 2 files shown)


test/Error.spec.js
```diff
@@ -60,6 +60,15 @@ describe('Error', () => {
       ).to.be.instanceOf(Error.StripeIdempotencyError);
     });
 
+    it('copies whitelisted properties', () => {
+      const e = new Error.StripeError({
+        charge: 'foo',
+        unknown_prop: 'bar',
+      });
+      expect(e).to.have.property('charge', 'foo');
+      expect(e).not.to.have.property('unknown_prop', 'bar');
```
ob-stripe: What happens if you try to access a property that _can_ exist but was not returned by the API, e.g. `decline_code` in this example?

richardm-stripe: `undefined`, but I'll express that in the test case. Thanks!
ptal @ob-stripe

richardm-stripe:
Oops, probably should have
cc @stripe/api-libraries

rattrayalex-stripe:
We should probably document this somewhere. https://github.com/stripe/stripe-node/wiki/Error-Handling at a minimum, perhaps also at stripe.com/docs/api

PR #703, opened 2019-10-04
New config object

paulasjes-stripe:
This is a long running branch for the changes outlined in https://github.com/stripe/stripe-node/issues/640

Steps to complete before this can be merged:

- [x] Move `setApiVersion` to be set via a config object in the Stripe object initialization, with the intention of deprecating `stripe.setApiVersion`
- [x] Move other setters like `setMaxNetworkRetires` to be set via the config object
- [x] Allow certain properties (e.g. network retries, timeout) to be set on a per-request basis as well as a global setting
- [x] Update documentation and README to reflect new config object

~I'll open this up for review once the previous steps are completed.~

This is now ready for review! All the code in the PR was previously reviewed (see branches that were merged into this one). This should theoretically be an easy thumbs up.

r? @rattrayalex-stripe @ob-stripe 
cc: @stripe/api-libraries

PR #706, opened 2019-10-07
New config object base and apiVersion

paulasjes-stripe:
Part 1 of https://github.com/stripe/stripe-node/pull/703

This adds the ability to pass in a configuration object as the 2nd parameter of `Stripe`. For example:
```js
const stripe = new Stripe('sk_test_123', {
  apiVersion: '2019-09-09',
)};
```

This allows the old behavior to still work:
```js
const stripe = new Stripe('sk_test_123', '2019-09-09');
```

Additionally this behavior will also still work:
```js
const stripe = new Stripe('sk_test_123');
stripe.setApiVersion('2019-09-09');
```

With the idea being of eventually deprecating `stripe.setApiVersion` in favor of the config object.

r? @rattrayalex-stripe @ob-stripe @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -263,25 +263,17 @@ Stripe.prototype = {
   },
 
   _getPropsFromConfig(config) {
-    // config can be an object or a string
-    const isString = typeof config === 'string';
-    const isObject = config === Object(config) && !Array.isArray(config);
-
     // If config is null or undefined, just bail early with no props
     if (!config) {
       return {};
     }
 
+    // config can be an object or a string
+    const isString = typeof config === 'string';
+    const isObject = config === Object(config) && !Array.isArray(config);
+
     if (!isObject && !isString) {
       throw new Error('Config must either be an object or a string');
-    }
-
-    // if an api version is provided, first validate that
-    const version = isString ? config : config.apiVersion;
-    if (version && !utils.validateApiVersion(version)) {
-      throw new Error(
-        'Api version string should be in the format YYYY-MM-DD or "latest"'
-      );
     }
 
     // If config is a string, we assume the old behavior of passing in a string representation of the api version
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -267,6 +267,11 @@ Stripe.prototype = {
     const isString = typeof config === 'string';
     const isObject = config === Object(config) && !Array.isArray(config);
 
+    // If config is null or undefined, just bail early with no props
+    if (!config) {
+      return {};
+    }
+
     if (!isObject && !isString) {
       throw new Error('Config must either be an object or a string');
     }
@@ -279,11 +284,11 @@ Stripe.prototype = {
       );
     }
 
-    let props = {};
-
     // If config is a string, we assume the old behavior of passing in a string representation of the api version
     if (isString) {
-      props.apiVersion = config;
+      return {
+        apiVersion: config,
+      };
     }
 
     // If config is an object, we assume the new behavior and make sure it doesn't contain any unexpected values
@@ -299,11 +304,9 @@ Stripe.prototype = {
           )}`
         );
       }
-
-      props = config;
     }
 
-    return props;
+    return config;
   },
 };
 
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -27,6 +27,7 @@ Stripe.MAX_NETWORK_RETRY_DELAY_SEC = 2;
 Stripe.INITIAL_NETWORK_RETRY_DELAY_SEC = 0.5;
 
 const APP_INFO_PROPERTIES = ['name', 'version', 'url', 'partner_id'];
+const ALLOWED_CONFIG_PROPERTIES = ['apiVersion'];
 
 const EventEmitter = require('events').EventEmitter;
 const utils = require('./utils');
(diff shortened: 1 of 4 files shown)


lib/stripe.js
```diff
@@ -251,6 +261,50 @@ Stripe.prototype = {
...
   },
+
+  _getPropsFromConfig(config) {
+    // config can be an object or a string
+    const isString = typeof config === 'string';
+    const isObject = config === Object(config) && !Array.isArray(config);
+
+    if (!isObject && !isString) {
+      throw new Error('Config must either be an object or a string');
+    }
```
richardm-stripe: nit: What do you think about checking *all four* cases (undefined, string, object, other) in a single place, and producing the canonical form {version: string} immediately, then performing the version string validation along with the other validation?

lib/stripe.js
```diff
@@ -251,6 +261,50 @@ Stripe.prototype = {
       this[utils.pascalToCamelCase(name)] = new resources[name](this);
     }
   },
+
+  _getPropsFromConfig(config) {
+    // config can be an object or a string
+    const isString = typeof config === 'string';
+    const isObject = config === Object(config) && !Array.isArray(config);
+
+    if (!isObject && !isString) {
```
rattrayalex-stripe: part of me thinks we should accept `null` as valid, and treat it as the same as `undefined` (that is, don't set `apiVersion`)

lib/stripe.js
```diff
@@ -251,6 +261,50 @@ Stripe.prototype = {
...
+      throw new Error(
+        'Api version string should be in the format YYYY-MM-DD or "latest"'
+      );
+    }
+
+    let props = {};
+
+    // If config is a string, we assume the old behavior of passing in a string representation of the api version
+    if (isString) {
+      props.apiVersion = config;
```
rattrayalex-stripe: nit, I think it might make sense to early-return `{apiVersion: config}` here, since it'll be impossible to set other properties later anyway. 

Then, later on, we can just `return config`

lib/utils.js
```diff
@@ -356,6 +356,10 @@ const utils = (module.exports = {
       return v.toString(16);
     });
   },
+
+  // Only allow versions in the format YYYY-MM-DD
```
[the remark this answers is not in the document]
paulasjes-stripe: TIL! Is it conceivable that api versions could take the form of `foo=bar YYYY-MM-DD` or would we only expect additional parameters after the date string?

lib/stripe.js
paulasjes-stripe: Damnit that's so much better

lib/utils.js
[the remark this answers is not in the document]
paulasjes-stripe: Also is something like this possible: `YYYY-MM-DD foo_beta=v1 bar_beta=v2`?

I'm thinking that maybe the regex just checks for the date at the beginning of the string and doesn't care about what comes after.

lib/stripe.js
paulasjes-stripe: I rejigged a bit and think this "return fast and early" strategy works, let me know what you think.

paulasjes-stripe:
Addressed your feedback and added more tests for edge cases, ptal @rattrayalex-stripe @richardm-stripe

lib/utils.js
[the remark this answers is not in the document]
rattrayalex-stripe: I think I'd rather forego regex checks unless they're really necessary, and rely on the api to tell the user that the string doesn't fit. Just buys us the most flexibility. Thoughts?

lib/stripe.js
```diff
@@ -251,6 +261,53 @@ Stripe.prototype = {
...
+
+    // If config is a string, we assume the old behavior of passing in a string representation of the api version
+    if (isString) {
+      return {
+        apiVersion: config,
+      };
+    }
+
+    // If config is an object, we assume the new behavior and make sure it doesn't contain any unexpected values
+    if (isObject) {
```
rattrayalex-stripe: nit, this conditional is unnecessary (fine to leave though)

lib/utils.js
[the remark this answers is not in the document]
paulasjes-stripe: I like the concept of failing as soon as possible if something is obviously wrong. If I were to initialize stripe-node like so:

```js
const stripe = new Stripe('sk_test_123', 'bananas');
```

Then stripe-node would happily return a version of Stripe with a bogus api version. The error won't occur until we actually attempt to make a request.

I think that this sort of clearly wrong input should be caught and handled before any request is made. 

Saying that, if we have too many API version format options and edge cases, then I'm happy to let the API take care of that for us.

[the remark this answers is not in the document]
rattrayalex-stripe: Ah sorry, missed this comment earlier – earlier error messages are certainly better than later ones, but the error message from the API about an invalid version string should come pretty quick, no? 

Furthermore the client can't have enough information to make an accurate decision – we can't disallow a valid date which isn't a valid api version, and may indeed allow `"bananas"` at one point in the future for a very special beta program!

paulasjes-stripe:
@rattrayalex-stripe You've convinced me that the regex is a bad idea, so I removed it. Have a look now

richardm-stripe:
lgtm

PR #709, opened 2019-10-15
README: imply context of stripe-node

dcousens:
This is subjective.
I read it several times as the node API docs (e.g https://nodejs.org/api/),  and somehow skipped it.

Up to you whether this is any better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ client side. This package should not be used for that purpose.
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


dcousens:
On a separate note,  the link re-directs to https://stripe.com/docs/api which include examples for several languages now, so, implying the link is `Node` specific is confusing anyway.

brandur-stripe:
Works for me so going to merge this.

@stripe/api-libraries We should probably come up with a single standardized line like this one to use across every language.

PR #714, opened 2019-10-22
Allow network retries and timeout to be set on a per-request basis

paulasjes-stripe:
Added network retries/timeout to be set on per-request basis and some general cleanup.

# Per-request settings

This now works:

```js
const charge = await stripe.charges.create({
  amount: 1000,
  currency: 'usd',
  source: 'tok_visa',
}, { 
  maxNetworkRetries: 1,
  timeout: 500,
});
```

`maxNetworkRetries` and `timeout` can still be set globally via the config object:

```js
const stripe = Stripe('sk_test_123', {
  maxNetworkRetries: 2,
  timeout: 1000,
});
```

Additionally `timeout` can still be globally set at any time:
```js
stripe.setTimeout(1000);
```

If set, a per-request value will get priority over a globally set value.

I had to make a bunch of very annoying small changes to tests, as our request object now includes a `settings` object. This is because the previously per-request values were always something that we'd include in the header object, which seemed like the wrong place for `timeout` and `maxNetworkRetries`.

# General cleanup
Previously you'd set per-request settings using snake case:

```js
const charge = await stripe.charges.create({...}, {
  idempotency_key: 'key',
});
```

This was odd since the rest of the library prefers snake case. Plus for header values we prefer the format `Idempotency-Key` when setting them for a request.

For new per-request values, I've only added camel case keys. For existing values, I've added camel case keys but left the old snake case ones for backwards compatibility. The idea being that we'd remove the snake case values entirely in a future major version.

```js
const charge = await stripe.charges.create({...}, {
  idempotencyKey: 'key', // this now works
  stripe_version: '2019-09-09', // this will still work even though "stripeVersion" exists
});
```

r? @rattrayalex-stripe @jlomas-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -219,9 +219,9 @@ StripeResource.prototype = {
   },
 
   // For more on when and how to retry API requests, see https://stripe.com/docs/error-handling#safely-retrying-requests-with-idempotency
-  _shouldRetry(res, numRetries) {
+  _shouldRetry(res, numRetries, maxRetries) {
     // Do not retry if we are out of retries.
-    if (numRetries >= this._stripe.getMaxNetworkRetries()) {
+    if (numRetries >= maxRetries) {
       return false;
     }
 
@@ -358,7 +358,7 @@ StripeResource.prototype = {
     }
   },
 
-  _request(method, host, path, data, auth, options, callback) {
+  _request(method, host, path, data, auth, options = {}, callback) {
     let requestData;
 
     const retryRequest = (
@@ -378,7 +378,11 @@ StripeResource.prototype = {
     };
 
     const makeRequest = (apiVersion, headers, numRetries) => {
-      const timeout = this._stripe.getApiField('timeout');
+      // timeout can be set on a per-request basis. Favor that over the global setting
+      const timeout =
+        options.settings && options.settings.timeout
+          ? options.settings.timeout
+          : this._stripe.getApiField('timeout');
       const isInsecureConnection =
         this._stripe.getApiField('protocol') == 'http';
       let agent = this._stripe.getApiField('agent');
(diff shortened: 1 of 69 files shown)


lib/utils.js
```diff
@@ -135,17 +145,29 @@ const utils = (module.exports = {
           );
         }
 
-        if (params.api_key) {
-          opts.auth = params.api_key;
+        /**
+         * TODO: Remove snake case params in next major version
+         */
+        if (params.api_key || params.apiKey) {
+          opts.auth = params.apiKey || params.api_key;
```
jlomas-stripe: Why are the assignments in the opposite order from the `if()`?

paulasjes-stripe: Oops my bad, fixed now.

test/StripeResource.spec.js
```diff
@@ -141,8 +141,8 @@ describe('StripeResource', () => {
     };
 
     afterEach(() => {
-      realStripe.setMaxNetworkRetries(0);
-      stripe.setMaxNetworkRetries(0);
+      realStripe._setApiNumberField('maxNetworkRetries', 0);
+      stripe._setApiNumberField('maxNetworkRetries', 0);
```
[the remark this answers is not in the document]
paulasjes-stripe: It's a public method now, but will be removed in the next major version (hence why `setMaxNetworkRetries` has a `deprecated` comment above it now).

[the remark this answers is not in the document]
rattrayalex-stripe: ah right. This particular method feels a little funny to use here – feels like an implementation detail that would be nicer to not leak to tests – but no strong feels

paulasjes-stripe:
Addressed your comments in a way that I hope makes sense, ptal @rattrayalex-stripe!

lib/StripeResource.js
```diff
@@ -378,7 +392,12 @@ StripeResource.prototype = {
     };
 
     const makeRequest = (apiVersion, headers, numRetries) => {
-      const timeout = this._stripe.getApiField('timeout');
+      // timeout can be set on a per-request basis. Favor that over the global setting
+      const timeout =
+        options.settings && options.settings.timeout
```
rattrayalex-stripe: I think we want `options.settings && Number.isInteger(options.settings.timeout)` so people can set it to zero, no?

🤔 not actually sure a timeout of zero makes any sense though... Users might want it for some kind of test situation?

jlomas-stripe: == It might make sense for this to be a positive integer, and maybe > 0?

paulasjes-stripe: I'm going to add 
```javascript
Number.isInteger(options.settings.timeout) && options.settings.timeout >= 0
```
As for test purposes a timeout of 0 could make sense.

PR #731, opened 2019-11-20
Update test to expect new config

richardm-stripe:
This test has been failing since https://github.com/stripe/stripe-node/commit/265827e3b2a64b781f8cccd967f19eed89751e4a -- somehow it didn't get updated like the others.


richardm-stripe:
Self-approving as this is trivial and test-only.

PR #738, opened 2019-12-06
Temporary user-agent change to track typescript usage

remi-stripe:
Changing the user-agent temporarily to append a  custom tag for the beta so that we can better track who is using it and reach out for feedback or once the production version is ready.

My main worry here is that changing the user agent format could break our parsing scripts though I hope not. cc @ob-stripe 

r? @rattrayalex-stripe 
cc @stripe/api-libraries


rattrayalex-stripe:
I'd rather find a way to include TS usage in the user agent in general

PR #740, opened 2019-12-09
Adding parameter table for customer.create

AhmadIbrahiim:
Create a collapsible section for `strip.customer.create` parameters, As I was looking for a way to send the customer name while creating the user but I could not find it directly. So I think adding it would be nice

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 # Stripe Node.js Library
 
 [![Version](https://img.shields.io/npm/v/stripe.svg)](https://www.npmjs.org/package/stripe)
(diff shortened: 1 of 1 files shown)


remi-stripe:
@AhmadIbrahiim Thanks for opening a PR! We try to keep the readme for information specific to the client library itself and not the API. All the parameters for our API are explicitly documented in our API reference with basic examples in all the languages we support. You can find all the documentation here: https://stripe.com/docs/api/customers/create

We won't be merging this as it would quickly get out of date as our API evolves and new properties or parameters are added but the API reference would always be up-to-date.

PR #748, opened 2020-01-05
Update SubscriptionSchedules.Phase.Plan.plan to use correct type

petja:
`SubscriptionSchedules.Phase.Plan.plan` was incorrectly using type `SubscriptionSchedules.Phase.Plan` so I changed it to use type `Stripe.Plan` as described [here](https://stripe.com/docs/api/subscription_schedules/object#subscription_schedule_object-phases-plans-plan)


rattrayalex-stripe:
Thanks for the contribution, @petja ! That looks like the right fix. 

cc @richardm-stripe @stripe/api-libraries

PR #749, opened 2020-01-06
Added typescript webhook example

paulasjes-stripe:
r? @rattrayalex-stripe

### DIFF
--- a/examples/webhook-signing/express-ts.ts
+++ b/examples/webhook-signing/express-ts.ts
@@ -1,0 +1,89 @@
+import Stripe from 'stripe';
+import express from 'express';
+import bodyParser from 'body-parser';
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+const stripe: Stripe = new Stripe(process.env.STRIPE_API_KEY!, {
+  apiVersion: '2019-12-03',
+  typescript: true,
+});
+
+/**
+ * You'll need to make sure this is externally accessible. ngrok (https://ngrok.com/)
+ * makes this really easy.
+ *
+ * Alternatively, you could use the stripe-cli in forward mode: https://github.com/stripe/stripe-cli
+ *
+ * To run this file, just provide your Secret API Key and Webhook Secret in a .env file in this directory like so:
+ *
+ * STRIPE_API_KEY=sk_test_XXX
+ * WEBHOOK_SECRET=whsec_XXX
+ *
+ * Then run "npm run tsc", which will convert this TypeScript file to JS and then run it.
+ *
+ * For use with the stripe-cli, run the following:
+ *
+ * 1. "stripe listen --forward-to localhost:3000/webhooks"
+ * 2. Copy the provided webhook signing secret to your .env file
+ * 3. In a new terminal window: "npm run tsc"
+ * 4. In yet another new terminal window: "stripe trigger payment_intents.succeeded"
+ */
+
+const webhookSecret: string = process.env.WEBHOOK_SECRET!;
+const app: express.Application = express();
+
+// Only use the raw body parser for webhooks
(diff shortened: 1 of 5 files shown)


examples/webhook-signing/express-ts.ts
```diff
@@ -0,0 +1,89 @@
...
+const app: express.Application = express();
+
+// Only use the raw body parser for webhooks
+app.use(
+  (
+    req: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ): void => {
+    if (req.originalUrl === '/webhooks') {
```
rattrayalex-stripe: Is this our best-practice? Does line 55 not work here?

examples/webhook-signing/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
+    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
+    "types": ["../../types/2019-12-03"],
+
+    /* Strict Type-Checking Options */
+    "strict": true /* Enable all strict type-checking options. */,
+    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
```
rattrayalex-stripe: I don't think this is needed any longer

examples/webhook-signing/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
+    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
+    "types": ["../../types/2019-12-03"],
```
rattrayalex-stripe: This shouldn't be needed once it lands, but I guess we can keep it in till then?

examples/webhook-signing/express-ts.ts
paulasjes-stripe: Line 55 does work, but I've seen people get tripped up by how express' middleware works. Having this block here will hopefully nip that confusion in the bud (and makes it so all other routes use the json body parser).

examples/webhook-signing/tsconfig.json
paulasjes-stripe: Yeah I wasn't sure if this is necessary or not, happy to remove once v8 lands.

examples/webhook-signing/tsconfig.json
paulasjes-stripe: It's needed for line 5 of `express-ts.ts` where we import `dotenv`.

examples/webhook-signing/express-ts.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+);
+
+// Stripe requires the raw body to construct the event
+app.post(
+  '/webhooks',
+  bodyParser.raw({type: 'application/json'}),
+  (req: express.Request, res: express.Response): express.Response | void => {
+    const sig: string = req.headers['stripe-signature'] as string;
```
rattrayalex-stripe: this is fixed; you no longer need to cast!

PR #752, opened 2020-01-08
Deprecate many library api's, unify others

rattrayalex-stripe:
r? @richardm-stripe 
cc @ob-stripe @jlomas-stripe @paulasjes-stripe 

Much of the tricky stuff here is wrestling with questions of what should be private, what should be private-but-you-can-poke-it-anyway, etc. It might be healthier for my sanity to kick these cans down the road a bit more, but probably healthier for the library to deal with them sooner rather than later, especially if we want to upgrade our internals to TS etc in the medium term.

---
# Breaking changes

## [BREAKING] Remove deprecated Error base export, extend method, and populate method

You can no longer do this:

    const MyStripeError = StripeError.extend({});

Instead, you can do this:

    class MyStripeError extends StripeError {}

You can no longer do this (which should not have had a legitimate use-case):

    const StripeGenericError = require('stripe/lib/Error');

Instead, use the named export `StripeError`:

    const {StripeError} = require('stripe/lib/Error');
    // or
    const Stripe = require('stripe');
    const {StripeError} = Stripe.errors;

## [BREAKING] Privatize several constants

The following constants are no longer accessible:

    Stripe.DEFAULT_HOST
    Stripe.DEFAULT_PORT
    Stripe.DEFAULT_BASE_PATH
    Stripe.DEFAULT_API_VERSION
    Stripe.DEFAULT_TIMEOUT
    Stripe.MAX_NETWORK_RETRY_DELAY_SEC
    Stripe.INITIAL_NETWORK_RETRY_DELAY_SEC

Instead, you can read them like so:

    Stripe.getConstant('DEFAULT_HOST');

Though we may remove this in the future, so please file an issue with your use case if you rely on this functionality.

To set a value you like this, use a config option when instantiating the Stripe client, e.g., 

    const stripe = new Stripe(apiKey, {
      host,
      port,
      basePath,
      timeout,
      apiVersion,
    });

# Non-breaking changes

## Deprecate snake_case options like stripe_account with warning messages 

Doing, e.g., `stripe.customers.create(params, {api_key: foo})` will now emit a warning message encouraging you to use `apiKey` instead.

## Add protocol to config object for local proxies etc

You can now instantiate a stripe client with a non-https protocol, e.g., 

    const stripe = new Stripe(apiKey, {
      host: '127.0.0.1',
      port: 3000,
      protocol: 'http',
    });

This is intended for tests and proxies. All production traffic must be https only.

This replaces `stripe.setProtocol()` and `stripe.setHost(host, port, protocol)`, which are deprecated.

## Add appInfo to config object

Stripe plugin authors can now instantiate a stripe client with `appInfo` instead of using `stripe.setAppInfo`, eg;

    const stripe = new Stripe(apiKey, {
      appInfo: {
        name: 'MyCoolPlugin',
      }
    });

## Mark all setter methods as deprecated, emit warnings.

Setter methods like `stripe.setApiVersion()` were previously deprecated; now, they emit deprecation warnings. Please move to the config object or request options instead. 

For example, instead of this:

    const stripe = new Stripe(apiKey);
    stripe.setApiVersion('2019-11-03');
    

You could do this:

    const stripe = new Stripe(apiKey, {apiVersion: '2019-11-03'});
    stripe.customers.list();

Or this:

    const stripe = new Stripe(apiKey);
    stripe.customers.list({apiVersion: '2019-11-03'});

## Mark methods prefixed with an underscore as private

Methods like `_setAppInfo` which are intended for private use now have a docstring indicating they are private, and may be removed in the future. 

If you rely on any underscore-prefixed methods, please open an issue on github detailing your use-case. 

## Mark a few getter methods which are not prefixed with underscores as private

The following getter methods were likely always intended for private use only, but may be used for reasons we haven’t yet thought of. We may remove them in the future. 

    stripe.getApiField()
    stripe.getConstant()
    stripe.getClientUserAgent()
    stripe.getClientUserAgentSeeded()
    stripe.getAppInfoAsString()

If you are relying on any of these, please let us know.

## Rename the request option stripeVersion to apiVersion

We previously used `stripeVersion` and did not have `apiVersion` here, which was inconsistent with how you instantiate the stripe client: `new Stripe(apiKey, {apiVersion: '2019-12-03'})`. 

Instead of this:

    stripe.customers.list({stripeVersion: '2019-11-03'});

Please do this instead:

    stripe.customers.list({apiVersion: '2019-11-03'});

The former will now log a warning message, but will continue to work as expected.

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -293,6 +293,13 @@ Stripe.prototype = {
     this._api[key] = value;
   },
 
+  /**
+   * @private
+   * Please open or upvote an issue at github.com/stripe/stripe-node
+   * if you use this, detailing your use-case.
+   *
+   * It may be deprecated and removed in the future.
+   */
   getApiField(key) {
     return this._api[key];
   },
@@ -305,6 +312,13 @@ Stripe.prototype = {
     return this._clientId;
   },
 
+  /**
+   * @private
+   * Please open or upvote an issue at github.com/stripe/stripe-node
+   * if you use this, detailing your use-case.
+   *
+   * It may be deprecated and removed in the future.
+   */
   getConstant: (c) => {
     switch (c) {
       case 'DEFAULT_HOST':
@@ -359,8 +373,16 @@ Stripe.prototype = {
     return INITIAL_NETWORK_RETRY_DELAY_SEC;
   },
 
-  // Gets a JSON version of a User-Agent and uses a cached version for a slight
-  // speed advantage.
+  /**
+   * @private
+   * Please open or upvote an issue at github.com/stripe/stripe-node
+   * if you use this, detailing your use-case.
+   *
+   * It may be deprecated and removed in the future.
+   *
+   * Gets a JSON version of a User-Agent and uses a cached version for a slight
+   * speed advantage.
+   */
   getClientUserAgent(cb) {
     if (Stripe.USER_AGENT_SERIALIZED) {
       return cb(Stripe.USER_AGENT_SERIALIZED);
@@ -371,8 +393,16 @@ Stripe.prototype = {
     });
   },
 
-  // Gets a JSON version of a User-Agent by encoding a seeded object and
-  // fetching a uname from the system.
+  /**
+   * @private
+   * Please open or upvote an issue at github.com/stripe/stripe-node
+   * if you use this, detailing your use-case.
+   *
+   * It may be deprecated and removed in the future.
+   *
+   * Gets a JSON version of a User-Agent by encoding a seeded object and
+   * fetching a uname from the system.
+   */
   getClientUserAgentSeeded(seed, cb) {
     utils.safeExec('uname -a', (err, uname) => {
       const userAgent = {};
@@ -391,6 +421,13 @@ Stripe.prototype = {
     });
   },
 
+  /**
+   * @private
+   * Please open or upvote an issue at github.com/stripe/stripe-node
+   * if you use this, detailing your use-case.
+   *
+   * It may be deprecated and removed in the future.
+   */
   getAppInfoAsString() {
     if (!this._appInfo) {
       return '';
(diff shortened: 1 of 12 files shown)


lib/stripe.js
```diff
@@ -2,13 +2,13 @@
 
 const resources = require('./resources');
 
-Stripe.DEFAULT_HOST = 'api.stripe.com';
-Stripe.DEFAULT_PORT = '443';
-Stripe.DEFAULT_BASE_PATH = '/v1/';
-Stripe.DEFAULT_API_VERSION = null;
```
rattrayalex-stripe: These constants I'm moderately worried about removing. Users could previously read/write them, though I don't think that was ever intended. 

They can still read them with `Stripe.getConstant` and all of them have proper ways of writing, eg `new Stripe(key, {host, port, basePath, apiVersion, timeout})`

rattrayalex-stripe: cc @ob-stripe @remi-stripe

lib/utils.js
```diff
@@ -140,28 +145,35 @@ const utils = (module.exports = {
...
+          const nonDeprecated = extraKeys.filter((key) => {
+            if (!DEPRECATED_OPTIONS[key]) {
+              return true;
+            }
+            const newParam = DEPRECATED_OPTIONS[key];
+            /**
+             * TODO turn this into a hard error in a future major version (once we have fixed our docs).
+             */
+            emitWarning(`'${key}' is deprecated; use '${newParam}' instead.`);
+            params[newParam] = params[key];
```
richardm-stripe: Should this hard error if params[key] is already set?

rattrayalex-stripe: Interesting... yeah I think that's sane / unlikely to cause much user frustration / generally more likely to be helpful than harmful.

rattrayalex-stripe: (Done)

PR #764, opened 2020-01-14
Added typescript webhook example

thorsten-stripe:
Closes #749 

r? @rattrayalex-stripe 

In Paul's absence, I'm looking to finish up the TypeScript webhook sample.


examples/webhook-signing/README.md
```diff
@@ -0,0 +1,48 @@
+# Checking webhook signatures
+
+Verify the events that Stripe sends to your webhook endpoints. Additional docs: https://stripe.com/docs/webhooks/signatures.
+
+Additional Node TypeScript server implementation sample: https://github.com/stripe-samples/accept-a-card-payment/tree/master/using-webhooks/server/node-typescript.
```
rattrayalex-stripe: Should we make this a markdown link rather than showing the full url?

examples/webhook-signing/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Requirements
+
+You’ll need the following:
+
+- [Node.js](http://nodejs.org) >=10.0.0
+- Stripe account to accept payments ([sign up](https://dashboard.stripe.com/register) for free).
+- [Stripe CLI](https://github.com/stripe/stripe-cli) or [ngrok](https://ngrok.com/) to tunnel requests to your local server.
```
rattrayalex-stripe: Love having reqs like this

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
...
+      // On error, return the error message
+      res.status(400).send(`Webhook Error: ${err.message}`);
+      return;
+    }
+
+    // Do something with event
+    console.log('Success:', event.id);
+
+    // Cast event data to Stripe object
+    let stripeObject: Stripe.PaymentIntent | Stripe.Charge;
```
rattrayalex-stripe: this we should not do... I think better to just do this:

```ts
    switch (event.type) {
      case 'payment_intent.succeeded':
        const pi = event.data.object as Stripe.PaymentIntent;
        console.log(`PaymentIntent status: ${pi.status}`);
        break;
      case 'charge.succeeded':
        const charge = event.data.object as Stripe.Charge;
        console.log(`Charge id: ${charge.id}`);
        break;
    }
```

Reasoning: 
- sharing the variable could make it easier for different switch cases to affect each other or to do weird things after the switch
- having a union (`|`) implies that _anywhere_ you use `stripeObject`, it could be either a PI or a Charge, whereas  reality is a little different. Also, users could forget to update the union even when handling more event types, which could be misleading

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
...
+const stripe: Stripe = new Stripe(process.env.STRIPE_SECRET_KEY, {
+  apiVersion: '2019-12-03',
+  typescript: true,
+});
+
+const webhookSecret: string = process.env.STRIPE_WEBHOOK_SECRET;
+
+const app: express.Application = express();
+
+// Only use the raw body parser for webhooks.
```
rattrayalex-stripe: I think we should remove this block, it implies that you need it, but I think you don't – the `bodyParser.raw` call is enough (worth double-checking, I'm not 100% sure
)

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
+import Stripe from 'stripe';
+import express from 'express';
+import bodyParser from 'body-parser';
+import env from 'dotenv';
+
+env.config();
+
+const stripe: Stripe = new Stripe(process.env.STRIPE_SECRET_KEY, {
```
rattrayalex-stripe: I think this cast is not idiomatic or necessary

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+);
+
+// Stripe requires the raw body to construct the event
+app.post(
+  '/webhook',
+  bodyParser.raw({type: 'application/json'}),
+  (req: express.Request, res: express.Response): void => {
+    const sig: string | string[] = req.headers['stripe-signature'];
```
rattrayalex-stripe: This cast is not necessary (and is kinda confusing)

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    let event: Stripe.Event;
+
+    try {
+      event = stripe.webhooks.constructEvent(req.body, sig, webhookSecret);
+    } catch (err) {
+      // On error, return the error message
```
rattrayalex-stripe: I wonder if we should also demonstrate logging a warning to the effect of:
```
console.warn('Failed to construct a webhook from Stripe. Could there be a malicious third party sending webhook payloads? Error message: ' + err.message)
```

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
...
+      res.status(400).send(`Webhook Error: ${err.message}`);
+      return;
+    }
+
+    // Do something with event
```
rattrayalex-stripe: comment is a little weird now that we're actually doing something with the event down below, but I do still like logging the event id

examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    switch (event.type) {
+      case 'payment_intent.succeeded':
+        stripeObject = event.data.object as Stripe.PaymentIntent;
+        console.log(`PaymentIntent status: ${stripeObject.status}`);
+        break;
+      case 'charge.succeeded':
+        stripeObject = event.data.object as Stripe.Charge;
+        console.log(`Charge id: ${stripeObject.id}`);
+        break;
```
rattrayalex-stripe: should we add a `default` case so that if people run this with some other event type, they see something here?

examples/webhook-signing/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Copy the environment variables file from the root of the repository:
+
+    cp .env.example .env
+
+Update `.env` with your own [Stripe API keys](https://dashboard.stripe.com/account/apikeys).
+
+### Install and run
+
+Install dependencies using npm:
```
rattrayalex-stripe:
```suggestion
Install dependencies:
```

(don't want to reply that `yarn` won't work for some reason)

rattrayalex-stripe: We should probably specify what directory all of these commands should be run in

examples/webhook-signing/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+To start the webhook forwarding run:
+
+    stripe listen --forward-to localhost:3000/webhook
+
+The Stripe CLI will let you know that webhook forwarding is ready and output your webhook signing secret:
+
+    > Ready! Your webhook signing secret is whsec_xxx
+
+Copy the webhook signing secret (`whsec_xxx`) to your `.env` file.
```
rattrayalex-stripe: nice, great job documenting CLI<>example interaction!

examples/webhook-signing/typescript-node-express/express-ts.ts
thorsten-stripe: Fair, problem is if I move the const declaration into the switch case the linter complains with `Unexpected lexical declaration in case block. eslint(no-case-declarations)`. I'll refactor to if else.

rattrayalex-stripe: Nice, yeah, +1 to if/else.

examples/webhook-signing/typescript-node-express/express-ts.ts
thorsten-stripe: I think this was specifically to showcase that for all other routes we want to use the json parser. See Paul's original comment on that: https://github.com/stripe/stripe-node/pull/749#discussion_r363573568

examples/webhook-signing/typescript-node-express/express-ts.ts
thorsten-stripe: It's more likely that they are passing a wrongly parsed body or whsec. Don't think we want to "fear monger" here.

examples/webhook-signing/README.md
thorsten-stripe: added to first setup sentence

examples/webhook-signing/typescript-node-express/express-ts.ts
rattrayalex-stripe: Yeah... if we did this:

```js
app.use(bodyParser.json());

app.post(
  '/webhook',
  bodyParser.raw({type: 'application/json'}),
  (req: express.Request, res: express.Response): void => {
```
would the app still work?

cc @adreyfus-stripe maybe you know?

examples/webhook-signing/typescript-node-express/express-ts.ts
rattrayalex-stripe: fair enough

PR #766, opened 2020-01-15
Update 'custom_fields' type to match documentation

qasim:
Updating the `custom_fields` type in Invoice creation/update to match the documentation, which specifies an array of hashes.

Not fully sure if these type definitions are auto-generated somewhere else, though noticed this inconsistency with what the endpoint expects. ☺️

### DIFF
--- a/types/2019-12-03/Invoices.d.ts
+++ b/types/2019-12-03/Invoices.d.ts
@@ -499,7 +499,7 @@ declare module 'stripe' {
       /**
        * A list of up to 4 custom fields to be displayed on the invoice.
        */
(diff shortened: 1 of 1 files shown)


remi-stripe:
@qasim Thanks a lot for the PR! I think the type is incorrect somewhere, it likely should be an array of `CustomField` instead (where we'd remove the S since it goes in an array).

r? @richardm-stripe to investigate whether this is the right fix and likely fix the related bug in codegen.

remi-stripe:
Closing as we fixed it in the other PR. Thanks again @qasim for the work and raising the issue

PR #767, opened 2020-01-15
Adding missing events for pending updates on `Subscription`

remi-stripe:
Codegen for openapi c2a8b90

r? @richardm-stripe 
cc @stripe/api-libraries


types/2019-12-03/InvoiceItems.d.ts
```diff
@@ -159,6 +154,11 @@ declare module 'stripe' {
        */
       amount?: number;
 
+      /**
+       * Three-letter [ISO currency code](https://www.iso.org/iso-4217-currency-codes.html), in lowercase. Must be a [supported currency](https://stripe.com/docs/currencies).
+       */
+      currency?: string;
```
richardm-stripe: this didn't have an analog in the stripe-java PR? Is it related to the missing events?

remi-stripe: Nah, it's the ordering of the field that seems to have changed for reasons beyond me, but it doesn't look like anything bad right?

richardm-stripe: It also is changing to optional

PR #775, opened 2020-01-22
Gracefully handle a missing `subprocess` module

### LINKED ISSUE #773: Feature Request: Support for Cloudflare Workers
I've been trying to integrate this library in my one of my Cloudflare Workers and I keep running into an issue with `child_process`. It seems this is because Workers run on a V8 isolate rather than process-level containers. Would love to be able to use Stripe when building Worker functions.

brandur:
As reported in #773, it seems that certain sandboxed environments like
CloudFlare Workers don't provide a `subprocess` module, and that causes
Node to throw when we try to require it.

Because the only thing we need `subprocess` for is getting `uname`, here
we extend the already-existing infrastructure built to gracefully handle
errors to also be able to handle this new error case.

Fixes #773.

r? @rattrayalex-stripe (Or recommend another reviewer if you like.)
cc @stripe/api-libraries

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -9,7 +9,7 @@ const hasOwn = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop);
 // Certain sandboxed environments (our known example right now are CloudFlare
 // Workers) may make `child_process` unavailable. Because `exec` isn't critical
 // to the operation of stripe-node, we handle this unavailability gracefully.
-var exec = null;
+let exec = null;
 try {
   exec = require('child_process').exec;
 } catch (e) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,11 +1,22 @@
 'use strict';
 
 const EventEmitter = require('events').EventEmitter;
-const exec = require('child_process').exec;
 const qs = require('qs');
 const crypto = require('crypto');
 
 const hasOwn = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop);
+
+// Certain sandboxed environments (our known example right now are CloudFlare
+// Workers) may make `child_process` unavailable. Because `exec` isn't critical
+// to the operation of stripe-node, we handle this unavailability gracefully.
+var exec = null;
+try {
+  exec = require('child_process').exec;
+} catch (e) {
+  if (e.code !== 'MODULE_NOT_FOUND') {
+    throw e;
+  }
+}
 
 const OPTIONS_KEYS = [
   'apiKey',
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Had a brief conversation about this with @ob-stripe just now, and he mentioned that the alternative to this fix might be to suggest people stub out `subprocess` themselves when it's not available. I don't have a super strong opinion on it, so I'll leave to DX Products to decide.

test/utils.spec.js
```diff
@@ -506,6 +506,22 @@ describe('utils', () => {
...
+      var actualErr = null;
+      var actualRes = null;
+      function myCb(err, res) {
+        actualErr = err;
+        actualRes = res;
+      }
+      utils.safeExec('hello', myCb);
+      expect(actualErr.toString()).to.equal(
+        new Error('exec not available').toString()
+      );
```
brandur-stripe: I'd love if someone who is familiar with Node convention could check whether they like this. Basically I found that errors are not treated as equal by Chai even if they have the same error message, so fell back to this `toString()` solution, but there might be a better way.

lib/utils.js
```diff
@@ -1,12 +1,23 @@
...
-const exec = require('child_process').exec;
 const qs = require('qs');
 const crypto = require('crypto');
 
 const hasOwn = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop);
 
+// Certain sandboxed environments (our known example right now are CloudFlare
+// Workers) may make `child_process` unavailable. Because `exec` isn't critical
+// to the operation of stripe-node, we handle this unavailability gracefully.
+var exec = null;
```
rattrayalex-stripe: minor style nit
```suggestion
let exec = null;
```

test/utils.spec.js
rattrayalex-stripe: This works. We should upgrade our testing library to Jest.

lib/utils.js
brandur-stripe: Oops, even I should know about that one. Thanks!

brandur-stripe:
Thank you Alex for going above and beyond the line of duty!

ptal @ob-stripe Going to hand off to you for a final decision on whether this is the right way to go for the library.

ob-stripe:
This PR makes it so `safeExec` doesn't crash, but it still returns an error, right? Wouldn't this prevent requests from being sent because `uname -a` will still fail to execute?

ptal @brandur-stripe

brandur-stripe:
Good question OB. If you look carefully at how `getClientUserAgentSeeded` is implemented, it actually doesn't care if `safeExec` errored out; the only effect is to replace the uname component with "UNKNOWN" because a real `uname` wasn't available:

``` javascript
    utils.safeExec('uname -a', (err, uname) => {
      const userAgent = {};
      for (const field in seed) {
        userAgent[field] = encodeURIComponent(seed[field]);
      }

      // URI-encode in case there are unusual characters in the system's uname.
      userAgent.uname = encodeURIComponent(uname || 'UNKNOWN');

      if (this._appInfo) {
        userAgent.application = this._appInfo;
      }

      cb(JSON.stringify(userAgent));
    });
  },
```

Not sure if that's the best design, but it'll work at least.

ptal @ob-stripe

brandur-stripe:
Thanks Olivier!

PR #776, opened 2020-01-23
Add support for new `type` on `CustomerTaxId`

remi-stripe:
Codegen for openapi 5013467

r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/resources/Accounts.js
+++ b/lib/resources/Accounts.js
@@ -7,6 +7,11 @@ const stripeMethod = StripeResource.method;
 module.exports = StripeResource.extend({
   path: '',
 
+  reject: stripeMethod({
+    method: 'POST',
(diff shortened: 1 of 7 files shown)


lib/resources/Accounts.js
```diff
@@ -7,6 +7,11 @@ const stripeMethod = StripeResource.method;
 module.exports = StripeResource.extend({
   path: '',
 
+  reject: stripeMethod({
```
remi-stripe: Flagging that something changed to change the order on some methods. It's not breaking so I'm leaving it but we need to fix codegen to avoid that

remi-stripe: cc @richardm-stripe as we're merging but this is something that needs to be fixed soon in codegen

PR #785, opened 2020-01-30
Fix file uploads with nested params

jlomas-stripe:
Previously we were getting our keys from a flattened object, but then using the flattened key to access the original object, so 💥.

Now with 100% less 💥.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
Released as 8.8.2.

PR #795, opened 2020-02-05
Update Documentation which is wrong for connected accounts

aloysius-tim:


remi-stripe:
@aloysius-tim If you use the latest version of stripe-node (8.*) then `stripeAccount` works and is the correct way to make the call. If you use `stripe_account`, you get a warning:

> 'stripe_account' is deprecated; use 'stripeAccount' instead.

Is it possible you're on an older version of the library where `stripeAccount` was not yet supported?

aloysius-tim:
Indeed it must be a version difference, my version should be one of the latest ?: "stripe": "^8".

But still, in the main documentation, they speak of stripe_account and not stripeAccount: https://stripe.com/docs/connect/authentication








—

TIMOTHY KEYNES
ENTREPRENEUR & 42 🚀
www.keynes.fr

P: +33 06 73 80 96 32
E: tim@keynes.fr

remi-stripe:
Ah thanks for the flag. That doc is incorrect and the more recent one says `stripeAccount`. I'll get the one you found fixed! Going to close the PR as the readme is correct in the end, I hope it's okay!

PR #807, opened 2020-02-15
Change timeout default to constant 80000 instead Node default

### LINKED ISSUE #806: Default http timeout changed in Node 13
In Node.js v13.0.0 [the default socket timeout has been set to 0](https://github.com/nodejs/node/pull/27558).

Currently in this Library the default timeout is being set based on [`require('http').createServer().timeout`](https://github.com/stripe/stripe-node/blob/1a6cd581514de22d1561ae68c2401a168485c07d/lib/stripe.js#L11).
https://github.com/stripe/stripe-node/blob/1a6cd581514de22d1561ae68c2401a168485c07d/lib/stripe.js#L11
 The default previously was 120 seconds, under v13 it is no timeout by default aka a value of `0`.

I noticed this when looking into https://github.com/stripe/stripe-node/issues/796

I haven't done any testing with this so I don't know if it represents an actual bug, but wanted to file this issue in case this was something the team was unaware of. 

This timeout was introduced in Node.js v0.9.12 as 120 seconds, and did not change until v13.0.0 when it was switched to 0. There is of course a possibility that this change will be reverted before v14 is released, but it's worth keeping an eye on.

jonchurch:
Closes #806 

Set `timeout` default to a constant 80 seconds instead of using the default value of the version of Node.js the user is running.

Updated tests, docs, comment in types def file.

**Note**: Not sure how you define breaking changes here, but this would be a departure from the previous default behavior of 120 seconds. This issue is currently out in the wild though (for anyone _\*brave\*_ enough to be using v13 in production) so I would suggest a release as soon as feasible. 

Switching this back to a constant 120 seconds and a minor version bump would prevent anyone from encountering the issue in the current major.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,15 +165,15 @@ const stripe = Stripe('sk_test_...', {
 });
 ```
 
-| Option              | Default                       | Description                                                                           |
-| ------------------- | ----------------------------- | ------------------------------------------------------------------------------------- |
-| `apiVersion`        | `null`                        | Stripe API version to be used. If not set the account's default version will be used. |
-| `maxNetworkRetries` | 0                             | The amount of times a request should be [retried](#network-retries).                  |
-| `httpAgent`         | `null`                        | [Proxy](#configuring-a-proxy) agent to be used by the library.                        |
-| `timeout`           | 120000 (Node default timeout) | [Maximum time each request can take in ms.](#configuring-timeout)                     |
-| `host`              | `'api.stripe.com'`            | Host that requests are made to.                                                       |
-| `port`              | 443                           | Port that requests are made to.                                                       |
-| `telemetry`         | `true`                        | Allow Stripe to send latency [telemetry](#request-latency-telemetry).                 |
+| Option              | Default            | Description                                                                           |
+| ------------------- | ------------------ | ------------------------------------------------------------------------------------- |
+| `apiVersion`        | `null`             | Stripe API version to be used. If not set the account's default version will be used. |
+| `maxNetworkRetries` | 0                  | The amount of times a request should be [retried](#network-retries).                  |
+| `httpAgent`         | `null`             | [Proxy](#configuring-a-proxy) agent to be used by the library.                        |
+| `timeout`           | 80000              | [Maximum time each request can take in ms.](#configuring-timeout)                     |
+| `host`              | `'api.stripe.com'` | Host that requests are made to.                                                       |
+| `port`              | 443                | Port that requests are made to.                                                       |
+| `telemetry`         | `true`             | Allow Stripe to send latency [telemetry](#request-latency-telemetry).                 |
 
 Note: Both `maxNetworkRetries` and `timeout` can be overridden on a per-request basis.
 
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -7,8 +7,7 @@ const DEFAULT_PORT = '443';
 const DEFAULT_BASE_PATH = '/v1/';
 const DEFAULT_API_VERSION = null;
 
-// Use node's default timeout:
-const DEFAULT_TIMEOUT = require('http').createServer().timeout;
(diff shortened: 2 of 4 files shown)


brandur-stripe:
Thanks for the thorough patch @jonchurch!

Going to take the risk here and release a minor change — our server side timeouts will almost certainly cut you off well well before you hit even 80 seconds, so I don't expect any user-discernible difference afterwards.

Just reading through Node docs, I also realize that we should also be setting a 30 second _connect_ timeout when generating a request with `https.request` to be consistent with other libraries, but we can tackle that one another day.

jonchurch:
Happy to help, and thank you for being so responsive ❤️.

Looking into this more (never realized there was this much to think about timeouts!) the current state of things for the library is that we are setting the timeout for idle TCP connections. Meaning the timeout will only be triggered after there is no activity over the connection for the duration of the timeout. 

I think that setting the timeout when creating the request will share that timeout for all stages of the TCP socket's lifecycle: initial attempt at connecting, DNS lookup, after connection, between read/writes over the socket. The timeout is reset at each stage, and at each read/write event. [Here's a decent blog post I found about this.](https://shuheikagawa.com/blog/2017/05/13/http-request-timeouts-in-javascript/).

In the library the change would happen here, when the request is first created:

https://github.com/stripe/stripe-node/blob/1d6207e34f978d8709d42d8a05d7d7e8be6599c7/lib/StripeResource.js#L417-L425

None of this really feels super high value IMO since the server will take care of timing out bad connections it looks like, but wanted to leave this comment to share what I learned.

brandur-stripe:
NP!

> the current state of things for the library is that we are setting the timeout for idle TCP connections. Meaning the timeout will only be triggered after there is no activity over the connection for the duration of the timeout.

I think you might have read the docs for [`net.Socket.setTimeout()`](https://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback) instead of for [`request.setTimeout(timeout[, callback])`](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback). My reading of the latter makes it sound that our invocation of `setTimeout` does what we expect it to — namely to time out a request that we generate after after a certain duration.

> In the library the change would happen here, adding a `timeout` to the options, when the request is first created:

Ah yep, that's where I was thinking. That timeout would have an effect on time to open the socket, which is 30 seconds for us in our other libraries. But yep, not it's also probably not a hugely critical change.

jonchurch:
@brandonl-stripe Hmm I think I'm going further down the rabbit hole here than is productive but this has piqued my interest. Feel free to ignore this thread of conversation.

Under the hood `request.setTimeout` calls `socket.setTimeout`. It just waits until the connection event happens before setting the socket timeout:

https://github.com/nodejs/node/blob/7c524fb092b143470795c8ca869de4654c728fe1/lib/_http_client.js#L797-L823

**I realize now my source of misunderstanding.** I was thinking of timeouts in terms of how long it took for the _entire_ response to be received. I think traditionally HTTP timeouts refer to _how long it takes to receive a response_, not how long it takes to receive the entire response. 

The above timeout affects how long a connection will wait to hear back from the server, not how long it takes to receive the entire response body. Since this library talks to the Stripe API and not a potentially bad actor very slowly sending the response, my interpretation was irrelevant. 

After connection, while waiting for the response to be sent from the server, the socket is idle. So if the response doesn't begin flowing before the set interval, we see the timeout.

You can also force a timeout if there is X time between packets sent from the server after the response has begun flowing if you reset the timeout after receiving the response. 

[Here's a gist](https://gist.github.com/jonchurch/85f787e51bf4cf2fd821046a6b9bca0d) where I thought through all this. I picked a really slow API (the [NASA Exoplanet API](https://exoplanetarchive.ipac.caltech.edu/docs/program_interfaces.html)).

brandonl-stripe:
@brandur-stripe ☝️ 😄

Issue #816, opened 2020-02-24
AccountLinkCreateParams.type is not TS-typed


cyrilgandon:
See https://github.com/stripe/stripe-node/blob/master/types/2019-12-03/AccountLinks.d.ts#L47
```
/**
* The type of account link the user is requesting. Possible values are `custom_account_verification` or `custom_account_update`.
*/
type: string;
```
It would be nice to have type on it.

```
type: AccountLinkCreateParams.Type;
```
```
namespace AccountLinkCreateParams {
    type Collect = 'currently_due' | 'eventually_due';
    type Type = 'custom_account_verification' | 'custom_account_update';
}
```

remi-stripe:
Closing the issue as you already opened a PR and we can talk on that one

PR #817, opened 2020-02-24
Add typings on AccountLinkCreateParams.type field

cyrilgandon:
Fixing https://github.com/stripe/stripe-node/issues/816


remi-stripe:
@cyrilgandon Thanks for opening a PR to fix this! 

For a bit of context, our Typescript definitions are generated automatically for the openapi spec representing our API. Today, that field is a `string` which is why it's not typed as an enum. 

We can't merge your PR just yet, because if we do, it will be reverted back to a string next time we release a change. We're working on changing the code internally to be a real enum. Once that's done, we'll be able to merge your change!

remi-stripe:
Flagging that the other PR was merged (automatically generated from our own openapi spec) and was just released as 8.23.0

PR #824, opened 2020-03-02
Update type for StripeError to extend Error

### LINKED ISSUE #823: [Typescript] StripeError missing "extends Error"
Hi all,

The `StripeError` class doesn't extend `Error` in the typings, leading to errors like `Error:(439, 44) TS2345: Argument of type 'StripeError' is not assignable to parameter of type 'Error'.
  Property 'name' is missing in type 'StripeError'.`

Example

```ts
import {Stripe} from 'stripe';

function test(err: Error) {}

const baseStripeError: Stripe.StripeError = null;

test(baseStripeError);
```

I've checked the sources and the class is actually extending `Error`.

Cheers

cjavilla-stripe:
This matches the implemented type

r? @richardm-stripe 

fixes #823

### DIFF
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -203,7 +203,7 @@ declare module 'stripe' {
       | 'authentication_error'
       | 'invalid_grant';
 
-    class StripeError {
+    class StripeError extends Error {
       static populate(type: RawErrorType): StripeError;
 
       /**


cjavilla-stripe:
This file is not auto genned, so I just updated it directly. Not sure if we need to (or how to) cut a new release for this type of thing.

cjavilla-stripe:
cc @stripe/api-libraries

PR #826, opened 2020-03-03
Updates Typescript definitions for the latest API version `2020-03-02` and related changes

remi-stripe:
Multiple API changes:
* Types are now for the API version `2020-03-02`
* Remove `uob_regional` as a value on `bank` for FPX as this is deprecated and was never used
* Add support for `next_invoice_sequence` on `Customer`
* Add support for `proration_behavior` on `SubscriptionItem` delete

r? @ob-stripe 
cc @stripe/api-libraries

PR #843, opened 2020-03-20
Add script to update the api version

rattrayalex-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @remi-stripe 

This automates the work done in https://github.com/stripe/stripe-node/pull/826 with a handy little regex script (and a new `API_VERSION` file, which is codegen'd). 

It also:

1. Removes `Addresses.d.ts`, which is obsolete – the definition for `type Address` lives in `tyeps/shared.d.ts`. I misadvised remi when I told him we need to keep it, I think.
2. Updates the openapi spec, because it's hard not to. 
3. Add (and revert) a demo commit to show what would happen if you ran this for api version 2020-12-12: https://github.com/stripe/stripe-node/commit/059273c2200d880ac81c12f3a617acaf7d6da703


rattrayalex-stripe:
Awesome thanks – I'll merge & release that one first.

rattrayalex-stripe:
I rebased and dropped the openapi update, since remi's PR should have been ahead of that if anything.

PR #848, opened 2020-03-24
Fix return type for autoPagingEach

mhagmajer:
`autoPagingEach` returns a `Promise`


richardm-stripe:
Hi @mhagmajer, thanks for the bug fix!

PR #849, opened 2020-03-24
Add support for `pause_collection` on `Subscription`

remi-stripe:
Codegen for openapi cadc76e

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/types/2020-03-02/Issuing/Disputes.d.ts
+++ b/types/2020-03-02/Issuing/Disputes.d.ts
@@ -18,24 +18,24 @@ declare module 'stripe' {
         /**
          * Disputed amount. Usually the amount of the `disputed_transaction`, but can differ (usually because of currency fluctuation or because only part of the order is disputed).
          */
-        amount: number;
+        amount?: number;
 
         /**
          * Time at which the object was created. Measured in seconds since the Unix epoch.
          */
-        created: number;
+        created?: number;
 
         /**
          * The currency the `disputed_transaction` was made in.
          */
-        currency: string;
+        currency?: string;
 
         /**
          * The transaction being disputed.
          */
-        disputed_transaction: string | Stripe.Issuing.Transaction;
+        disputed_transaction?: string | Stripe.Issuing.Transaction;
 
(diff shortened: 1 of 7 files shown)


types/2020-03-02/Issuing/Disputes.d.ts
```diff
@@ -18,24 +18,24 @@ declare module 'stripe' {
         /**
          * Disputed amount. Usually the amount of the `disputed_transaction`, but can differ (usually because of currency fluctuation or because only part of the order is disputed).
          */
-        amount: number;
+        amount?: number;
```
ob-stripe: Why did all dispute fields except `id` and `object` become optional?

types/2020-03-02/PaymentMethods.d.ts
```diff
@@ -126,7 +126,7 @@ declare module 'stripe' {
         exp_year: number;
 
         /**
-         * Uniquely identifies this particular card number. You can use this attribute to check whether two customers who've signed up with you are using the same card number, for example.
+         * Uniquely identifies this particular card number. You can use this attribute to check whether two customers who've signed up with you are using the same card number,for example. For payment methods that tokenize card information (Apple Pay, Google Pay), the tokenized number might be provided instead of the underlying card number.
```
ob-stripe: That comment change doesn't seem correct.

types/2020-03-02/Issuing/Disputes.d.ts
remi-stripe: Internal reasons mostly

types/2020-03-02/PaymentMethods.d.ts
remi-stripe: Why isn't this correct? It's not greatly worded, I agree, but it just means we calculate the fingerprint based on the DPAN

ob-stripe: We should fix the comment to be clearer, but that's not a blocker for this PR.

Issue #541, opened 2019-01-05
Expose constants


adamreisnz:
Would it possible to expose constants on the relevant API endpoints (or in its own namespace)? For example:

```js
stripe.constants.subscriptionStatuses = {
  TRIALING: 'trialing', 
  ACTIVE: 'active', 
  PAST_DUE: 'past_due', 
  CANCELED: 'canceled',
  UNPAID: 'unpaid',
};
```

This would make defining and maintaining such constants yourself unnecessary.

ob-stripe:
Hi @adamreisnz! This is a reasonable request, however our experience with other libraries has shown that it's fairly difficult to define those constants and keep them up-to-date manually. We do plan to add this in the future, but we want to do it in an automated manner.

I can't provide any ETA yet, but I'll leave this issue open and tag it as `future`. We'll update here as soon as we have something to share.

adamreisnz:
Great thanks. Yes, you can imagine then how difficult it is for developers to keep the constants up-to-date and stay on top of any changes that occur. I thought if it's done in the official Stripe libraries at least it's closer to the source and probably better to have it concentrated in one place.

remi-stripe:
Going to close this one as the library has had Typescript definitions for a long time now which should mostly solve that problem: https://github.com/stripe/stripe-node#usage-with-typescript

PR #858, opened 2020-04-02
Proposal: Add enum values to Javascript resources

richardm-stripe:
This is an approach to solving #541
cc @stripe/api-libraries @ob
Wanted to get feedback before proceeding to codegen.

Usage: 

When making requests, instead of
```js
stripe.paymentIntents.cancel('pi_xxx', {
  cancellation_reason: 'duplicate'
}
```
this permits
```js
stripe.paymentIntents.cancel('pi_xxx', {
  cancellation_reason: s.paymentIntents.constants.cancellation_reason.duplicate
})
```

When handling responses, instead of 
```js
if (pi.cancellation_reason === 'duplicate') {
  // do something
}
```

do

```js
if (pi.cancellation_reason === s.paymentIntents.constants.cancellation_reason.duplicate) {
  // do something
}
```

**Q: Why?**
Some users stylistically prefer to avoid "magic strings". I think this is less important in the world of Typescript, since typescript has string literal types, but not all users use typescript.

In theory this gives you the ability to autocomplete the enum values
image

I don't think we should use the constants in the documentation, but we can provide them.

**Q: How you handle differences between method-level and resource-level enums**
I propose that we *only* define enums at the resource level, and don't namespace anything per-method like we do in the typescript definitions. The enums at the resource level will include all enum entries, defined on the resource or in any of the resource's methods.

### DIFF
--- a/lib/resources/PaymentIntents.js
+++ b/lib/resources/PaymentIntents.js
@@ -22,4 +22,27 @@ module.exports = StripeResource.extend({
     method: 'POST',
     path: '/{intent}/confirm',
   }),
+
+  constants: {
+    cancellation_reason: {
+      abandoned: 'abandoned',
+      automatic: 'automatic',
+      duplicate: 'duplicate',
+      failed_invoice: 'failed_invoice',
+      fraudulent: 'fraudulent',
+      requested_by_customer: 'requested_by_customer',
+      void_invoice: 'void_invoice',
+    },
+    last_payment_error: {
+      type: {
+        api_connection_error: 'api_connection_error',
+        api_error: 'api_error',
+        authentication_error: 'authentication_error',
+        card_error: 'card_error',
+        idempotency_error: 'idempotency_error',
+        invalid_request_error: 'invalid_request_error',
+        rate_limit_error: 'rate_limit_error',
+      },
+    },
+  },
 });
--- a/types/2020-03-02/PaymentIntents.d.ts
+++ b/types/2020-03-02/PaymentIntents.d.ts
@@ -1529,6 +1529,13 @@ declare module 'stripe' {
         id: string,
         options?: RequestOptions
       ): Promise<Stripe.PaymentIntent>;
+
+      constants: {
+        cancellation_reason: {[K in PaymentIntent.CancellationReason]: K};
+        last_payment_error: {
+          type: {[K in PaymentIntent.LastPaymentError.Type]: K};
+        };
+      };
     }
   }
 }


richardm-stripe:
@adamreisnz Since you opened #541 I wondered if you had any feedback about this.

Also would you be willing to elaborate a bit more on why you choose to define constants vs. hardcoding the strings. Do you use typescript?

adamreisnz:
@richardm-stripe No I don't use typescript.

I always define any constant values as constants, because that makes programming a lot easier if you need to deal with ever changing those values.

For example, Stripe uses a string called `month` for billing interval, and the string `active` is a valid value for a subscription status. If Stripe ever decides to change those values, and if I would just be using them as strings in my codebase, I would have to search for them everywhere to change them. You can imagine searching for words like "month" or "active" would bring up a hell of a lot of results that have nothing to do with Stripe.

If there'd be constants defined for each and every one of those strings, then I wouldn't even have to worry about it. Stripe could change the strings 50 times a week for all I cared, and my code would still work, as I would be using Stripe's constant definitions.

For now, I have to make do with my own constant export, which contains a few of the constants that I've encountered and work with:

```js
/**
 * Stripe constants
 */
module.exports = {
  common: {
    NOW: 'now',
  },
  collectionMethods: {
    CHARGE_AUTOMATICALLY: 'charge_automatically',
    SEND_INVOICE: 'send_invoice',
  },
  billingIntervals: {
    MONTH: 'month',
    YEAR: 'year',
  },
  prorationBehaviours: {
    CREATE_PRORATIONS: 'create_prorations',
    ALWAYS_INVOICE: 'always_invoice',
    NONE: 'none',
  },
  subscriptionStatuses: {
    TRIALING: 'trialing',
    ACTIVE: 'active',
    INCOMPLETE: 'incomplete',
    INCOMPLETE_EXPIRED: 'incomplete_expired',
    PAST_DUE: 'past_due',
    CANCELLED: 'canceled',
    UNPAID: 'unpaid',
  },
  taxExemptOptions: {
    EXEMPT: 'exempt',
    NONE: 'none',
    REVERSE: 'reverse',
  },
  taxIdTypes: [
    'nz_gst',
    'au_abn',
    'eu_vat',
  ],
};
```

I use them for example as follows:

```js
//Stripe constants
const {
  subscriptionStatuses: {
    ACTIVE, TRIALING, PAST_DUE, CANCELLED, UNPAID,
    INCOMPLETE, INCOMPLETE_EXPIRED,
  },
} = require('../constants/stripe.js');

//...later in code
if (trialEnded && status === ACTIVE) {
  //do stuff
}
```

If the `active` status ever changes to something else, all I need to do is change it in one place and I don't need to worry about what the string actually is supposed to be.

richardm-stripe:
@adamreisnz Thanks! This is very useful feedback.

I don't think my plan matches your expectations completely. Specifically, as I had planned it, the constant names would always correspond to the most recent version of the Stripe API and exactly match the values.

That is, if we changed "active" to "not_inactive", the SDK would go from
```js
subscriptionStatuses: {
    ACTIVE: 'active',
    ...
}
```
to
```js
subscriptionStatuses: {
    NOT_INACTIVE: 'not_inactive',
    ...
}
```
in the next release.

We would not do
```js
subscriptionStatuses: {
  ACTIVE: 'not_inactive'
}
```

so when you upgraded your version of the SDK, you would still have to find `subscriptionStatuses.ACTIVE` everywhere to change it to `subscriptionStatuses.NOT_INACTIVE`.

So I'm curious
* Is this still advantageous?
* Is this confusing?

adamreisnz:
Yeah, I realise that probably makes sense, but it would still be useful because it should be easier to find usage of and replace constants as opposed to plain strings.

Another advantage of course is that you don't have to remember the string, and avoid potential errors due to typos or misremembered strings. If you are using an IDE you will be able to use auto completion to pick the right constant every time, so I am still all for this proposal 👍

adamreisnz:
Agree with the capitalisation. Constants are generally not lowercase, and a
lot of developers in JS don't use snake case, at least not for lower case
variable names. For constants it would be fine and expected.

PR #859, opened 2020-04-03
Add support for `calculatedStatementDescriptor` on `Charge`

remi-stripe:
Codegen for openapi afb5b18

r? @richardm-stripe @rattrayalex-stripe 
cc @stripe/api-libraries

### DIFF
--- a/types/2020-03-02/Issuing/Authorizations.d.ts
+++ b/types/2020-03-02/Issuing/Authorizations.d.ts
@@ -31,12 +31,12 @@ declare module 'stripe' {
         authorization_method: Authorization.AuthorizationMethod;
 
         /**
-         * The amount that has been authorized. This will be `0` when the object is created, and increase after it has been approved.
+         * [DEPRECATED] The amount that has been authorized. This will be `0` when the object is created, and increase after it has been approved.
          */
         authorized_amount?: number;
 
         /**
-         * The currency that was presented to the cardholder for the authorization. Three-letter [ISO currency code](https://www.iso.org/iso-4217-currency-codes.html), in lowercase. Must be a [supported currency](https://stripe.com/docs/currencies).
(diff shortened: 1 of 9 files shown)


types/2020-03-02/Issuing/Authorizations.d.ts
```diff
@@ -103,17 +103,17 @@ declare module 'stripe' {
...
 
         /**
-         * The amount the user is requesting to be authorized. This field will only be non-zero during an `issuing.authorization.request` webhook.
+         * [DEPRECATED] The amount the user is requesting to be authorized. This field will only be non-zero during an `issuing_authorization.request` webhook.
          */
         pending_authorized_amount?: number;
 
         /**
-         * The additional amount Stripe will hold if the authorization is approved. This field will only be non-zero during an `issuing.authorization.request` webhook.
+         * [DEPRECATED] The additional amount Stripe will hold if the authorization is approved. This field will only be non-zero during an `issuing_authorization.request` webhook.
```
rattrayalex-stripe, line 111: Ideally we'd say what to do instead of the deprecated thing, but not a part of this PR.

types/2020-03-02/Issuing/Cardholders.d.ts
```diff
@@ -6116,7 +6116,7 @@ declare module 'stripe' {
         status?: CardholderListParams.Status;
 
         /**
-         * Only return cardholders that have the given type. One of `individual`, `business_entity`, or `company`.
+         * Only return cardholders that have the given type. One of `individual` or `company`.
```
rattrayalex-stripe, line 6119: Just calling out that this docstring and the one above now disagree with the types. I think that's probably fine (just implies that `business_entity` is rare/unexpected, but possible).

PR #883, opened 2020-05-01
Multiple API changes

remi-stripe:
Multiple API changes
  * Add `issuing` on `Balance`
  * Add `br_cnpj` and `br_cpf` as `type` on `TaxId`
  * Add `price` support in phases on `SubscriptionSchedule`
  * Make `quantity` nullable on `SubscriptionSchedule` for upcoming API version change

Codegen for openapi db4b68c

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/types/2020-03-02/SubscriptionSchedules.d.ts
+++ b/types/2020-03-02/SubscriptionSchedules.d.ts
@@ -254,9 +254,14 @@ declare module 'stripe' {
           plan: string | Stripe.Plan | Stripe.DeletedPlan;
 
           /**
+           * ID of the price to which the customer should be subscribed.
+           */
+          price?: string | Stripe.Price | Stripe.DeletedPrice;
+
+          /**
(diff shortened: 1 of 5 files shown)


types/2020-03-02/SubscriptionSchedules.d.ts
```diff
@@ -253,10 +253,15 @@ declare module 'stripe' {
            */
           plan: string | Stripe.Plan | Stripe.DeletedPlan;
 
+          /**
+           * ID of the price to which the customer should be subscribed.
+           */
+          price?: string | Stripe.Price | Stripe.DeletedPrice;
```
remi-stripe, line 259: Yes this is expected. It wasn't in node because we merged before we finished the change we put in Java here https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/SubscriptionSchedule.java#L730-L734

PR #895, opened 2020-05-18
Multiple API changes

remi-stripe:
Multiple API changes
  * Add support for `issuing_dispute` as a `type` on `BalanceTransaction`
  * Add `balance_transactions` as an array of `BalanceTransaction` on Issuing `Dispute`
  * Add `fingerprint` and `transaction_id` in `payment_method_details[alipay]` on `Charge`
  * Add `transfer_data[amount]` on `Invoice`
  * Add `transfer_data[amount_percent]` on `Subscription`
  * Add `price.created`, `price.deleted` and `price.updated` on `Event`.

Codegen for openapi 58089ba

r? @cjavilla-stripe 
cc @stripe/api-libraries


types/2020-03-02/InvoiceLineItems.d.ts
```diff
@@ -59,7 +59,7 @@ declare module 'stripe' {
       /**
        * The price of the line item.
        */
-      price?: Stripe.Price | null;
+      price: Stripe.Price | null;
```
cjavilla-stripe, line 62: Pretty sure these prices moving from optional to not is fine. It's the one thing that sticks out to me the most in this PR

remi-stripe: yeah it's fine in node because it's a TS definition and it doesn't impact other libs because it's a string/object. If it was a long or decimal that'd be breaking in dotnet :p

PR #897, opened 2020-05-20
Multiple API changes

remi-stripe:
Multiple API changes:
  * Add `anticipation_repayment` as a `type` on `BalanceTransaction`
  * Add `interac_present` as a `type` on `PaymentMethod`
  * Add `payment_method_details[interac_present]` on `Charge`
  * Add `transfer_data` on `SubscriptionSchedule`

Codegen for openapi a8528a8

r? @cjavilla-stripe 
cc @stripe/api-libraries


types/2020-03-02/Checkout/Sessions.d.ts
```diff
@@ -58,9 +58,9 @@ declare module 'stripe' {
         display_items?: Array<Session.DisplayItem>;
 
         /**
-         * The line items purchased by the customer. [Expand](https://stripe.com/docs/api/expanding_objects) this field to include it in the response.
+         * The line items purchased by the customer.
          */
-        line_items?: ApiList<Stripe.LineItem> | null;
+        line_items?: ApiList<Stripe.LineItem>;
```
cjavilla-stripe, line 63: I'm not 100% sure how to confirm if this is right based on the spec.

cjavilla-stripe:
```yaml
        line_items:
          description: The line items purchased by the customer. [Expand](https://stripe.com/docs/api/expanding_objects)
            this field to include it in the response.
          nullable: true
          properties:
            data:
              description: Details about each object.
              items:
                "$ref": "#/components/schemas/item"
              type: array
            has_more:
              description: True if this list has another page of items after this
                one that can be fetched.
              type: boolean
            object:
              description: String representing the object's type. Objects of the same
                type share the same value. Always has the value `list`.
              enum:
              - list
              type: string
            url:
              description: The URL where this list can be accessed.
              maxLength: 5000
              type: string
          required:
          - data
          - has_more
          - object
          - url
          title: PaymentPagesCheckoutSessionListLineItems
          type: object
          x-expandableFields:
          - data
```

remi-stripe: Ah yeah so this is because `line_items` is includable. So by default it's not returned at all, and if it's returned it's a real list it can never be `line_items: null`
there's a bug with how the openapi spec represents includable fields today (which we're fixing) and this change is basically correcting the types

PR #920, opened 2020-06-12
Expose StripeResource on instance

jlomas-stripe:
r? @stripe/api-libraries
cc @stripe/api-libraries @remi-stripe @paulasjes-stripe 

This just surfaces the `StripeResource` on the instance as well, so you can continue to do `const stripe = require('stripe')(apiKeyHere);` but still be able to add resources while working with beta things.

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -104,6 +104,9 @@ function Stripe(key, config = {}) {
 
   this._prevRequestMetrics = [];
   this._enableTelemetry = props.telemetry !== false;
+
+  // Expose StripeResource on the instance too
+  this.StripeResource = Stripe.StripeResource;
 }
 
 Stripe.errors = require('./Error');


remi-stripe:
re-assigning to @richardm-stripe to both approve but also clarify if this should be done in the codegen part instead of here.

jlomas-stripe:
I think it additionally needs a change to one of the preambles in the TS codegen stuff.

richardm-stripe:
I mean, users will get a type error if they try to use it and we *don't* add such a type. The problem is, the type of S`tripeResource` is super complicated and difficult to type well -- this is an approximation, I think -- but it uses `any` a lot, and it only reflects the "public interface" of StripeResource...
https://github.com/stripe/stripe-node/commit/ebc1a3ae98207b5ad978178e80b50d2058af05e1

jlomas-stripe:
Hey @richardm-stripe!  I think in this case it's just a matter of adding `StripeResource` to [here](https://github.com/stripe/stripe-node/blob/master/types/2020-03-02/index.d.ts). since this PR exposes it there; I'm not sure if there's further typing that we can even do wrt custom `StripeResource` implementations.

richardm-stripe:
Merging this without a typedef for now. (I don't think there's much advantage to adding StripeResource: any over omitting the typedef altogether)

PR #938, opened 2020-06-30
Document the 'protocol' configuration option

siliconcat:
There is an undocumented `protocol` option that can be used. I was trying to figure a way to use http instead of https (in the older version of the client I was using in my project) for testing and couldn't find it. But I can see it is now there and just not documented in the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ const stripe = Stripe('sk_test_...', {
   timeout: 1000,
   host: 'api.example.com',
   port: 123,
+  protocol: 'http',
   telemetry: true,
 });
 ```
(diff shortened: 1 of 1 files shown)


richardm-stripe:
Hi @siliconcat, thanks for the pull request.

The `protocol` configuration option does work to send plain HTTP, but it is not officially supported and we don't want to document it as part of the public interface. We recognize there are a few legitimate uses, but we want to discourage as heavily as possible from from sending their credentials over an insecure channel when it can be avoided.

siliconcat:
Hi @richardm-stripe,

Since it is not official I guess it could then be removed at anytime without warning, which leaves me uneasy. IMO it would be better to document it and warn accordingly, and maybe explain its use-cases, rather than obfuscate that it exists. Otherwise I would remove it altogether so that is the best way to "discourage" its use. If it is there, people would find it online (like I did) and even use it in a "bad" way, since there is no warning at all from the library maintainers. 

In my case, I came to it after a few hours of trying to find a good solution to run my app against a mock to capture the payloads we are sending to it in our tests, using Docker. Any other good solution for that? Since I use Docker for Mac, it is tricky to access localhost containers on https, hence using http did save the day.

Many thanks!

richardm-stripe:
Thanks for the feedback and detailing your use case @siliconcat. We're trying the "document it and warn accordingly" as you suggest in #940.

PR #948, opened 2020-07-07
fix StripeConfig.apiVersion type

jaredh159:
The type for `StripeConfig.apiVersion` seems incorrect.  It is not required, and `null` can be passed (which it even states in the comment), but as it stands right now, the only way to avoid a typescript error is to always set the property and set it to _exactly_ the string `"2020-03-02"`.

This PR makes the type match what the client constructor actually accepts -- `apiVersion` is optional, and accepts a string or null. A stricter type could be created by making a union of all possible api versions, like: `'2020-03-02' | '2019-12-03'` etc, but there are so many historical api versions it seems to make more sense just to make the type `string`.

### DIFF
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -21,7 +21,7 @@ declare module 'stripe' {
        *
        * @docs https://stripe.com/docs/api/versioning
        */
-      apiVersion: LatestApiVersion;
+      apiVersion?: null | string;
 
       /**
        * Optionally indicate that you are using TypeScript.


richardm-stripe:
Hi @jaredh159, this type is intentional. The reasoning for this is that the typescript bindings throughout stripe-node are necessarily 'pinned' to a specific API version (presently "2020-03-02"), so if you are using stripe-node with any other api version, the types will not in all cases accurately reflect the responses you receive from the Stripe API in all cases (nor the parameters you must send). 

We try and explain this behavior in the [README](https://github.com/stripe/stripe-node#using-old-api-versions-with-typescript)

> Types can change between API versions (e.g., Stripe may have changed a field from a string to a hash), so our types only reflect the latest API version.
> We therefore encourage upgrading your API version if you would like to take advantage of Stripe's TypeScript definitions.
> If you are on an older API version (e.g., 2019-10-17) and not able to upgrade, you may pass another version or apiVersion: null to use your account's default API version, and use a comment like `// @ts-ignore stripe-version-2019-10-17` to silence type errors here and anywhere the types differ between your API version and the latest. When you upgrade, you should remove these comments.
> We also recommend using `// @ts-ignore` if you have access to a beta feature and need to send parameters beyond the type definitions.

I'm closing the PR, though I wonder if there's a way we can surface this information more explicitly somehow.

jaredh159:
@richardm-stripe yeah, i read the readme, but it seems like a weird usage of typescript.  The underlying javascript function doesn't require the config to be passed at all, and if passed, the `apiVersion` property is not required.  So in this case you're using the TypeScript definitions to sort of create a "I accept your terms and conditions" step, like a strange meta-layer.  In my mind, types exist to document the actual api of the code, not to draw the user's attention to a gotcha about the library.  Seems like a warning in the readme about using the latest api version plus a type that reflects the actual code API would be more appropriate.  Just my 2 cents.

richardm-stripe:
Thanks for the perspective @jaredh159. There is a benefit to the hardcoded version string, beyond confronting users with the apiversion disclaimer, so to speak. For a user who does set their version to the hardcoded string, when we bump the pinned version in the library, and they bump their npm module, the type error they get on the version string will hopefully prevent them from forgetting to change it when they upgrade their method calls to conform to the new version.

I see what you are saying about the strangeness of the Typescript API being more restrictive than the code API. The whole library is strange in this way -- e.g. the type of `stripe.charges.create` prevents you from passing a parameter like `ammount` that does not exist, but nothing in the Javascript implementation prevents this -- stripe-node will gladly forward `ammount` through to the Stripe API, and only throws an exception once the Stripe API has responded with 400.

In theory, we could change this, pin the client to a particular version and perform client-side validation, and then the types would become a true reflection of the code, but the tradeoff is that this prevents the usage of arbitrary versions, for non-typescript users as well as typescript users.

jaredh159:
@richardm-stripe thanks for the comments and your time, i understand your point of view.  👍

Issue #950, opened 2020-07-08
StripeAuthenticationError when loading stripe secret from environment variables


jordykoppen:
Hi all,

I'm currently working on a Stripe Checkout integration and I'm running into an issue when calling my api to create a Stripe session.

I'm using Next.js with API Routes and call my `create-stripe-session` endpoint to generate a session.
I load the Stripe Secret from my .env file create a stripe instance. However, by doing this I get an error from the following error from the Stripe API:
```
StripeAuthenticationError: Invalid API Key provided: "sk_test********************************M6F"
    at IncomingMessage.<anonymous> (********/node_modules/stripe/lib/StripeResource.js:169:21)
    at Object.onceWrapper (events.js:421:28)
    at IncomingMessage.emit (events.js:327:22)
    at endReadableNT (_stream_readable.js:1201:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  type: 'StripeAuthenticationError',
  raw: {
    message: 'Invalid API Key provided: "sk_test********************************M6F"',
    type: 'invalid_request_error',
    headers: {
      server: 'nginx',
      date: 'Wed, 08 Jul 2020 19:05:29 GMT',
      'content-type': 'application/json',
      'content-length': '148',
      connection: 'keep-alive',
      'access-control-allow-credentials': 'true',
      'access-control-allow-methods': 'GET, POST, HEAD, OPTIONS, DELETE',
      'access-control-allow-origin': '*',
      'access-control-expose-headers': 'Request-Id, Stripe-Manage-Version, X-Stripe-External-Auth-Required, X-Stripe-Privileged-Session-Required',
      'access-control-max-age': '300',
      'cache-control': 'no-cache, no-store',
      'www-authenticate': 'Bearer realm="Stripe"',
      'strict-transport-security': 'max-age=31556926; includeSubDomains; preload'
    },
    statusCode: 401,
    requestId: undefined
  },
  rawType: 'invalid_request_error',
  code: undefined,
  doc_url: undefined,
  param: undefined,
  detail: undefined,
  headers: {
    server: 'nginx',
    date: 'Wed, 08 Jul 2020 19:05:29 GMT',
    'content-type': 'application/json',
    'content-length': '148',
    connection: 'keep-alive',
    'access-control-allow-credentials': 'true',
    'access-control-allow-methods': 'GET, POST, HEAD, OPTIONS, DELETE',
    'access-control-allow-origin': '*',
[...]

PR #951, opened 2020-07-08
Detect api keys that begin with quotation marks

richardm-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 
Initializing the client now throws an exception when the provided api key begins with `'` or `"`. This should prevent confusion like https://github.com/stripe/stripe-node/issues/950 -- quotation marks can make the normal server-side error message super confusing.

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -107,6 +107,7 @@ function Stripe(key, config = {}) {
   }
 
   this._prepResources();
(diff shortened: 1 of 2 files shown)


richardm-stripe:
Closing for now, and adding a task to do this in a unified way on our internal ticket tracker

PR #952, opened 2020-07-10
Add new fields to lastResponse: apiVersion, stripeAccount, idempotencyKey

polybuildr:
r? @jlomas-stripe  
cc @stripe/api-libraries @brandur-stripe 

If any of these three fields are set in the headers, copy them over to lastResponse as suggested in #406.
Also add tests for the new fields and refactor an existing test for lastResponse.

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -124,9 +124,24 @@ StripeResource.prototype = {
         const headers = res.headers || {};
         // NOTE: Stripe responds with lowercase header names/keys.
 
-        // For convenience, make Request-Id easily accessible on
+        // For convenience, make some headers easily accessible on
         // lastResponse.
         res.requestId = headers['request-id'];
+
+        const stripeAccount = headers['stripe-account'];
+        if (stripeAccount) {
+          res.stripeAccount = stripeAccount;
+        }
+
+        const apiVersion = headers['stripe-version'];
+        if (apiVersion) {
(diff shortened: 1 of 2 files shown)


polybuildr:
Hi @richardm-stripe! I'm glad you brought that up! I started with hardcoding the key but then came across a problem. The test isn't hermetic and actually communicates with the Stripe API, so if I hardcode an idempotency key, running the test twice gives me an error saying: `No such customer: cus_HdkOjCApDMTkzx`.

I'm not sure why it's that error specifically -- I didn't investigate. I can look into it later if you don't already know off the top of your head.

richardm-stripe:
Ah thanks @polybuildr, I forgot about that. Can we roll our own random identifier instead of introducing a dependency?
`(Math.random() * Number.MAX_SAFE_INTEGER).toString(16)` or something?

polybuildr:
Ah yes, that's a reasonably good solution. I'll update to include a random identifier without using a dependency. :)

polybuildr:
Done! I know it doesn't matter, but I was still uncomfortable using `Math.random` so I used the Node crypto module's [randomBytes](https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback). No additional dependencies, so that should still be okay. :)

richardm-stripe:
Released in 8.74.0

polybuildr:
Thanks for the review! Glad to be of help. :)

PR #959, opened 2020-07-16
Refactor remaining 'var' to 'let/const' usages

vladikoff:
Cleans up the remaining `var` usages and converts them to `let/const` usages. 

Let me know if you can find any other `var`s and I will update the pull request. Thanks!

### DIFF
--- a/lib/Webhooks.js
+++ b/lib/Webhooks.js
@@ -48,7 +48,7 @@ const Webhook = {
         opts.secret
       );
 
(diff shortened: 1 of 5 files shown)


richardm-stripe:
Looks good to me! Thanks for the cleanups.

richardm-stripe:
Released in 8.74.0

vladikoff:
Thank you for looking at these!

PR #960, opened 2020-07-16
Use strict equality check for 'protocol' field for consistency

vladikoff:
Minor change, but the code below uses triple `=`, so might be worth cleaning this up


richardm-stripe:
Looks good to me, thanks for the cleanup!

richardm-stripe:
Released in 8.74.0

PR #968, opened 2020-07-28
Update type definitions for customers.create

architgarg:
What's done - Updated type definitions for CustomerCreateParams to take optional id. 

Issue - Actual customers.create works well with id and create new customer with the specified id, but it was not there is type declarations.

### DIFF
--- a/types/2020-03-02/Customers.d.ts
+++ b/types/2020-03-02/Customers.d.ts
@@ -207,6 +207,11 @@ declare module 'stripe' {
 
     interface CustomerCreateParams {
       /**
+       * Unique identifier for the object.
+       */
+      id?: string;
+      
+      /**
        * The customer's address.
        */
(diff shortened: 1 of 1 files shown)


remi-stripe:
@architgarg Thanks for starting a PR! Unfortunately, this is not something we would document in our client library. Passing your own custom id when creating a customer is a legacy feature that we have deprecated in 2014 and we actively and strongly discourage any integration from relying on this feature publicly. 

There are definitely some legacy integrations out there that still rely on this and some users with business models that require it but since it's discouraged and not documented, it's not something we would add to our Typescript types at the moment.

If your integration is relying on this feature, I'd recommend upgrading it to rely on Stripe's official id generation logic instead and store your own internal id in `metadata` for example. You don't have to move obviously since it does work and we don't plan to remove this so if you need this it's fine to continue using it.

PR #1001, opened 2020-09-11
fix: make documentation more readable and fix typo

rajvijen:
- fixing some typos and grammatical mistakes.
- First, commit to starting contributing in Stripe.


brandur-stripe:
Thanks for the PR @rajvijen! Hopefully this doesn't sound too harsh, but you are actually introducing grammatical mistakes rather than fixing them. Also, "behavior" is the accepted spelling of the word in the US (although I'm from Canada, so I can related to adding a "u"), and we generally go by US English.

Issue #994, opened 2020-09-07
Property 'StripeResource' does not exist on type 'Stripe'


RDGthree:
https://github.com/stripe/stripe-node/pull/920 - PR where StripeResource was exposed.

Looks like this was intentionally merged without a typedef, but I'd personally prefer having this typed as `any` - alternative being littering the code with `@ts-ignore` for the time being. Generally we try to avoid `@ts-ignore` over here just so that Typescript will start yelling at us when types are eventually added.

If there's some clean way for me to cast it to `any` that I'm not thinking about, that would be fine too - the discussion in the PR makes me think I may be missing something easy. Very conscious that opening this issue may have me looking like a bozo - don't hesitate to let me know if I am!

Regardless, it's not a huge deal, just thought this type error was a local problem at first so I figured I'd open the issue for others who might run into it. I won't be offended if you guys want to stick with the original plan and close this issue quickly!

PR #1011, opened 2020-09-21
Typedef for Stripe.StripeResource

richardm-stripe:
Fixes https://github.com/stripe/stripe-node/issues/994
r? @jlomas-stripe 
cc @stripe/api-libraries 

We considered adding this for https://github.com/stripe/stripe-node/pull/920 but then opted not to, but per #994 this is not convenient for users. We could just expose `any` but why not try and get the types right?

The types in this PR are an approximation that reflect mostly the "public interface" of Stripe.StripeResource, and still use `any` a great deal, but it's a starting place.

### DIFF
--- a/types/2020-08-27/index.d.ts
+++ b/types/2020-08-27/index.d.ts
@@ -105,6 +105,8 @@ declare module 'stripe' {
 
     setAppInfo(info: Stripe.AppInfo): void;
 
+    StripeResource: Stripe.StripeResource;
+
     /**
      * Top Level Resources
      */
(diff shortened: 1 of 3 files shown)


jlomas-stripe:
Hey @richardm-stripe!  I know I contributed a bunch to these initially but my TypeScript isn't that great; probably better to have someone else with more familiarity review.  Sorry!

paulasjes-stripe (approved):
Looks good, but should we also include [`MAX_BUFFERED_REQUEST_METRICS`](https://github.com/stripe/stripe-node/blob/master/lib/StripeResource.js#L27) since we've listed everything else on the `StripeResource` object?

richardm-stripe:
ptal @paulasjes-stripe 
added MAX_BUFFERED_REQUEST_METRICS

paulasjes-stripe (approved):

PR #1018, opened 2020-09-25
More idiomatic types

richardm-stripe:
Changes `FileLinkUpdateParams.expires_at^ to be "nullable" rather than "emptyStringable".

stripe-node converts nulls to empty strings before sending over the wire (urlencoded has no concept of 'null'), and using null is a more idiomatic way of "unsetting" a property so our convention for these type bindings is to provide null instead of empty string, but up till now our codegen infrastructure wasn't property handling this case before since it got confused by the presence of "now".

r? @ctrudeau-stripe 
cc @stripe/api-libraries

### DIFF
--- a/types/2020-08-27/FileLinks.d.ts
+++ b/types/2020-08-27/FileLinks.d.ts
@@ -89,7 +89,7 @@ declare module 'stripe' {
       /**
        * A future timestamp after which the link will no longer be usable, or `now` to expire the link immediately.
        */
-      expires_at?: 'now' | number | '';
+      expires_at?: 'now' | number | null;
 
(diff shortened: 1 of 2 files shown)


richardm-stripe:
Not sure why travis is misbehaving but this change is innocuous and `yarn test-typescript`, `yarn lint` and `yarn test` pass locally.

ctrudeau-stripe:
Is it possible to kick travis to retry?

richardm-stripe:
I've tried twice, no luck.

ctrudeau-stripe (approved):
LGTM

PR #1019, opened 2020-09-25
Add support for `oxxo` as a valid `type` on the List PaymentMethod API

remi-stripe:
Codegen for openapi 0cb9b14

r? @ctrudeau-stripe 
cc @stripe/api-libraries


ctrudeau-stripe (approved):
LGTM and builds pass

PR #1020, opened 2020-09-28
add more enums for disabled/failure reasons

JipSterk:


richardm-stripe:
Hi @JipSterk, thanks for the submission. Unfortunately we can't merge your PR as is, because these type bindings are automatically generated from internal code -- but I will look into seeing if we can make a change internally to make the types here more specific!

PR #1033, opened 2020-10-09
Update Typescript types

richardm-stripe:
r? @ctrudeau-stripe 

Updates the type bindings for several discount-related fields to indicate that these fields will now always be present.


ctrudeau-stripe (approved):
LGTM

PR #1034, opened 2020-10-09
Codegen for openapi c08e2d9

richardm-stripe:
r? @ctrudeau-stripe 

Adds new internal-only properties to `card_present`.


ctrudeau-stripe (approved):
LGTM

Issue #1029, opened 2020-10-08
RequestOptions with deprecated key are mutated without removing the old key


sktt:
The follow code will throw an exception with
```
Error: Both 'stripeAccount' and 'stripe_account' were provided; please remove 'stripe_account', which is deprecated.
```
```js
const opts = {stripe_account: "acct_...." }
const invoice = await stripe.finalizeInvoice(invoiceId, opts)
// here opts = {stripe_account: "acct_...", stripeAccount: "acct_..." }
await stripe.paymentIntents.confirm(invoice.payment_intent.id, opts)
```
Cheers

richardm-stripe:
Thank you @sktt for reporting this. I agree, ideally `opts` shouldn't be mutated at all. We're investigating a fix.

PR #1041, opened 2020-10-13
Do not mutate user-supplied opts

vcheung-stripe:
### Summary

When a user supplies opts to a function, the user's opts are mutated. This becomes a problem when a user supplies an opt that is deprecated; e.g. `stripe_account` instead of `stripeAccount`. In this scenario:

- Before calling the function:
```javascript
opts = {
  ...
  stripe_account: ...,  // user supplies a deprecated opt
  ...
}
```
- After calling the function, the object is mutated:
```javascript
opts = {
  ...
  stripe_account: ...,
  stripeAccount: ...,  // the function unexpectedly inserts the non-deprecated version of the opt
  ...
}
```
Because of this mutation, the user is warned that they have supplied both of these args, when in fact they have only supplied one.

This change makes a shallow copy of the user's opts so that this does not happen.

### Motivation
fixes https://github.com/stripe/stripe-node/issues/1029

### Test plan
- Unit tests

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -149,7 +149,7 @@ const utils = (module.exports = {
       if (typeof arg === 'string') {
         opts.auth = args.pop();
       } else if (utils.isOptionsHash(arg)) {
-        const params = args.pop();
+        const params = {...args.pop()};
 
         const extraKeys = Object.keys(params).filter(
           (key) => !OPTIONS_KEYS.includes(key)
--- a/test/makeRequest.spec.js
+++ b/test/makeRequest.spec.js
@@ -1,0 +1,29 @@
+'use strict';
+
+require('../testUtils');
+
+const makeRequest = require('../lib/makeRequest');
+const expect = require('chai').expect;
+
+describe('makeRequest', () => {
+  describe('args', () => {
+    it('does not mutate user-supplied deprecated opts', () => {
+      const args = [
+        {
+          stripe_account: 'bad',
+        },
+      ];
+      const mockSelf = {
+        createResourcePathWithSymbols: () => {},
+        createFullPath: () => {},
+        _request: () => {},
+      };
+      makeRequest(mockSelf, args, {}, {});
+      expect(args).to.deep.equal([
+        {
+          stripe_account: 'bad',
+        },
+      ]);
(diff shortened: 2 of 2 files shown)


lib/makeRequest.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const utils = require('./utils');
+const cloneDeep = require('lodash.clonedeep');
```
richardm-stripe, line 4: I'd try to avoid the extra dependency if at all possible. Do you think it is necessary to deep clone? I haven't checked, but my guess would be we aren't doing any mutations more than a single level down and a shallow clone `{...opts}` would be enough.

richardm-stripe (changes requested):

lib/makeRequest.js
vcheung-stripe: Using shallow copying with the `...` operator, the opts do get mutated, so I think we do need a deep copy. What works is `JSON.parse(JSON.stringify(args))`, of course this isn't the most robust way of copying. Any thoughts?

vcheung-stripe: Actually I may have misread your comment, I will try the shallow copy one level down and report back.

vcheung-stripe: Shallow copying one level down works! We do not need this dependency anymore. ptal @richardm-stripe

richardm-stripe (approved):
LGTM!

jacoblee93:
Hey @vcheung-stripe and @richardm-stripe, we're using Node 6 this line broke our build -- Node 6 doesn't support object destructuring:

`const params = {...args.pop()};`

Can we change this to something else? Version deprecations shouldn't be done in a minor release at any rate. I'll open a PR shortly.

paulasjes-stripe:
Hi @jacoblee93, we ended support for Node 6 as of version 8 of stripe-node: https://github.com/stripe/stripe-node/wiki/Migration-guide-for-v8

I admit that finding that information was harder that it should have been. I'll put up a PR shortly that clearly states our supported versions in the main readme.

jacoblee93:
Dang. Well thanks for letting me know.

PR #1046, opened 2020-10-16
chore: replace recommended extension sublime babel with babel javascript

pavanjadhaw:
Sublime Babel which is no longer available on the vscode marketplace and we get below message when trying to open project in vscode.
Babel Javascript is well maintained alternative.
[differences](https://github.com/michaelgmcd/vscode-language-babel/issues/1)

image

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,7 +2,7 @@
   "recommendations": [
     // JS Language plugins (~required)
     "dbaeumer.vscode-eslint",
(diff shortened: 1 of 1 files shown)


remi-stripe:
Thanks for the PR @pavanjadhaw we'll have a look next week and follow up with any feedback!

richardm-stripe (approved):
Thank you for the contribution @pavanjadhaw!

Issue #1051, opened 2020-10-23
TypeScript: Updating a subscriptions default_tax_rates is incomplete


Bene-Graham:
Node: 14.13.1
Stripe Lib: 8.116.0

The type `SubscriptionUpdateParams` has a field called `default_tax_rates` with the following comment: 

```
The tax rates that will apply to any subscription item that does not have `tax_rates` set. Invoices created will have their `default_tax_rates` populated from the subscription. Pass an empty string to remove previously-defined tax rates.
```

It states to pass an empty string to remove the tax rates, which the typings don't allow.

richardm-stripe:
Hi @Bene-Graham, sorry about the confusion here. You ought to use `null` here. `stripe-node` converts null to empty string over the wire (null is a more idiomatic way of deleting something in Javascript, but Stripe's API uses URL encoding, which has no concept of null), and we generate the docstring from a source that describes the parameter from the API's perspective, not node's.

Bene-Graham:
@richardm-stripe Perfect, that worked. Thank you for your help.

PR #1052, opened 2020-10-26
Add Stripe.Emptyable type definition

richardm-stripe:
r? @suz-stripe 
cc @stripe/api-libraries 

Defines `Stripe.Emptyable<T> = null | '' | T` type. Many params in the Stripe API has a "pass empty string to unset" idiom, and stripe-node converts nulls to empty strings over the wire. Presently, the type bindings encode all of these as `T | null`, but this [is a bit confusing](https://github.com/stripe/stripe-node/issues/1051) because some of the docstrings recommend empty strings.

I think adding an "Emptyable" wrapper like this will make the meaning of these type definitions a little bit more clear, and it also gives us a single place where we can document this behavior.

In VSCode, the present mouseover looks like
image
and the new one will be this:
image

It seems like VSCode omits mention of `| null` altogether, so the new mouseover is the only one that actually communicates the emptyable nature of the field .

### DIFF
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -199,7 +199,7 @@ declare module 'stripe' {
     /**
      * The Stripe API uses url-encoding for requests, and stripe-node encodes a
      * `null` param as an empty string, because there is no concept of `null`
-     * in url-encoding.
+     * in url-encoding. Both `null` and `''` behave identically.
      */
     export type Emptyable<T> = null | '' | T;
 
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -196,6 +196,13 @@ declare module 'stripe' {
       autoPagingToArray(opts: {limit: number}): Promise<Array<T>>;
     }
 
+    /**
+     * The Stripe API uses url-encoding for requests, and stripe-node encodes a
+     * `null` param as an empty string, because there is no concept of `null`
+     * in url-encoding.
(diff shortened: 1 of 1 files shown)


types/lib.d.ts
```diff
@@ -196,6 +196,13 @@ declare module 'stripe' {
       autoPagingToArray(opts: {limit: number}): Promise<Array<T>>;
     }
 
+    /**
+     * The Stripe API uses url-encoding for requests, and stripe-node encodes a
+     * `null` param as an empty string, because there is no concept of `null`
+     * in url-encoding.
```
richardm-stripe, line 202:
```suggestion
    /**
     * The Stripe API uses url-encoding for requests, and stripe-node encodes a
     * `null` param as an empty string, because there is no concept of `null`
     * in url-encoding. Both `null` and `''` behave identically.
```

remi-stripe (approved):
Approving because the code is correct. I don't know much about Typescript though and I'm not sure how this will be used and whether it should be named `Empty` instead of `Emptyable` but deferring to you here. Don't hesitate to ask a review from someone familiar with TS instead

suz-stripe (approved):
I think this looks ok and don't have any strong feelings about the term `Emptyable`. After rolling alternatives around in my head, this seems like the most descriptive and specific way of putting it.

PR #1054, opened 2020-10-27
Replace object spread operator with Object.assign

jacoblee93:
This preserves Node 6 support, as the spread operator is not supported for objects until Node 8.


jacoblee93:
This was introduce in a minor version update in #1041

remi-stripe:
@jacoblee93 Since stripe-node only supports Node 8, 10 and higher as of the last major version, we likely shouldn't merge this since it'd reintroduce the assumption that we support Node 6. Does that work for you if we just close this PR?

ctrudeau-stripe:
We're closing this as we have no intention of fixing this because node6 is EOL.

PR #1057, opened 2020-10-27
Standardize on CRULD order for method definitions

richardm-stripe:
## Summary
In the typescript definitions, the order of methods has always been "create, retrieve, update, list, delete, ...customMethods". This standardizes on that order in the node resource definitions as well

## Notify
r? @ctrudeau-stripe 
cc @stripe/api-libraries


ctrudeau-stripe (approved):
LGTM

Issue #1013, opened 2020-09-22
invoices.retrieveUpcoming return type incorrect


WillSquire:
`invoices.retrieveUpcoming` returns `Stripe.Invoice`, but the upcoming invoice doesn't return an `id`. `id` should either be optional or this should be a seperate type.

richardm-stripe:
Hi @WillSquire, sorry for the slow response here.

We experimented with adding a separate type for "UpcomingInvoice", but unfortunately this is a little bit unwieldy. We don't want to change the type of id to optional, either, because in most cases id is not optional, and that would cause a lot of users to have to do an extra type refinement. I think the right thing to do would be to use the "Omit" utility type in the return value of the method (https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys), but unfortunately it will take a little bit of work to add the ability to express this in our code generation infrastructure. I'm labeling this as "future".

richardm-stripe:
Fixed in #1629, released in https://github.com/stripe/stripe-node/releases/tag/v11.2.0

PR #1060, opened 2020-11-04
Create 'UpcomingInvoice' type

richardm-stripe:
Establishes a more accurate type for the return of `Invoice.retrieveUpcoming`.
Fixes https://github.com/stripe/stripe-node/issues/1013


richardm-stripe:
Closing, I think we are going to favor a more robust solution that results simply in marking "id" as nullable on a regular invoice.

PR #1068, opened 2020-11-17
Adding check to verify secret passed in is for webhooks

ctrudeau-stripe:
When using webhooks you need to pass in the `whsec_` found in the dashboard webhooks section. If you pass in the secret key not related to webhooks it leads to a confusing error. This adds a check to verify it's the webhook secret being passed in. 

r? @richardm-stripe

### DIFF
--- a/lib/Webhooks.js
+++ b/lib/Webhooks.js
@@ -104,6 +104,13 @@ const signature = {
       });
     }
 
+    if (secret.substring(0, 6) != 'whsec_') {
+      throw new StripeSignatureVerificationError({
+        message:
+          'Webhook secret key passed in does not start with whsec_. Webhook secrets can be found here: https://dashboard.stripe.com/webhooks.',
+      });
+    }
+
     const expectedSignature = this._computeSignature(
       `${details.timestamp}.${payload}`,
       secret
--- a/test/Webhook.spec.js
+++ b/test/Webhook.spec.js
@@ -127,6 +127,24 @@ describe('Webhooks', () => {
       }).to.throw(expectedMessage);
     });
 
+    it('should raise a SignatureVerificationError when the secret key does not contain whsec_', () => {
+      const header = stripe.webhooks.generateTestHeaderString({
+        payload: EVENT_PAYLOAD_STRING,
+        secret: SECRET,
+      });
+
+      const expectedMessage =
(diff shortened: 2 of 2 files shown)


richardm-stripe (approved):

lib/Webhooks.js
```diff
@@ -104,6 +104,13 @@ const signature = {
       });
     }
 
+    if (secret.substring(0, 6) != 'whsec_') {
```
paulasjes-stripe, line 107: Can simplify with `!secret.startsWith('whsec_')`

lib/Webhooks.js
```diff
@@ -104,6 +104,13 @@ const signature = {
...
+      throw new StripeSignatureVerificationError({
+        message:
+          'Webhook secret key passed in does not start with whsec_. Webhook secrets can be found here: https://dashboard.stripe.com/webhooks.',
```
paulasjes-stripe, line 110: Suggest rewording this for clarity:

`Wrong signing secret provided. Please ensure you are passing in a valid webhook signing secret that starts with whsec_. Webhook secrets can be found at: https://dashboard.stripe.com/webhooks.`

remi-stripe:
I'm quite ambivalent to merge this as is because the secret is opaque and if we ever change the prefix, like we did for API keys multiple times, then it will lead to confusing error messages for developers trying to debug their code.

paulasjes-stripe:
@remi How would you feel about instead of checking for `whsec_` we check that length of the key passed in is larger than 0 and that the key doesn't begin with `sk_` or `pk_`? Seems unlikely that we'll ever use those prefixes for a webhook signing secret.

Reason for this PR is that a user got confused and passed in their secret key rather than a webhook signing secret.

remi-stripe:
Maybe, but there are other keys prefix like rk_ and ek_. Overall, I would personally punt on this kind of helper methods in client libraries.

lib/Webhooks.js
ctrudeau-stripe: I went with a solution I felt was the most backwards compatible. I'm not sure which version we target for node. If we are up to date and can do startsWith and that won't break anyone.

ctrudeau-stripe:
Punting to find a different solution.

PR #1073, opened 2020-11-18
More consistent namespacing for shared types

richardm-stripe:
r? @ctrudeau-stripe 
cc @stripe/api-libraries 
## Summary

Some types that represent common structures that appear in many places throughout the Stripe API, such as `Address` or `Metadata` are defined in `types/shared.d.ts`, so that their definitions aren't repeated everywhere. However when these types are referenced from other types, currently they are **not** namespaced with `Stripe`, e.g. `Account` has `metadata: Metadata`, not `metadata: Stripe.Metadata`. This is inconsistent with how the other types in the library are referenced, e.g. `Account` has `individual: Stripe.Person`, not just `individual: Person`.

This PR corrects that, and makes the referenced to shared types consistent with all the other references. This should have no affect. The references all refer to the same types, they are just more specific (and I have added some lines to typescript-test specifically to exercise the shared types).


ctrudeau-stripe (approved):
LGTM, thanks for the pairing demo!

PR #1080, opened 2020-11-27
fixed promises example

JDIZM:
promises example wasn't working, fixed it. Now creates a customer with the customer id and sends invoice.


remi-stripe (approved):
Thanks for the fix @JDIZM

PR #1096, opened 2020-12-25
typo receieved -> received

kaznovac:
fix few typos in `autoPagination.js`


remi-stripe (approved):
Thanks for the quick fix @kaznovac

PR #1098, opened 2020-12-26
ts declaration for onDone on autoPagingEach and autoPagingToArray

kaznovac:
add ts declaration for `onDone` callback parameter on `autoPagingEach` and `autoPagingToArray`

as implemented:
- https://github.com/stripe/stripe-node/blob/ec960ead66f9942d68804f1a91dce8988d8d9660/lib/autoPagination.js#L170
- https://github.com/stripe/stripe-node/blob/ec960ead66f9942d68804f1a91dce8988d8d9660/lib/autoPagination.js#L187

### DIFF
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -190,10 +190,14 @@ declare module 'stripe' {
       extends Promise<Response<ApiList<T>>>,
         AsyncIterableIterator<T> {
       autoPagingEach(
-        handler: (item: T) => boolean | void | Promise<boolean | void>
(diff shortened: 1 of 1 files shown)


kaznovac:
i find more strict to use `unknown` instead of `any` for err type, however that would lead to backward api breakage

richardm-stripe (approved):

richardm-stripe:
Hello @kaznovac. Sorry I lost track of this. Thank you for the contribution!

PR #1099, opened 2020-12-26
use LIMIT instead of magic-number (test fails if LIMIT constant is changed)

kaznovac:
uses `LIMIT` constant instead of exact number `7` -> better robustness of the `autoPagination` test if the constants are changed


suz-stripe:
thanks for this patch and sharp spotting @kaznovac!

PR #1104, opened 2021-01-05
make request timeout errors eligible for retry

### LINKED ISSUE #1058: maxNetworkRetries setting does not work
environment: node v10.x on AWS Lambda
package version `"stripe": "^8.84.0"`

we have the following code in our checkout lambda function, as suggested in official docs https://www.npmjs.com/package/stripe#network-retries

```js
const stripe = require('stripe')(process.env.STRIPE_API_KEY, {
  maxNetworkRetries: 3,
  timeout: 4000,
  telemetry: false,
})

module.exports = async (req, res) => {
  // ...
  await stripe.invoices.pay(invoice.id)
  // ...
}
```

in one of executions it thrown exception

```
Request aborted due to timeout being reached (4000ms)
```

we expected request would be retried at least 3 times, but it *did not*, cause Lambda function duration was ~5000ms

_scr_ 2020-10-28 at 14 34 44

richardm-stripe:
fixes #1058
r? @tjfontaine-stripe @ctrudeau-stripe 

Presently errors that result from the node http request timeout [are not eligible to be retried](https://github.com/stripe/stripe-node/compare/richardm-retry-request-timeouts?expand=1#diff-d5d3bf39549deedcf2b6f61e2f2f7f8398c71b21ba178612e0ff5c81a06d1ce5L227-L231).

I doubt this is deliberate and it seems like these requests should generally be retried, the same as ECONNRESET or any other sort of connection error.

This is technically a behavior change, so my inclination is to wait until the next major version, since there are a lot of open network-related issues and it would be good to address several of them.

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -98,17 +98,7 @@ StripeResource.prototype = {
       const timeoutErr = new TypeError('ETIMEDOUT');
       timeoutErr.code = 'ETIMEDOUT';
 
-      req._isAborted = true;
-      req.abort();
-
-      callback.call(
-        this,
-        new StripeConnectionError({
-          message: `Request aborted due to timeout being reached (${timeout}ms)`,
-          detail: timeoutErr,
-        }),
-        null
-      );
+      req.destroy(timeoutErr);
     };
   },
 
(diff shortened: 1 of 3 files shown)


ctrudeau-stripe (approved):
LGTM, great tests. Nice refactor to account for ETIMEDOUT error.

tjfontaine-stripe:
similarly this lgtm, though I think we should consider a `requestDurationTimeout` in addition to `req.setTimeout` behavior here.

PR #1107, opened 2021-01-12
More consistent type definitions

richardm-stripe:

A change in our codegen infra revealed a few inconsistencies/bugs in the Typescript definitions. This PR resolves those.

1. `CustomerSource` is now qualified as `Stripe.CustomerSource` inside `Customers.d.ts`, which is more consistent with similar references. This is a no-op. 

 2. Extra method definitions have been removed from `Refund.d.ts`. This is a bug fix. The cause of this was that the openapi spec defines methods for both `/v1/refunds/{refund}` and `/v1/charges/{charge}/refunds/{refund}`. However, the actual Javascript code in `refunds.js` only supports making requests to the `/v1/refunds/{refund}` routes, so I have removed the extra type definitions because they are incorrect and do not describe valid uses of the library.
 
3. The type definition of `stripe.customers.updateSource` without the `params` field has been removed. It doesn't make sense to call an "update" method without specifying params (although the API will accept it), and this is more consistent with what we do with other "update" methods. This is also more consistent with how we handle the multiple dispatch for `updateSource` in Java, which doesn't provide a paramsless signature.

### DIFF
--- a/types/2020-08-27/Customers.d.ts
+++ b/types/2020-08-27/Customers.d.ts
@@ -698,7 +698,7 @@ declare module 'stripe' {
         id: string,
         params: CustomerSourceCreateParams,
         options?: RequestOptions
-      ): Promise<Stripe.Response<CustomerSource>>;
+      ): Promise<Stripe.Response<Stripe.CustomerSource>>;
 
       /**
        * Retrieve a specified source for a given customer.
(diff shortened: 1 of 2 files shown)


ctrudeau-stripe (approved):
LGTM

PR #1129, opened 2021-03-14
Replaced body-parser package with latest express support

MohamedJakkariya:
**Importance of changes**
- [ ] `body-parser` npm package was deprecated with the latest express.js.
- [ ] So no longer it was used instead we can use inbuild express.json() and so on.
- [ ] Mainly, you're example doesn't work. After research, I was found a solution to this. So finally it was working with Stripe CLI and through the dashboard as well.
- [ ] Hopefully, It'll make sense.

### DIFF
--- a/examples/webhook-signing/node-express/express.js
+++ b/examples/webhook-signing/node-express/express.js
@@ -1,29 +1,31 @@
 require('dotenv').config();
 const stripe = require('stripe')(process.env.STRIPE_SECRET_KEY);
 const express = require('express');
-const bodyParser = require('body-parser');
 
 const webhookSecret = process.env.STRIPE_WEBHOOK_SECRET;
 
 const app = express();
 
-// Use JSON parser for all non-webhook routes
-app.use((req, res, next) => {
-  if (req.originalUrl === '/webhook') {
-    next();
-  } else {
-    bodyParser.json()(req, res, next);
-  }
-});
+// TODO: Use buffer data for stripe webhook routes
+app.use(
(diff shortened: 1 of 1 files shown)


remi-stripe:
@MohamedJakkariya Thanks for opening the PR! I don't think `body-parser` is deprecated at the moment and the code example we have in the repo does work as expected While Express 4 has introduced their own version a few years ago the external npm package should just work.

Looking at the PR you seem to have changed `body` into`rawBody` amongst other things. Maybe the issue you were encountering wasn't about the example itself but having webhook signatures work in your own environment? This can definitely be tricky with node in general as there are many different ways to do this depending on your setup. There's an issue on the repo with many solutions and you might want to look at it directly: https://github.com/stripe/stripe-node/issues/341

Since the example we have locally is simpler and should work for vanilla installs we're going to keep it as it and close this PR but hopefully you're unblocked. If you have follow-up questions don't hesitated to [contact our support team](https://support.stripe.com/contact). We also have developers on IRC in #stripe on Freenode!

MohamedJakkariya:
But I don't know these examples always not working for me so finally coming with this. If it's working fine then no issues.

MohamedJakkariya:
But I don't know these examples always not working for me so finally coming with this. If it's working fine then no issues.

remi-stripe:
@MohamedJakkariya What you have is totally fine if it works for you. We wouldn't update our sample though as it is what works for many developers but as long as you made it work it's great!

PR #1131, opened 2021-03-19
types: export StripeRawError type

### LINKED ISSUE #1130: Export StripeRawError type definition
The `StripeRawError` type is not exported for some reason, and it can be useful for having strong typing on certain situations. For instance:

```ts
try {
  paymentIntent = await stripe.paymentIntents.create(
      ...
  );
} catch (error) {
  if (
    error instanceof Stripe.StripeCardError &&
    error.code === 'authentication_required'
  ) {
    paymentIntent = await stripe.paymentIntents.retrieve(
      (error.raw as StripeRawError).payment_intent?.id,
    );

    return {
      paymentIntentSecret: paymentIntent.client_secret,
      requiresConfirmation: true,
    };
  }

  throw error;
}
```

https://github.com/stripe/stripe-node/blob/82316bdf872ac271819f0cb98769305c21964341/types/Errors.d.ts#L27-L47

rdsedmundo:
Closes #1130

### DIFF
--- a/types/Errors.d.ts
+++ b/types/Errors.d.ts
@@ -24,7 +24,7 @@ declare module 'stripe' {
       | 'authentication_error'
       | 'invalid_grant';
 
-    type StripeRawError = {
+    export type StripeRawError = {
       message?: string;
 
       type: RawErrorType;


richardm-stripe:
Seems reasonable to me! Thanks for the contribution.

PR #1146, opened 2021-04-27
Upgrade devDependencies

richardm-stripe:
This PR bumps some devDependencies that are marked as vulnerable. This is merely for hygiene: the vulnerabilities do not affect this library -- we do not rely on the vulnerable code paths, nor does the development workflow involve live traffic.

The upgraded dependencies are not compatible with Node 8, and so this PR also disables Node 8 in CI. We hope to officially deprecate Node 8 soon, but in the meantime we will be not be able to rely on CI to be sure Node 8 does not break and should verify that any potentially incompatible changes are tested manually.

cc @stripe/api-libraries 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache:
   directories:
     - node_modules
 node_js:
-  - '8'
   - '10'
+  - '16'
 sudo: false
 after_success: yarn report && yarn coveralls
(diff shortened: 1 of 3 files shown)


richardm-stripe:
Oh stop the presses, we lose node 8 in CI with this.

ob-stripe:
Node 8 is past EOL, so maybe we should just drop support for it :)

richardm-stripe:
r? @ob-stripe 

Took your suggestion.

richardm-stripe:
I agree but we should do that in a major.

ob-stripe (approved):
lgtm

seang-stripe (approved):
LGTM

PR #1152, opened 2021-05-05
Remove deprecated bodyParser dep in webhook examples

williammartin:
In express 4.16+, bodyParser is now accessible through the `express` module directly: https://stackoverflow.com/questions/66525078/bodyparser-is-deprecated

image

Tada!

image

Thanks!

### DIFF
--- a/examples/webhook-signing/node-express/express.js
+++ b/examples/webhook-signing/node-express/express.js
@@ -1,7 +1,6 @@
 require('dotenv').config();
 const stripe = require('stripe')(process.env.STRIPE_SECRET_KEY);
 const express = require('express');
-const bodyParser = require('body-parser');
 
 const webhookSecret = process.env.STRIPE_WEBHOOK_SECRET;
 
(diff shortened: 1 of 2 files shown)


richardm-stripe:
Sounds good to me! Express 4.16.0 was [released in 2017-09-28](https://expressjs.com/en/changelog/4x.html) and contained security updates. I think it's probably safe to assume most users will be on a version more recent than this.

williammartin:
Thanks @richardm-stripe.

Can you give some guidance on when you would consider version updates for some of these dependencies? 

I ask because I'm creating a new project with more recent dep versions and experience some type errors. For example, in node 14 (released LTS Nov 2020) something appears to have changed in `http`. I'd be happy to go through and update (I'll need to do so for my own development), but not sure if you'd accept a PR.

Cheers.

richardm-stripe:
What dependencies do you mean? We only have 2 (direct) [production dependencies](https://github.com/stripe/stripe-node/blob/master/package.json#L46-L49) -- `@types/node` and `qs`. Do I understand correctly that you want to bump the version of `@types/node`?

williammartin:
Sorry, I'm referring to the dependencies in the examples package: 

https://github.com/stripe/stripe-node/blob/master/examples/webhook-signing/package.json#L14
https://github.com/stripe/stripe-node/blob/master/examples/webhook-signing/package-lock.json#L48-L51

However, now that I see you are pinned to 13.* in your prod dependencies, I suspect that they should probably move together in this repo - which maybe isn't what you'd want yet.

PR #1157, opened 2021-05-11
Binary streaming

richardm-stripe:
## Notify
r? @dcr-stripe 
cc @stripe/api-libraries  

## Summary
This PR adds support for defining 'streaming' stripe methods on the Stripe client.  Unlike existing methods, 'streaming' methods do not buffer the response body into memory and try to parse it into json -- they pass the underlying `http.Request` object to the caller, so that the caller can pipe it to a file or socket without having to buffer it.

The motivation for this is an upcoming `/pdf` method that will return a pdf instead of JSON.

Also incidentally adds the ability to set "host" per-request, this was useful in my testing.

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -105,6 +105,8 @@ StripeResource.prototype = {
   _addHeadersDirectlyToResponse(res, headers) {
     // For convenience, make some headers easily accessible on
     // lastResponse.
+
+    // NOTE: Stripe responds with lowercase header names/keys.
     res.requestId = headers['request-id'];
     res.stripeAccount = res.stripeAccount || headers['stripe-account'];
     res.apiVersion = res.apiVersion || headers['stripe-version'];
@@ -130,10 +132,14 @@ StripeResource.prototype = {
   },
 
   /**
-   * Used by methods with spec.streaming === true. For these methods, we do not buffer the
-   * response into memory, or do any parsing into stripe objects or errors, we
-   * delegate that all of that to the user and pass back the raw http.Response
-   * object to the callback.
+   * Used by methods with spec.streaming === true. For these methods, we do not
+   * buffer successful responses into memory or do parse them into stripe
+   * objects, we delegate that all of that to the user and pass back the raw
+   * http.Response object to the callback.
+   *
+   * (Unsuccessful responses shouldn't make it here, they should
+   * still be buffered/parsed and handled by _jsonResponseHandler -- see
+   * makeRequest)
    */
   _streamingResponseHandler(req, callback) {
     return (res) => {
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -106,21 +106,9 @@ StripeResource.prototype = {
     // For convenience, make some headers easily accessible on
     // lastResponse.
     res.requestId = headers['request-id'];
-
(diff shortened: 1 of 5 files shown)


lib/StripeResource.js
```diff
@@ -102,7 +102,71 @@ StripeResource.prototype = {
...
 
-  _responseHandler(req, callback) {
+  _addHeadersDirectlyToResponse(res, headers) {
+    // For convenience, make some headers easily accessible on
+    // lastResponse.
+    res.requestId = headers['request-id'];
+
+    const stripeAccount = headers['stripe-account'];
+    if (stripeAccount) {
+      res.stripeAccount = stripeAccount;
```
paulasjes-stripe, line 112: Any reason why we don't do something like

`res.stripeAccount = headers['stripe-account'] || res.stripeAccount;`

for all of these?

richardm-stripe: Only a poorly considered stare decisis, fixed! Thanks!

lib/StripeResource.js
```diff
@@ -102,7 +102,59 @@ StripeResource.prototype = {
...
+      request_end_time: requestEndTime,
+    });
+  },
+
+  /**
+   * Used by methods with spec.streaming === true. For these methods, we do not buffer the
+   * response into memory, or do any parsing into stripe objects or errors, we
+   * delegate that all of that to the user and pass back the raw http.Response
+   * object to the callback.
+   */
```
dcr-stripe, line 137: This threw me off a bit due to the mention of errors. I'd potentially clarify that the errors should be handled before this?

lib/StripeResource.js
```diff
@@ -112,43 +164,8 @@ StripeResource.prototype = {
       });
       res.once('end', () => {
         const headers = res.headers || {};
-        // NOTE: Stripe responds with lowercase header names/keys.
```
dcr-stripe, line 115: Should we keep this note in _addHeadersDirectlyToResponse?

dcr-stripe (approved):
Looks good overall! Extremely minor comments - thanks Richard!

lib/StripeResource.js
richardm-stripe: Great catch

lib/StripeResource.js
richardm-stripe: Sure!

richardm-stripe:
r? @dcr-stripe - the problem was that there were test cases that were supposed to trigger timeouts by never calling `res.end()`. In these cases, the "finish" or "close" events wouldn't fire. There's not really a cute way to automatically close the server in these cases after the test case is finished, so I had the helper expose a `closeServer` function to trigger manually on the test case, for now.
https://github.com/stripe/stripe-node/pull/1157/commits/2923bde5ce413ebfccf2ac0e21049bccde7ae1b3

dcr-stripe (approved):
> problem was that there were test cases that were supposed to trigger timeouts by never calling `res.end()`. In these cases, the "finish" or "close" events wouldn't fire. There's not really a cute way to automatically close the server in these cases after the test case is finished, so I had the helper expose a `closeServer` function to trigger manually on the test case, for now.

Ahhh good find. Makes sense! Thanks for digging into this Richard - LGTM! 🚢

PR #1169, opened 2021-06-07
Deprecate travis

richardm-stripe:
Github actions has been around for a couple months now and appears to be as reliable as Travis, plus requiring less maintenance. It is now de facto our main CI (see broken travis on `stripe-go`). Let's make it official.
acc @stripe/api-libraries


dcr-stripe (approved):
LG! Nice!

PR #1172, opened 2021-06-09
Fix webhook signatures docs link

jsteele-stripe:
Webhook signatures are documented on their own page now. This PR updates a README link to reflect that.


remi-stripe (approved):
thanks for the fix!

PR #1174, opened 2021-06-17
Added paper_check to Invoices/PaymentMethodType

lunaroyster:
Payment Methods includes `paper_check`, but the types don't include it. 

See here: https://stripe.com/docs/invoicing/overview#payment-methods

image

There is another instance of PaymentMethods in `types/2020-08-27/Checkout/Sessions.d.ts` but I'm not sure if paper checks are supported in that case.

### DIFF
--- a/types/2020-08-27/Invoices.d.ts
+++ b/types/2020-08-27/Invoices.d.ts
@@ -617,6 +617,7 @@ declare module 'stripe' {
           | 'fpx'
           | 'giropay'
           | 'ideal'
+          | 'paper_check'
           | 'sepa_credit_transfer'
           | 'sepa_debit'
           | 'sofort';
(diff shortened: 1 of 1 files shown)


remi-stripe:
@lunaroyster Thanks for the PR! This API feature is in private beta/secret right now and not something public that can be accessed by most users. This is why it's not in the types for stripe-node (or other client libraries) and why its endpoints/APIs are not available.

The file you're modifying is automatically generated from our public [openapi spec](https://github.com/stripe/openapi) which lists all public features which is why it's missing.

Unfortunately, that means we wouldn't merge this change since it'd make it look like the feature exists but it can't be used by most developers. I hope this makes sense but let me know if you have any questions!

lunaroyster:
Derp! I didn't even consider that it was unavailable because it's in beta.

Thanks!

PR #1198, opened 2021-07-26
Add support for SearchResult.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Adds support for search methods and `search_result` objects returned by the Stripe API. This requires updating auto pagination to support this. These paginate similarly to lists, however they rely on a `next_page` token included in the response rather than using object IDs + `starting_before`/`ending_after`. Thus, only forward pagination is supported

### Test plan

Added unit tests. We'll want to add some more rigorous tests once a resource is using this.

### DIFF
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -39,6 +39,8 @@ function stripeMethod(spec) {
       callback
     );
 
+    // Please note `spec.methodType === 'search'` is beta functionality and this
+    // interface is subject to change/removal at any time.
     if (spec.methodType === 'list' || spec.methodType === 'search') {
       const autoPaginationMethods = makeAutoPaginationMethods(
         this,
--- a/lib/StripeMethod.js
+++ b/lib/StripeMethod.js
@@ -39,7 +39,7 @@ function stripeMethod(spec) {
       callback
     );
 
-    if (spec.methodType === 'list') {
+    if (spec.methodType === 'list' || spec.methodType === 'search') {
       const autoPaginationMethods = makeAutoPaginationMethods(
         this,
         args,
--- a/lib/autoPagination.js
+++ b/lib/autoPagination.js
@@ -6,15 +6,40 @@ const utils = require('./utils');
 function makeAutoPaginationMethods(self, requestArgs, spec, firstPagePromise) {
   const promiseCache = {currentPromise: null};
   const reverseIteration = isReverseIteration(requestArgs);
-  let listPromise = firstPagePromise;
+  let pagePromise = firstPagePromise;
   let i = 0;
 
-  function iterate(listResult) {
+  // Search and List methods iterate differently.
+  // Search relies on a `next_page` token and can only iterate in one direction.
+  // List relies on either an `ending_before` or `starting_after` field with
+  // an item ID to paginate and is bi-directional.
+  let getNextPagePromise;
+  if (spec.methodType === 'search') {
+    getNextPagePromise = (pageResult) => {
+      if (!pageResult.next_page) {
+        throw Error(
+          'Unexpected: Stripe API response does not have a well-formed `next_page` field, but `has_more` was true.'
+        );
+      }
+      return makeRequest(self, requestArgs, spec, {
+        next_page: pageResult.next_page,
+      });
+    };
+  } else {
+    getNextPagePromise = (pageResult) => {
+      const lastId = getLastId(pageResult, reverseIteration);
+      return makeRequest(self, requestArgs, spec, {
+        [reverseIteration ? 'ending_before' : 'starting_after']: lastId,
+      });
+    };
+  }
+
+  function iterate(pageResult) {
     if (
       !(
-        listResult &&
-        listResult.data &&
-        typeof listResult.data.length === 'number'
+        pageResult &&
+        pageResult.data &&
+        typeof pageResult.data.length === 'number'
       )
     ) {
       throw Error(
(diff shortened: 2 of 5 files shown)


lib/StripeMethod.js
```diff
@@ -20,19 +22,26 @@ const makeAutoPaginationMethods = require('./autoPagination')
...
+    );
+  }
   return function(...args) {
     const callback = typeof args[args.length - 1] == 'function' && args.pop();
 
-    spec.urlParams = utils.extractUrlParams(
-      this.createResourcePathWithSymbols(spec.path || '')
-    );
+    spec.urlParams = !!spec.fullPath
+      ? []
```
richardm-stripe, line 34: `fullPath` is guaranteed not to have url params?

richardm-stripe:
This LGTM, but I wonder - would the code be simpler if we removed all relative paths and used `fullPath` everywhere? I don't think there is any virtue to relative paths.

lib/autoPagination.js
```diff
@@ -6,42 +6,58 @@ const utils = require('./utils');
...
+      // calls paginate based on `ending_before` / `starting_after`.
+
       // Reset counter, request next page, and recurse.
       i = 0;
-      const lastId = getLastId(listResult, reverseIteration);
-      listPromise = makeRequest(self, requestArgs, spec, {
-        [reverseIteration ? 'ending_before' : 'starting_after']: lastId,
-      });
-      return listPromise.then(iterate);
+      if (spec.methodType === 'search') {
```
richardm-stripe, line 37: nit: this code would feel slightly better to me if the `spec.methodType` check was done once, top-level in `makeAutoPaginationMethods`, rather than each time within `iterate`.

Either something like
```javascript
let getNextPagePromise;
if (spec.methodType === 'search') {
  getNextPagePromise = ...
} else {
  getNextPagePromise = ...
}
function iterate(pageResult) {
  ...
  return getPagePromise(...).then(iterate)
}
```
or

```javascript
let iterate;
if (spec.methodType === 'search') {
  iterate = function (pageResult) {
    ...
  };
} else {
  iterate = ...
}
```
(the second would likely require factoring out a couple helper methods to keep it DRY)

richardm-stripe (approved):
Nothing strictly blocking, though I'm curious what you think about fullPath everywhere

lib/StripeMethod.js
dcr-stripe: Fixed!

lib/autoPagination.js
dcr-stripe: Done - good call!

dcr-stripe:
Thanks for the comments!

For full paths, I'll lay down the ground work in this PR and then in a follow up can update the other methods to use it as it depends on basic methods.

r? @richardm-stripe

dcr-stripe:
Removing reviews for a second. I think it'd be good to split our the fullPath logic of this PR into its own PR.

lib/StripeMethod.js
```diff
@@ -20,19 +22,24 @@ const makeAutoPaginationMethods = require('./autoPagination')
  * @param [spec.host] Hostname for the request.
  */
 function stripeMethod(spec) {
+  if (!!spec.path && !!spec.fullPath) {
```
richardm-stripe, line 25: super-nit: let's rule out empty strings too, to forbid
```javascript
{
  fullPath: '',
  path: '',
}
```

```suggestion
  if (!!(spec.path || spec.path === '') && !!(spec.fullPath || spec.path === '')) {
```

dcr-stripe: Resolved in https://github.com/stripe/stripe-node/pull/1200/files.

dcr-stripe:
I split out https://github.com/stripe/stripe-node/pull/1200/files for full path support. This should be good to go now.

r? @richardm-stripe

richardm-stripe (approved):

richardm-stripe (approved):

richardm-stripe:
I am merging this. Despite the fact that this functionality is part of a private beta, this will not affect the public interface of the library and can be safely removed/changed.

lib/StripeMethod.js
```diff
@@ -39,7 +39,7 @@ function stripeMethod(spec) {
       callback
     );
 
-    if (spec.methodType === 'list') {
+    if (spec.methodType === 'list' || spec.methodType === 'search') {
```
richardm-stripe, line 42:
```suggestion
    // Please note `spec.methodType === 'search'` is beta functionality and this
    // interface is subject to change/removal at any time.
    if (spec.methodType === 'list' || spec.methodType === 'search') {
```

types/lib.d.ts
```diff
@@ -18,7 +18,7 @@ declare module 'stripe' {
         method: string;
         path?: string;
         fullPath?: string;
-        methodType?: 'list';
+        methodType?: 'list' | 'search';
```
richardm-stripe, line 21:
```suggestion
        // Please note, methodType === 'search' is beta functionality and is subject to
        // change/removal at any time.
        methodType?: 'list' | 'search';
```

types/lib.d.ts
```diff
@@ -202,6 +202,43 @@ declare module 'stripe' {
       autoPagingToArray(opts: {limit: number}): Promise<Array<T>>;
     }
 
+    /**
+     * A container for paginated lists of search results.
+     * The array of objects is on the `.data` property,
+     * and `.has_more` indicates whether there are additional objects beyond the end of this list.
+     * The `.next_page` field can be used to paginate forwards.
```
richardm-stripe, line 211:
```suggestion
     * The `.next_page` field can be used to paginate forwards.
     *
     * Please note, ApiSearchResult<T> is beta functionality and is subject to change/removal
     * at any time.
```

lib/autoPagination.js
```diff
@@ -6,42 +6,65 @@ const utils = require('./utils');
...
   const reverseIteration = isReverseIteration(requestArgs);
-  let listPromise = firstPagePromise;
+  let pagePromise = firstPagePromise;
   let i = 0;
 
-  function iterate(listResult) {
+  // Search and List methods iterate differently.
+  // Search relies on a `next_page` token and can only iterate in one direction.
+  // List relies on either an `ending_before` or `starting_after` field with
+  // an item ID to paginate and is bi-directional.
```
richardm-stripe, line 15:
```suggestion
  // an item ID to paginate and is bi-directional.
  //
  // Please note: spec.methodType === 'search' is beta functionality and is
  // subject to change/removal at any time.
```

PR #1200, opened 2021-07-27
Add support for fullPath on methods.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Adds support for `fullPath` to StripeMethod, letting methods specify a full path (ie. a path that is not a suffix of the resource base).

This is necessary for search URLs and in the future we plan on migrating all methods to this, removing some of the URL magic behind the scenes.

### Test plan

Added unit tests.

### DIFF
--- a/test/makeRequest.spec.js
+++ b/test/makeRequest.spec.js
@@ -3,6 +3,7 @@
 require('../testUtils');
 
 const makeRequest = require('../lib/makeRequest');
+const utils = require('../lib/utils');
 const expect = require('chai').expect;
 
 describe('makeRequest', () => {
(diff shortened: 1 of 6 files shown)


test/makeRequest.spec.js
```diff
@@ -26,4 +27,42 @@ describe('makeRequest', () => {
...
+      const args = ['hello', 'world'];
+
+      const fullPath = '/v1/parent/{param1}/child/{param2}';
+      const spec = {
+        fullPath,
+        urlParams: utils.extractUrlParams(fullPath),
+      };
+
+      let actualPath;
+      const mockSelf = {
```
richardm-stripe, line 42: nice

richardm-stripe (approved):

Issue #1202, opened 2021-07-27
Spawning process for every parallel request results in high CPU usage


richardscarrott:
So this issue was [raised a while back](https://github.com/stripe/stripe-node/issues/313) but there wasn't any reproduction steps which I guess is why it was closed.

I was profiling our node server and noticed from cold one of our APIs was spending a lot of time in the stripe SDK.

Here's a contrived way to reproduce the issue:

```js
const paymentIntentIds = new Array(50).fill('pi_123');
await Promise.all(paymentIntentIds.map(id => stripe.paymentIntents.retrieve(id));
```

If run from cold (i.e. without having made a prior call to the SDK), it will spawn a new process for all 50 of the above requests because the [`getClientUserAgent` caching is a little naive](https://github.com/stripe/stripe-node/blob/40eaaab26cc7d28b718daa649d129061e5d8a6e0/lib/stripe.js#L409).

You can see the difference between a cold call:
<img width="1363" alt="Screenshot 2021-07-27 at 18 11 51" src="https://user-images.githubusercontent.com/539309/127207453-be86f96d-1e5f-4623-8954-41cfb2e99392.png">

And a 2nd warm call:
<img width="972" alt="Screenshot 2021-07-27 at 18 12 57" src="https://user-images.githubusercontent.com/539309/127207479-34e19d48-d83e-4dc6-b6f3-e1cc0dcb1797.png">

If you're wondering why one might be making so many parallel requests at once, our use case is fetching a page of orders in our DB which all have a FK reference to a payment intent in stripe (we don't currently replicate the payment intents in our DB) and unlike older stripe resources, there's no batch endpoint for payment intents AFAIK.

I discovered the issue in 7.5.3 but can also reproduce in **v8.165.0**.

richardscarrott:
In the meantime, I've patched our stripe client with:

```js
Stripe.prototype.getClientUserAgent = cb =>
  cb(`HACK: https://github.com/stripe/stripe-node/issues/1202`);
```

Hopefully won't cause any problems with the telemetry on your end?

richardm-stripe:
Hello @richardscarrott, thank you for the report and the great detective work. A fix for this in #1205 has been merged and will be released soon.

PR #1205, opened 2021-07-28
Prevent fetching the user agent concurrently

richardm-stripe:
r? @dcr-stripe 

## Motivation
Fixes https://github.com/stripe/stripe-node/issues/1202

## Summary
Right now, when calculating the proper user agent, we check a cache and shell out to `uname` if the cache is empty. This is bad because we can shell out to `uname` multiple times concurrently before the cache is populated. For some bulk operation usage patterns this causes an unwanted CPU spike (see #1202). It would be better to only shell out once.

This PR makes two changes:

1. Makes the cache use a promise. Promises are very good at resolving a value only once.
2. Caches the result of `uname` directly, instead of the derived `user_agent`. This allows me to remove some questionable cache invalidation. We should only ever need to call `uname` once, `uname` is not expected to change.

**Breakingness:**

Stripe.USER_AGENT_SERIALIZED isn't marked as private, but I think we should consider it safe to just remove this property. It was an internal cache and I cannot imagine any legitimate use case relying on it.
## Tests
Existing tests for `getClientUserAgent` pass.

### DIFF
--- a/lib/stripe.js
+++ b/lib/stripe.js
@@ -21,7 +21,8 @@ Stripe.USER_AGENT = {
   typescript: false,
 };
 
-Stripe.USER_AGENT_SERIALIZED = null;
+/** @private */
+Stripe._UNAME_CACHE = null;
 
 const MAX_NETWORK_RETRY_DELAY_SEC = 2;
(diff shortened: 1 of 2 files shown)


dcr-stripe (approved):
LGTM!

Issue #1204, opened 2021-07-28
Stripe.Response<Stripe.Charge> does not contain header information


joeythomaschaske:
The `charges.create` method returns a `Stripe.Response<Stripe.Charge>` object. The `Stripe.Response` object is supposed to have a `headers` attribute like so

```typescript
export type Response<T> = T & {
  headers: {[key: string]: string};
  lastResponse: {
    requestId: string;
    statusCode: number;
    apiVersion?: string;
    idempotencyKey?: string;
    stripeAccount?: string;
  };
};
```

In reality, this attribute does not exist in the raw JSON.

In a [conversation](https://discord.com/channels/841573134531821608/841573134531821616/869978109833777192) on the stripe discord server it appears the header information is actually stored on the `lastResponse` object. But again looking at the `Stripe.Response` definition, it is not present on that object.

That leads us to having to rely on undocumented attributes and telling typescript to ignore them

```typescript
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
console.log(`Last Response: ${JSON.stringify(charge.lastResponse?.headers)}`);
```

joeythomaschaske:
One thing that is odd is that when the `charges.create` throws an exception, the error object contains the header information

```json
{
    "type": "StripeInvalidRequestError",
    "raw": {
        "code": "token_in_use",
        "doc_url": "https://stripe.com/docs/error-codes/token-in-use",
        "message": "There is currently another in-progress request using this Stripe token (that probably means you clicked twice, and the other charge is still going through): ba_1JFKbuHaVrFFEQvKxUkfstDG. This token can not be used again if that charge is successful.'",
        "type": "invalid_request_error",
        "headers": {
            "server": "nginx",
            "date": "Wed, 28 Jul 2021 16:14:28 GMT",
            "content-type": "application/json",
            "content-length": "420",
            "connection": "keep-alive",
            "access-control-allow-credentials": "true",
            "access-control-allow-methods": "GET, POST, HEAD, OPTIONS, DELETE",
            "access-control-allow-origin": "*",
            "access-control-expose-headers": "Request-Id, Stripe-Manage-Version, X-Stripe-External-Auth-Required, X-Stripe-Privileged-Session-Required",
            "access-control-max-age": "300",
            "cache-control": "no-cache, no-store",
[...]

PR #1206, opened 2021-07-28
Fix Typescript definition for `StripeResource.LastResponse.headers`

richardm-stripe:
## Notify
cc @stripe/api-libraries 

## Summary
Move typescript definition for `headers` to the correct location. 

You can see that they are added to the object that eventually becomes `.lastResponse` [here](https://github.com/stripe/stripe-node/blob/master/lib/StripeResource.js#L174)

## Motivation
Fixes https://github.com/stripe/stripe-node/issues/1204


dcr-stripe (approved):
LGTM! Thanks @richardm-stripe !

PR #1217, opened 2021-08-12
Update nock.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Updates `nock` version. The original version had a bug with timeouts where `req.destroy` with a timeout error didn't eventually send an `error` event so we couldn't test for timeouts and was problematic when testing the new HTTP client interface.


richardm-stripe (approved):
Sounds like you had fun today! Thanks!!

Issue #997, opened 2020-09-09
[Feature Request] Using stripe-node outside of Node.js (e.g. Cloudflare Workers or Deno)


brillout:
I'm trying to setup a Stripe service as a serverless function using the Cloudflare Workers platform.

But the `stripe-node` package seems to depend on Node.js APIs. Is it possible to use a subset of `stripe-node`'s functionalities for Cloudflare Workers?

Is supporting Cloudlfare Workers something you are interested in? Using Stripe in a no-cold-start serverless environment would be quite lovely.

Would it be possible to extract the business logic that is agnostic to Node.js and expose it to users? So that we can use all kinds of platforms beyond Node.js, such as the Deno or Cloudflare Workers. It could be a new package, e.g. `stripe-js`.

Thanks for what you are doing at Stripe, it's such a lovely product.

Related tickets: https://github.com/stripe/stripe-node/issues/871 https://github.com/stripe/stripe-node/issues/773.

richardm-stripe:
Hi @brillout! Thanks for the feature request.

Cloudflare or Deno isn't something we've investigated in depth yet but it would be interesting to learn more. You might be able to help us out a little bit by elaborating a little bit about your use case -- how you would interact with with Stripe from a Cloudflare worker and what value this would unlock for you.

We did a little bit of work towards decoupling the library from the 'child_process' API in #775, which we were just using incidentally. If I understand correctly, I think in order to meet your need we would need to decouple from the 'http' Node API as well, that we [use](https://github.com/stripe/stripe-node/blob/c6a9d5bf16aeb3d5876a7ffe2ebcc133f88b7b35/lib/StripeResource.js#L432) in [StripeResource](https://github.com/stripe/stripe-node/blob/c6a9d5bf16aeb3d5876a7ffe2ebcc133f88b7b35/lib/StripeResource.js#L17-L18), and we are unfortunately quite far from today.

brillout:
Hi @richardm-stripe !

I want to be able to accept donations with credit cards on https://www.clocktab.com/ and https://www.timer-tab.com/. These two website are static/serverless; I don't need to maintain any server which is super convenient. I'd like to keep the serverless architecture.

I could use AWS lambda instead and that would just work. But I'm intrigued by the Cloudflare Workers Webassembly approach and avoiding cold start would be a neat thing. So my use case is not crucial; I'm just checking out options.

[...]

PR #1218, opened 2021-08-12
Add an HttpClient interface and NodeHttpClient implementation.

dcr-stripe:
### Notify

r? @richardm-stripe + @tjfontaine-stripe (for extra 👀 )

### Summary

Adds an `HttpClient` interface and moves out the Node `http`/`https` request logic into a `NodeHttpClient` implementation. Update Stripe config to allow specifying a custom `HttpClient`.

This is the first step in allowing a custom client (say one that uses fetch for https://github.com/stripe/stripe-node/issues/997).

The interface is summarized in the typescript. In brief:
- `makeRequest` returns a `Response` as a promise.
- `Response#toStream(streamCompleteCallback)` returns a streamable version of the content. The Node behavior is unchanged here of just returning the raw response itself with the headers tacked on.
- `Response#toJSON()` consumes the response and returns the JSON representation as a promise.

This should effectively be a no-op, but this should be released as its own version in case there are issues.

You can see an example rough implementation of this for fetch in https://github.com/stripe/stripe-node/pull/1208.

### Test plan

Added unit tests both to `StripeResource` and to `NodeHttpClient` itself.

### DIFF
--- a/lib/net/HttpClient.js
+++ b/lib/net/HttpClient.js
@@ -1,0 +1,60 @@
+'use strict';
+
+/* eslint-disable class-methods-use-this */
+
+/**
+ * Encapsulates the logic for issuing a request to the Stripe API. This is an
+ * experimental interface and is not yet stable.
+ */
+class HttpClient {
+  makeRequest(
+    host,
+    port,
+    path,
+    method,
+    headers,
+    requestData,
+    protocol,
+    timeout
+  ) {
+    throw new Error('makeRequest not implemented.');
+  }
+
+  /** Helper to make a consistent timeout error across implementations. */
+  static makeTimeoutError() {
+    const timeoutErr = new TypeError(HttpClient.TIMEOUT_ERROR_CODE);
+    timeoutErr.code = HttpClient.TIMEOUT_ERROR_CODE;
+    return timeoutErr;
+  }
+}
+
+HttpClient.TIMEOUT_ERROR_CODE = 'ETIMEDOUT';
+
+HttpClient.Response = class {
+  constructor(statusCode, headers) {
+    this._statusCode = statusCode;
+    this._headers = headers;
+  }
+
+  getStatusCode() {
+    return this._statusCode;
+  }
+
+  getHeaders() {
+    return this._headers;
+  }
+
+  getRawResponse() {
+    throw new Error('getRawResponse not implemented.');
+  }
+
+  toStream(streamCompleteCallback) {
+    throw new Error('toStream not implemented.');
+  }
+
+  toJSON() {
+    throw new Error('toJSON not implemented.');
+  }
+};
+
+module.exports = HttpClient;
--- a/lib/net/HttpClient.js
+++ b/lib/net/HttpClient.js
@@ -30,7 +30,7 @@ class HttpClient {
 
 HttpClient.TIMEOUT_ERROR_CODE = 'ETIMEDOUT';
 
-HttpClient.Response = class {
+class HttpClientResponse {
   constructor(statusCode, headers) {
     this._statusCode = statusCode;
     this._headers = headers;
@@ -55,6 +55,6 @@ HttpClient.Response = class {
   toJSON() {
     throw new Error('toJSON not implemented.');
   }
-};
+}
 
-module.exports = HttpClient;
+module.exports = {HttpClient, HttpClientResponse};
--- a/lib/net/NodeHttpClient.js
+++ b/lib/net/NodeHttpClient.js
@@ -3,7 +3,7 @@
 const http = require('http');
 const https = require('https');
 
-const HttpClient = require('./HttpClient');
+const {HttpClient, HttpClientResponse} = require('./HttpClient');
 
 const defaultHttpAgent = new http.Agent({keepAlive: true});
 const defaultHttpsAgent = new https.Agent({keepAlive: true});
@@ -51,7 +51,7 @@ class NodeHttpClient extends HttpClient {
       });
 
       req.on('response', (res) => {
-        resolve(new NodeHttpResponse(res));
+        resolve(new NodeHttpClientResponse(res));
       });
 
       req.on('error', (error) => {
@@ -80,7 +80,7 @@ class NodeHttpClient extends HttpClient {
   }
 }
 
-class NodeHttpResponse extends HttpClient.Response {
+class NodeHttpClientResponse extends HttpClientResponse {
   constructor(res) {
     super(res.statusCode, res.headers || {});
     this._res = res;
@@ -117,6 +117,4 @@ class NodeHttpResponse extends HttpClient.Response {
   }
 }
 
-NodeHttpClient.Response = NodeHttpResponse;
-
-module.exports = NodeHttpClient;
+module.exports = {NodeHttpClient, NodeHttpClientResponse};
--- a/lib/net/NodeHttpClient.js
+++ b/lib/net/NodeHttpClient.js
@@ -1,0 +1,122 @@
+'use strict';
+
+const http = require('http');
+const https = require('https');
+
+const HttpClient = require('./HttpClient');
+
+const defaultHttpAgent = new http.Agent({keepAlive: true});
+const defaultHttpsAgent = new https.Agent({keepAlive: true});
+
+/**
+ * HTTP client which uses the Node `http` and `https` packages to issue
+ * requests.`
+ */
+class NodeHttpClient extends HttpClient {
+  constructor(agent) {
+    super();
+    this._agent = agent;
+  }
+
+  makeRequest(
+    host,
+    port,
+    path,
+    method,
+    headers,
+    requestData,
+    protocol,
+    timeout
+  ) {
+    const isInsecureConnection = protocol === 'http';
+
+    let agent = this._agent;
+    if (!agent) {
+      agent = isInsecureConnection ? defaultHttpAgent : defaultHttpsAgent;
+    }
+
+    const requestPromise = new Promise((resolve, reject) => {
+      const req = (isInsecureConnection ? http : https).request({
+        host: host,
+        port: port,
+        path,
+        method,
+        agent,
+        headers,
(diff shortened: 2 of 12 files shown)


types/net/net.d.ts
```diff
@@ -0,0 +1,70 @@
...
+         * The streamCompleteCallback should be invoked by the response
+         * implementation when the stream has been consumed.
+         */
+        toStream(streamCompleteCallback: () => void): any; //eslint-disable-line @typescript-eslint/no-explicit-any
+
+        /**
+         * Converts the response content into a JSON object, failing if JSON
+         * couldn't be parsed.
+         */
+        toJSON(): Promise<object>;
```
jt-stripe, line 33: Do we have a union type that's, like, all possible stripe resource types?

types/net/net.d.ts
```diff
@@ -0,0 +1,70 @@
...
+      makeRequest(
+        host: string,
+        port: string | number,
+        path: string,
+        method: 'GET' | 'POST' | 'PUT' | 'DELETE',
+        headers: {[key: string]: string},
+        requestData: string | null,
+        protocol: Stripe.HttpProtocol,
+        timeout: number
+      ): Promise<HttpClient.Response>;
```
jt-stripe, line 67: I'm not a Typescript expert, but is this declaration required when it's explicitly saying that it implements the `HttpClient` interface?

Also is it important that the headers are a `{[key: string]: string}` but an `object` in the interface?

lib/net/NodeHttpClient.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
+
+const http = require('http');
+const https = require('https');
+
+const HttpClient = require('./HttpClient');
+
+const defaultHttpAgent = new http.Agent({keepAlive: true});
+const defaultHttpsAgent = new https.Agent({keepAlive: true});
```
jt-stripe, line 9: Is there a reason not to move this into the constructor? It'd be nice to not have side effects on import in the case that this is being imported in a non-node environment.

lib/stripe.js
```diff
@@ -49,6 +50,9 @@ const {emitWarning} = utils;
 Stripe.StripeResource = require('./StripeResource');
 Stripe.resources = resources;
 
+Stripe.HttpClient = require('./net/HttpClient');
+Stripe.NodeHttpClient = require('./net/NodeHttpClient');
```
jt-stripe, line 54: This will fail in non-Node environments. Maybe instead of exposing an instantiated `NodeHttpClient` we could expose a `createNodeHttpClient` function?

```js
Stripe.createNodeHttpClient = (agent) => {
   const NodeHttpClient = require('./net/NodeHttpClient');
   return new NodeHttpClient(agent);
};
```

lib/stripe.js
```diff
@@ -92,7 +98,8 @@ function Stripe(key, config = {}) {
       props.maxNetworkRetries,
       0
     ),
-    agent: props.httpAgent || null,
+    agent: agent,
+    httpClient: props.httpClient || new Stripe.NodeHttpClient(agent),
```
jt-stripe, line 102: Then taking my suggestion above this would be 
```js
httpClient: props.httpClient || Stripe.createNodeHttpClient(agent)
```

jt-stripe (commented):
A few questions about the TS definitions and a suggestion for how to expose this... I didn't carefully review all the implementation details yet!

types/net/net.d.ts
dcr-stripe: > I'm not a Typescript expert, but is this declaration required when it's explicitly saying that it implements the HttpClient interface?

Unfortunately yes. See [this playground](https://www.typescriptlang.org/play?ssl=5&ssc=1&pln=5&pc=24#code/JYOwLgpgTgZghgYwgAgILIN4ChnICYD2AygQLYRgAWoA5gBQAeAXMiAK6kBG0AlMiwDcCwPAG4sAXyxYEAGzgBnBcgBCyYKQAOsiOXDL0GCUA) as an example. If we remove it, we get the error `Class 'NodeHttpClient' incorrectly implements interface 'HttpClient'.
  Property 'makeRequest' is missing in type 'NodeHttpClient' but required in type 'HttpClient'.

> Also is it important that the headers are a {[key: string]: string} but an object in the interface?

Good catch, that was a stale type. Both should be `object`.
`

types/net/net.d.ts
dcr-stripe: I don't think so unfortunately. @richardm-stripe to confirm?

richardm-stripe: We don't but could easily have one if we wanted.

Though `object` feels better to me here, since `HTTPClient` should be mostly stripe-domain-agnostic.

lib/stripe.js
dcr-stripe: Done. Good call.

lib/net/NodeHttpClient.js
richardm-stripe: We do need a singleton -- if the user instantiates multiple stripe clients (without overriding the agent) then they should all share the same connection pool. We could lazily instantiate it when the constructor is called, though.

lib/net/NodeHttpClient.js
```diff
@@ -0,0 +1,122 @@
...
+          reject(e);
+        }
+      });
+    });
+  }
+}
+
+NodeHttpClient.Response = NodeHttpResponse;
+
+module.exports = NodeHttpClient;
```
richardm-stripe, line 122: very nit: WDYT of 
```
module.exports = {NodeHttpClient, NodeHttpResponse}
```
putting one "on" the other feels strange to me (although I guess we put all this stuff on `Stripe`, so...)

lib/net/NodeHttpClient.js
jt-stripe: I think this is resolved by @dcr-stripe 's latest changes

lib/net/NodeHttpClient.js
dcr-stripe: Done.

lib/net/HttpClient.js
```diff
@@ -0,0 +1,60 @@
...
+    headers,
+    requestData,
+    protocol,
+    timeout
+  ) {
+    throw new Error('makeRequest not implemented.');
+  }
+
+  /** Helper to make a consistent timeout error across implementations. */
+  static makeTimeoutError() {
```
richardm-stripe, line 24: This is great

types/net/net.d.ts
```diff
@@ -0,0 +1,54 @@
...
+      /**
+       * This returns the content as a stream. This is implementation-specific
+       * and so we cannot use a given type as different platforms have
+       * different stream primitives.The expectation is that content
+       * will not have been buffered into memory at this point by the client.
+       *
+       * The streamCompleteCallback should be invoked by the response
+       * implementation when the stream has been consumed.
+       */
+      toStream(streamCompleteCallback: () => void): any; //eslint-disable-line @typescript-eslint/no-explicit-any
```
richardm-stripe, line 43: Should we pursue generic type parameters?
```
export interface HttpClient<Response> {
  makeRequest(...): Promise<Response>
}

interface HttpClientResponse<RawResponse, Stream> {
  ...
  getRawResponse(): RawResponse,
  toStream(...): Stream,
  ...
}
```

test/StripeResource.spec.js
```diff
@@ -557,38 +657,96 @@ describe('StripeResource', () => {
...
+          {},
+          (req, res) => {
+            res.write(JSON.stringify(returnedCharge));
+            res.end();
+          },
+          async (err, stripe, closeServer) => {
+            if (err) {
+              return done(err);
+            }
+            const result = await stripe.charges.create(options.data);
```
richardm-stripe, line 698: I think this is missing a `.catch` so if stripe.charges.create fails the error won't be properly propagated to the test runner

test/StripeResource.spec.js
```diff
@@ -208,6 +216,98 @@ describe('StripeResource', () => {
         );
       });
 
+      it('throws an error on invalid JSON', (done) => {
```
richardm-stripe, line 219: Would there be anyway to add a test for when there's a connection error *after* the headers have been transmitted, but halfway through when the body is being transmitted?

(this should be pretty similar to the test case above on line #200).

That's the one case I can think of where (correct me if I'm wrong) I don't think we have any test coverage but it still would be good to know that this PR is consistent with the previous behavior (and that `fetch` and `nodehttpclient` will behave consistently, too)

richardm-stripe (commented):
This is absolutely wonderful. In particular I think the decision to use a promise interface for HttpClient vs. more conservatively attempting to preserve the callbacky interface represents a huge improvement to the understandability of the code and will make a far better experience for implementers of this interface.

I left a few comments, though just some ideas, nothing blocking. I'll re-assign you to the PR, feel free to assign me again once you've had a look/responded and I will approve.

test/StripeResource.spec.js
dcr-stripe: Done.

types/net/net.d.ts
dcr-stripe: Done.

test/StripeResource.spec.js
dcr-stripe: Done.

dcr-stripe (commented):
Great ideas all around. PTAL @richardm-stripe

test/net/NodeHttpClient.spec.js
```diff
@@ -0,0 +1,215 @@
...
+ */
+class ArrayReadable extends Readable {
+  constructor(values) {
+    super();
+    this._index = 0;
+    this._values = values;
+  }
+
+  _read() {
+    if (this._index == this._values.length) {
```
richardm-stripe, line 26:
```suggestion
    if (this._index === this._values.length) {
```

richardm-stripe (approved):
LGTM, (pending fixing that TS complaint about `any`)

PR #1221, opened 2021-08-16
Add client name property to HttpClient.

dcr-stripe:
r? @richardm-stripe 

Adds a client name property to be implemented by `HttpClient` which will be shipped in the user agent under the `httplib` property similar to Python.

### DIFF
--- a/lib/net/HttpClient.js
+++ b/lib/net/HttpClient.js
@@ -7,13 +7,9 @@
  * experimental interface and is not yet stable.
  */
 class HttpClient {
-  constructor(name) {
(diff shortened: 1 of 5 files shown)


richardm-stripe (approved):
LGTM!

dcr-stripe:
Added a test. I also moved it out of the constructor and into a hard coded method, since it's unnecessary to hold on to the state and regardless implementations that don't include this will fail. Happy to move it back if you'd rather.

PTAL @richardm-stripe

richardm-stripe (approved):

PR #1229, opened 2021-09-01
Add setting to set additional `https.request` params

bryanvaz:
## Changes made
Adds the `requestOptions` parameter to the `Stripe` class constructor and `makeRequest` function. Can be used to send additional params to the `https.request()` function call, which cannot be set any other way. 

## Purpose of changes
Initially created to allow local testing against HTTPS mock endpoints without setting the `NODE_EXTRA_CA_CERTS` or `NODE_TLS_REJECT_UNAUTHORIZED` command line options, allowing the `https` module to trust self-signed certificates used by the local mock HTTPS endpoints.

This setting is also required when using HTTP Proxy-based PCI data vaults which tokenize and detokenize data in-flight (e.g. verygoodsecurity/vgs-satellite#194) without revealing card data to the client. Passing the vault's RootCA to the Stripe client only when required is far safer than trusting the vault's CA globally (via `NODE_EXTRA_CA_CERTS`) or ignoring all invalid certificate completely (via `NODE_TLS_REJECT_UNAUTHORIZED`), both of which open a system up to MITM attacks.

This is also much safer than monkey patching the `https.request` or `tls.secureContext` functions to inject additional options.

## Sample Usage
```js
  apiVersion: '2019-08-08',
  maxNetworkRetries: 1,
  timeout: 1000,
  host: 'api.example.com',
  port: 123,
  telemetry: true,
  requestOptions: {
    ca: fs.readFileSync('/path/to/local/rootca.pem')
  },
});
```

## Breaking change
No. `requestOptions` parameter is completely optional and defaults to an empty object if not set. Does not affect any existing functionality.

## Documentation updated
* README


pakrym-stripe:
Hi, I appologize for the delayed response. 

The `Stripe` object constructor allows providing a custom [HttpClient](https://github.com/stripe/stripe-node/blob/master/src/net/HttpClient.ts) instance. With that you can override any behaviors related to sending HTTP requests .

You can inherit [the default implementation](https://github.com/stripe/stripe-node/blob/master/src/net/NodeHttpClient.ts)  and override the `request` method to your needs.

pakrym-stripe:
Closing for housekeeping purposes. Please free to reopen!

PR #1236, opened 2021-09-08
Add an HTTP client which uses fetch.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Adds an `HttpClient` implementation that uses the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

This paves the way for unblocking https://github.com/stripe/stripe-node/issues/997.

### Test plan

The common tests are shared between NodeHttpClient and FetchHttpClient to make sure both meet the requirements for the interface (modulo some streaming/raw response tests, as this is implementation-dependant). In order to test `fetch` as part of our Node testing environment, the [node-fetch](https://www.npmjs.com/package/node-fetch) polyfill is used.

### DIFF
--- a/lib/net/FetchHttpClient.js
+++ b/lib/net/FetchHttpClient.js
@@ -13,12 +13,10 @@ class FetchHttpClient extends HttpClient {
     this._fetchFn = fetchFn;
   }
 
-  /* eslint-disable class-methods-use-this */
   /** @override. */
   getClientName() {
     return 'fetch';
   }
-  /* eslint-enable class-methods-use-this */
 
   makeRequest(
     host,
--- a/lib/net/FetchHttpClient.js
+++ b/lib/net/FetchHttpClient.js
@@ -4,8 +4,8 @@ const {HttpClient, HttpClientResponse} = require('./HttpClient');
 
 /**
  * HTTP client which uses a `fetch` function to issue requests. This fetch
- * function is expected to be the Web Fetch API function or an equivalent (such
- * as the function provided by the node-fetch package).
+ * function is expected to be the Web Fetch API function or an equivalent, such
+ * as the function provided by the node-fetch package (https://github.com/node-fetch/node-fetch).
  */
 class FetchHttpClient extends HttpClient {
   constructor(fetchFn) {
--- a/lib/net/FetchHttpClient.js
+++ b/lib/net/FetchHttpClient.js
@@ -44,6 +44,19 @@ class FetchHttpClient extends HttpClient {
       body: requestData || undefined,
     });
 
+    // The Fetch API does not support passing in a timeout natively, so a
+    // timeout promise is constructed to race against the fetch and preempt the
+    // request, simulating a timeout.
+    //
+    // This timeout behavior differs from Node:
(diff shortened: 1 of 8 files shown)


lib/net/FetchHttpClient.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
+
+const {HttpClient, HttpClientResponse} = require('./HttpClient');
+
+/**
+ * HTTP client which uses a `fetch` function to issue requests. This fetch
+ * function is expected to be the Web Fetch API function or an equivalent (such
+ * as the function provided by the node-fetch package).
```
richardm-stripe, line 8: Let's link it?
```suggestion
 * function is expected to be the Web Fetch API function or an equivalent, such
 * as the function provided by the node-fetch package (https://github.com/node-fetch/node-fetch).
```

lib/net/FetchHttpClient.js
```diff
@@ -0,0 +1,122 @@
...
+ */
+class FetchHttpClient extends HttpClient {
+  constructor(fetchFn) {
+    super();
+    this._fetchFn = fetchFn;
+  }
+
+  /* eslint-disable class-methods-use-this */
```
richardm-stripe, line 16: Shall we globally disable this rule? We seem to violate it all the time and I don't really agree with it.

lib/net/FetchHttpClient.js
```diff
@@ -0,0 +1,122 @@
...
+    return this._res.json();
+  }
+
+  static _transformHeadersToObject(headers) {
+    // Fetch uses a Headers instance so this must be converted to a barebones
+    // JS object to meet the HttpClient interface.
+    const headersObj = {};
+
+    for (const entry of headers) {
+      headersObj[entry[0]] = entry[1];
```
richardm-stripe, line 115: WDYT of throwing an exception here if `entry` is not an array with two entries? 
I think maybe it's worth being more defensive here because `fetchFn` is injected

test/net/helpers.js
```diff
@@ -0,0 +1,184 @@
...
+          const response = await sendRequest();
+
+          // Headers get transformed into lower case.
+          expect(response.getHeaders()).to.be.deep.equal({
+            'x-header-1': '123',
+            'x-header-2': 'test',
+          });
+        });
+
+        it('toJSON accumulates all data chunks in utf-8 encoding', async () => {
```
richardm-stripe, line 152: nice

richardm-stripe (approved):
LGTM! Left a few minor suggestions - I am so excited for this

lib/net/FetchHttpClient.js
dcr-stripe: Good call! Done.

lib/net/FetchHttpClient.js
dcr-stripe: Done. Agreed.

dcr-stripe:
Added fixes. PTAL @richardm-stripe ? Thanks for the feedback!

richardm-stripe (approved):

PR #1237, opened 2021-09-10
Add a CryptoProvider interface and NodeCryptoProvider implementation.

dcr-stripe:
### Notify

r? @richardm-stripe

### Summary

Adds an `CryptoProvider` interface and moves out the Node `crypto` logic into a `NodeCryptoProvider` implementation. 

This initially only supports computing HMAC signatures, but lets us support this in a pluggable way. The Webhooks functions are updated to support passing in a custom provider. This can be used in environments where no 'crypto' package is available (removing another Node dependency for https://github.com/stripe/stripe-node/issues/997).

### Test plan

Added a small common set of test cases which can be re-used by any implementation to ensure they meet the contract.

### Looking forwards

Some crypto APIs (eg. the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API)) only provide async functions for computing HMACs (and other primitives). In the future, we may want to extend this interface with an `asyncComputeHMACSignature` function that returns a promise alongside `computeHMACSignature`. We would need to update the Webhooks code to support both sync and async computations (eg. providing `asyncConstructEvent` and `asyncVerifyHeader`).

### DIFF
--- a/lib/Webhooks.js
+++ b/lib/Webhooks.js
@@ -1,19 +1,18 @@
 'use strict';
 
-const crypto = require('crypto');
-
 const utils = require('./utils');
 const {StripeError, StripeSignatureVerificationError} = require('./Error');
 
 const Webhook = {
   DEFAULT_TOLERANCE: 300, // 5 minutes
 
-  constructEvent(payload, header, secret, tolerance) {
+  constructEvent(payload, header, secret, tolerance, cryptoProvider) {
     this.signature.verifyHeader(
       payload,
       header,
       secret,
-      tolerance || Webhook.DEFAULT_TOLERANCE
+      tolerance || Webhook.DEFAULT_TOLERANCE,
+      cryptoProvider
     );
 
     const jsonPayload = JSON.parse(payload);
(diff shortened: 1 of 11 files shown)


lib/Webhooks.js
```diff
@@ -104,7 +99,8 @@ const signature = {
       });
     }
 
-    const expectedSignature = this._computeSignature(
+    cryptoProvider = cryptoProvider || getNodeCryptoProvider();
+    const expectedSignature = cryptoProvider.computeHMACSignature(
```
richardm-stripe, line 103: This might be evil, but for async support:
```diff
- const expectedSignature = cryptoProvider.computeHMACSignature(
-    ...
- );
- // Do a bunch of stuff with expectedSignature
+ const compare = (expectedSignature) => {
+   // Do a bunch of stuff with expectedSignature
+ };
+ const expectedSignature = cryptoProvider.computeHMACSignature(
+   ...
+ );
+ if (expectedSignature && typeof expectedSignature.then === 'function') {
+   return expectedSignature.then(compare);
+ } else {
+   return compare(expectedSignature);
+ }
```

richardm-stripe (approved):
Looks good to me!

lib/Webhooks.js
dcr-stripe: Interesting. I think this could potentially simplify the implementation over having two separate methods... let's think about it and see how the need comes up? Either change would require updating the types though I imagine changing `constructEvent`'s return type from `Stripe.Event` to `Stripe.Event | Promise<Stripe.Event>` wouldn't technically be backwards compatible? 

Think we can merge this and handle async in a separate PR.

Issue #879, opened 2020-04-27
external_account type: should be flexible to accept string or object


alisherk:
The official stripe documentation mentions external_account accepts token as string and object. Your type expects String only. It should be flexible to accept both types. 

`  const params: Stripe.AccountCreateParams = {
   type: 'custom', 
   country: data.country, 
   email: data.email, 
   requested_capabilities: ['card_payments', 'transfers'],
   business_type: data.businessType, 
   tos_acceptance: {
    date: Math.floor(Date.now() / 1000),
    ip: context.rawRequest.ip,
   },
   
   external_account: {
    object: 'bank_account',
    account_number: data.bankAccount,

  }  
  
  }`

richardm-stripe:
Hi @alisherk, thanks for the report. We're planning on tackling this soon!

alisherk:
Thank you for the great lib

dncrews:
Any chance of movement on this, @alisherk?

dncrews:
Any chance of movement on this @richardm-stripe

richardm-stripe:
Released in [v8.145.0](https://github.com/stripe/stripe-node/releases/tag/v8.145.0).

PR #1239, opened 2021-09-11
external_account can be an object

giulianok:
The documentation says that the value of `external_account` when creating a bank account can be either a token generated by the SDK (aka string) or a dictionary/object with required fields (see https://stripe.com/docs/api/external_account_bank_accounts/create?lang=node).

The TS type currently says that only `string` is supported.

## Right now
Screen Shot 2021-09-10 at 8 22 44 PM

## After change
Screen Shot 2021-09-10 at 8 22 30 PM

### DIFF
--- a/types/2020-08-27/ExternalAccounts.d.ts
+++ b/types/2020-08-27/ExternalAccounts.d.ts
@@ -6,7 +6,7 @@ declare module 'stripe' {
       /**
        * Please refer to full [documentation](https://stripe.com/docs/api) instead.
        */
-      external_account: string;
+      external_account: string | ExternalAccount;
 
       /**
        * When set to true, or if this is the first external account added in this currency, this account becomes the default external account for its currency.


remi-stripe:
@giulianok Thanks for the PR submission. This limitation is something that we do on purpose to decrease the risk that a developer defaults to sending raw card details via the API since the information is sensitive and protected by PCI compliance. While our API does support it for advanced users, it's highly discouraged and our types reflect the correct way to integrate.

This is something we've discussed a few times before such as in https://github.com/stripe/stripe-node/issues/879.

I'm going to close this PR out as we wouldn't merge this change, but I hope this clarifies the reasoning.

giulianok:
Thanks @remi-stripe,

Quick question, the reason I need to pass an object instead of a token is that I wasn't able to find a solution to generate a bank account token from the front-end (stripe-js), any idea if that's supported?

remi-stripe:
@giulianok It is supported and documented here: https://stripe.com/docs/js/tokens_sources/create_token?type=bank_account

PR #1241, opened 2021-09-13
search-beta: typescript, default version, and README

richardm-stripe:
Note: the target for this pr is `search-beta`, not master.
## Notify
r? @dcr-stripe 
cc @stripe/api-libraries 

## Summary
* Bumped the version from `0.0.2` to `8.174.0` to match the mainline library.
* beta-specific README.md.
* Changed the library so that `stripe-version:2020-08-27;search_api_beta=v1` header is sent by default on all requests.
* Fix the typescript types:
  * Changed the module name to `stripe-search-beta` - in my testing this was necessary for the Typescript types to be picked up my by example project after changing the import statement.
  * Changed the `apiVersion` type to "2020-08-27;search_api_beta=v1" -- so that users will get a Typescript error if they are not configuring their library to send that API version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,518 +1,57 @@
-# Stripe Node.js Library
+# stripe-search-beta
 
-[![Version](https://img.shields.io/npm/v/stripe.svg)](https://www.npmjs.org/package/stripe)
-[![Build Status](https://travis-ci.org/stripe/stripe-node.svg?branch=master)](https://travis-ci.org/stripe/stripe-node)
-[![Coverage Status](https://coveralls.io/repos/github/stripe/stripe-node/badge.svg)](https://coveralls.io/github/stripe/stripe-node)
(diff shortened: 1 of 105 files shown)


README.md
```diff
@@ -1,518 +1,57 @@
...
-  const customer = await stripe.customers.create({
-    email: 'customer@example.com',
-  });
-
-  console.log(customer.id);
-})();
+```diff
+-const stripe = require('stripe')("<your-api-key>")
++const stripe = require('stripe-search-beta')("<your-api-key>")
 ```
```
dcr-stripe, line 23: This looks great! Does anything need to change in their typescript files? I don't _think_ so but double checking.

dcr-stripe (approved):
LGTM - amazing! 🚢

README.md
richardm-stripe: They'll need to change the package they import from, but that should be it.

PR #1247, opened 2021-09-16
Update README.md

seantcanavan:
Link to updated code instead of the deprecated / readonly previous project.


remi-stripe (approved):
Thanks a lot for the fix @seantcanavan!

PR #1257, opened 2021-09-24
API Updates

richardm-stripe:
Codegen for openapi 8442a02.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `amount_authorized` and `overcapture_supported` on `Charge.payment_method_details.card_present`

### DIFF
--- a/test/resources/GeneratedTests.spec.js
+++ b/test/resources/GeneratedTests.spec.js
@@ -768,8 +768,8 @@ describe('TaxId', function() {
 
   it('retrieveTaxId method', async function() {
     const taxId = await stripe.customers.retrieveTaxId(
-      'txi_xxxxxxxxxxxxx',
-      'cus_xxxxxxxxxxxxx'
+      'cus_xxxxxxxxxxxxx',
+      'txi_xxxxxxxxxxxxx'
     );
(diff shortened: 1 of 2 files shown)


test/resources/GeneratedTests.spec.js
```diff
@@ -768,8 +768,8 @@ describe('TaxId', function() {
 
   it('retrieveTaxId method', async function() {
     const taxId = await stripe.customers.retrieveTaxId(
-      'txi_xxxxxxxxxxxxx',
-      'cus_xxxxxxxxxxxxx'
+      'cus_xxxxxxxxxxxxx',
+      'txi_xxxxxxxxxxxxx'
```
dcr-stripe, line 772: Hmmmmm....... maybe we have something non-deterministic here? 🤔  Seems unrelated to this...

dcr-stripe (approved):
LGTM!

test/resources/GeneratedTests.spec.js
richardm-stripe: Not non-determinism -- this is the result of a minor bugfix: the correct ordering is [parentId, childId] and this had been called with the arguments in the wrong order.

PR #1288, opened 2021-11-09
Add SubtleCryptoProvider and update Webhooks to allow async crypto.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Introduce a new `CryptoProvider` which uses the [`SubtleCrypto` API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/). This is the main implementation for the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API).

This requires adding support for async HMAC computations in `CryptoProvider` and adding new async Webhook creation/verification functions which can use these.

As a result, we can now construct and verify a webhook asynchronously by doing:

```
const cryptoProvider = Stripe.createSubtleCryptoProvider();

const event = await stripe.webhooks.constructEventAsync(
  body, 
  signature, 
  secret,
  tolerance, 
  cryptoProvider
);
```

I opted to go down the route of having sync and async methods rather than parameterizing the CryptoProvider. We could have done something like `CryptoProvider<T>` and then `constructEvent():T`. However our internal code would still need to check whether the `T` was a promise or not. Instead this forces one to be explicit about whether you want to operate synchronously or asynchronously, and gives us an opportunity to error if trying to mix contexts.

### Motivation

We need this for Deno support (https://github.com/stripe/stripe-node/issues/997). Deno's `crypto` package doesn't implement HMAC and instead relies on the Web Crypto API, which is all async.

### Test Plan

Added tests throughout. Note that unfortunately the `SubtleCryptoProvider` can only be tested on Node 16, as Node only added a `SubtleCrypto` shim in Node 15. All tests are shared between sync and async versinos.

I've also set up a Deno template which processes webhooks asynchronously using this

### DIFF
--- a/test/crypto/SubtleCryptoProvider.spec.js
+++ b/test/crypto/SubtleCryptoProvider.spec.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const SubtleCryptoProvider = require('../../lib/crypto/SubtleCryptoProvider');
+const webcrypto = require('crypto').webcrypto;
+const expect = require('chai').expect;
+
+// webcrypto is only available on Node 15+.
+if (!webcrypto || !webcrypto.subtle) {
+  console.log(
+    `Skipping SubtleCryptoProvider tests. No 'webcrypto.subtle' available for ${process.version}.`
+  );
+  return;
+}
+
+const {createCryptoProviderTestSuite} = require('./helpers');
+
+describe('SubtleCryptoProvider', () => {
+  const provider = new SubtleCryptoProvider(webcrypto.subtle);
+
+  createCryptoProviderTestSuite(provider, true);
+
+  it('does not support sync calls', () => {
+    expect(() => {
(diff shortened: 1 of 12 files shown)


test/crypto/SubtleCryptoProvider.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  );
+  return;
+}
+
+const {createCryptoProviderTestSuite} = require('./helpers');
+
+describe('SubtleCryptoProvider', () => {
+  const provider = new SubtleCryptoProvider(webcrypto.subtle);
+
+  createCryptoProviderTestSuite(provider, true);
```
richardm-stripe, line 20: nit: I try to avoid calling functions that call `describe` and `it` since this interacts poorly with test runners, but nbd

richardm-stripe (approved):
LGTM! I like the choice of completely parallel async and sync pathways.

PR #1291, opened 2021-11-11
Improve typescript support

jnraine-stripe:
## Notify

r? @richardm-stripe 

## Summary

This improves three friction points when working with Typescript:

1. **`new StripeResource(stripe)` now works.** Prior to this change, attempting to construct `StripeResource` with its one required argument would fail a type check. Support for a constructor was added in #1245 and extended in #1249 (never merged). Thanks @rattrayalex!
2. **`StripeResource.extend` now returns a class that is properly typed when instantiated.** Prior to this change, you could extend `StripeResource` but your extensions wouldn't appear on the object after construction. Now, it does.
3. **`StripeResource.method` returns a function that returns `Response<object>`.** Prior to this change, extensions created with `StripeResource.method` would have a return type of `object`. Returning `Response<object>` makes it possible to call functions like `lastResponse` without casting the value.
4. **`StripeResource.method` now accepts generic type that sets the response object's type.** For example, if you're hitting a customer endpoint, you can do this: `StripeResource.method<Customer>(...)`. When you use that extension, it will return `Response<Customer>`. This eliminates the need to cast when calling StripeResource extension methods.

See changes to `typescriptTest.ts` for these updates in action.

---

fixes #1249

### DIFF
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -5,6 +5,13 @@ import {Agent} from 'http';
 
 declare module 'stripe' {
   namespace Stripe {
+    type StripeResourceClass = typeof StripeResource;
+
+    interface StripeResourceExtension<T extends object>
+      extends StripeResourceClass {
+      new (stripe: Stripe): StripeResource & T;
+    }
+
     export class StripeResource {
       static extend<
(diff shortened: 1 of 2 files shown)


richardm-stripe:
This looks good to me!

PR #1292, opened 2021-11-11
API Updates

richardm-stripe:
Codegen for openapi 21065d4.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `expire` method on resource `Checkout.Session`
* Add support for `status` on `Checkout.Session`


dcr-stripe (approved):
LGTM!

PR #1312, opened 2021-12-17
API Updates

dcr-stripe:
Codegen for openapi 26c0399.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `processing` on `PaymentIntent`

PR #1314, opened 2021-12-22
API Updates

dcr-stripe:
Codegen for openapi 067bdc5.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `au_becs_debit` on `PaymentIntentCreateParams.payment_method_options`, `PaymentIntentUpdateParams.payment_method_options`, `PaymentIntentConfirmParams.payment_method_options`, and `PaymentIntent.payment_method_options`
* Change type of `PaymentIntent.processing.type` from `string` to `literal('card')`. This is not considered a breaking change as the field was added in the same release.

### DIFF
--- a/types/2020-08-27/PaymentIntents.d.ts
+++ b/types/2020-08-27/PaymentIntents.d.ts
@@ -516,6 +516,8 @@ declare module 'stripe' {
 
         alipay?: PaymentMethodOptions.Alipay;
 
+        au_becs_debit?: PaymentMethodOptions.AuBecsDebit;
+
         bancontact?: PaymentMethodOptions.Bancontact;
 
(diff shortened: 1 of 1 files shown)


types/2020-08-27/PaymentIntents.d.ts
```diff
@@ -787,7 +791,7 @@ declare module 'stripe' {
         /**
          * Type of the payment method for which payment is in `processing` state, one of `card`.
          */
-        type: string;
+        type: 'card';
```
dcr-stripe, line 794: This change conversion from string to enum is fine since we haven't actually released https://github.com/stripe/stripe-node/pull/1312 yet.

yejia-stripe (approved):

PR #1333, opened 2022-01-20
Customer tax_ids is not included by default

agix:
According to the documentation https://stripe.com/docs/api/customers/object#customer_object-tax_ids

> This field is not included by default.

Typing suppose it is.

Doing `stripe.customers.retrieve` show it is indeed not included by default.


dcr-stripe (approved):
Looks good to me! Great catch @agix . Thank you for the contribution - it's really appreciated.

PR #1334, opened 2022-01-21
refactor(http-client): dry duplicated code

bpinto:
I'm not sure if the following change is desired, but I've noticed this while working on something else and decided to open a PR to discuss this separately.

### DIFF
--- a/lib/net/FetchHttpClient.js
+++ b/lib/net/FetchHttpClient.js
@@ -88,10 +88,6 @@ class FetchHttpClientResponse extends HttpClientResponse {
     this._res = res;
   }
 
-  getRawResponse() {
-    return this._res;
-  }
-
   toStream(streamCompleteCallback) {
     // Unfortunately `fetch` does not have event handlers for when the stream is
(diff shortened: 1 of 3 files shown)


dcr-stripe:
Hi @bpinto - thanks for the contribution here! 

This assumes that `this._res` always exists but the base `HttpClientResponse` class doesn't have this variable. If we wanted to do this we would have to pass in the raw response to the constructor of `HttpClientResponse` and store it there. I think it makes sense to leave the `res` object within the sub-classes as they will be interacting with the raw response anyways when implementing `toJSON` and `toStream`. Really `HttpClientResponse` is just meant to be a pure interface/contract with no implementation.

I'll be closing this given that but please feel free to re-open if you have any questions or comments!

bpinto:
I see what you are saying, the other two methods are receiving the method via the constructor. 👍 

I was assuming that nothing was using the base `HttpClientResponse` class directly and only using the sub-classes directly. In any case, I can see the point of leaving this to the sub-classes to implement, even if the code is the same.

PR #1335, opened 2022-01-21
fix(http-client): use named import

bpinto:
Custom request timeout message was never used since we were incorrectly
requiring the HttpClient class.

I believe this has stopped working on #1218.


dcr-stripe (approved):
Great catch - thank you for the contribution @bpinto ! Merging in and this will be included in the next release.

Issue #1040, opened 2020-10-13
Intermittent Error: write EPIPE when running stripe client in AWS Lambda


hisham:
We're using the stripe node client 8.71.0 on an AWS Lambda running node 12.x. A stripe customers.list call is called first thing when the lambda executes. 33% of the time - we get this error on that call. It consistently happens so does not seem to be transient. 

I did read https://github.com/stripe/stripe-node/issues/650, and setting maxNetworkRetries in stripe to 2 seems to resolve the issue. However it seems that just masks the issue. 

Is this a stripe issue or AWS Lambda issue? Probably lambda, I submitted a request with AWS. But putting this here in case others run into it.  

2020-10-13T12:02:58.032Z	c184006d-fe96-490a-9bfe-696b8271769a	ERROR	StripeConnectionError: An error occurred with our connection to Stripe.
    at /var/task/node_modules/stripe/lib/StripeResource.js:234:9
    at ClientRequest.<anonymous> (/var/task/node_modules/stripe/lib/StripeResource.js:489:67)
    at ClientRequest.emit (events.js:315:20)
    at ClientRequest.EventEmitter.emit (domain.js:483:12)
    at TLSSocket.socketErrorListener (_http_client.js:426:9)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket.EventEmitter.emit (domain.js:483:12)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  type: 'StripeConnectionError',
  raw: {
    message: 'An error occurred with our connection to Stripe.',
    detail: Error: write EPIPE
        at WriteWrap.onWriteComplete [as oncomplete] (internal/stream_base_commons.js:92:16)
        at writevGeneric (internal/stream_base_commons.js:132:26)
        at TLSSocket.Socket._writeGeneric (net.js:784:11)
        at TLSSocket.Socket._writev (net.js:793:8)
        at doWrite (_stream_writable.js:401:12)
        at clearBuffer (_stream_writable.js:519:5)
        at TLSSocket.Writable.uncork (_stream_writable.js:338:7)
        at ClientRequest.end (_http_outgoing.js:774:17)
        at ClientRequest.<anonymous> (/var/task/node_modules/stripe/lib/StripeResource.js:506:15)
        at Object.onceWrapper (events.js:422:26) {
[...]

PR #1336, opened 2022-01-21
feat(http-client): retry closed connection errors

bpinto:
## Changes made

Requests that fail with closed connection errors (`ECONNRESET`, `EPIPE`) are now automatically retried.

- `ECONNRESET` (Connection reset by peer): A connection was forcibly
  closed by a peer.closed by a peer. This normally results from a loss
  of the connection on the remote socket due to a timeout or reboot.
  Commonly encountered via the http and net modules.

- `EPIPE` (Broken pipe): A write on a pipe, socket, or FIFO for which
  there is no process to read the data. Commonly encountered at the net
  and http layers, indicative that the remote side of the stream being
  written to has been closed.

## Purpose of changes

Fixes: #1040

> From my perspective, handling these errors by retrying is likely the proper approach, and shouldn't necessarily be viewed as a workaround, or masking an underlying issue, because it is expected/unavoidable that these broken connections will come to exist, and there doesn't seem to be an obvious way of asking Node "how long has it been since the last keep-alive probe on this connection" besides writing to the connection and triggering the error.

> At the same time, I think we should look into the possibility of making stripe-node handle errors like this by default/more transparently, so that users don't have to configure the retries themselves. That seems to be what Amazon started doing for errors like this for their own SDKs about a month ago (thank you @hisham for linking to that issue, by the way).

See: https://github.com/stripe/stripe-node/issues/1040#issuecomment-748405623

Depends on #1335

## TODO

- [ ] Should this become optional through a config option?
- [ ] Update documentation.
- [ ] Test this change on a lambda environment.


bpinto:
Hey @richardm-stripe 👋 

I'm excited about this change as well! Please let me know what you think we could do differently.

richardm-stripe:
Hi @bpinto, just to keep you updated, we are planning on merging this change in the next major as we're a bit hesitant to change the default retry behavior in a minor.

bpinto:
Totally understandable, thanks for the follow up.

richardm-stripe (approved):

PR #1344, opened 2022-02-03
API Updates

yejia-stripe:
Codegen for openapi 10fe9f5.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new value `au_becs_debit` on enum `CheckoutSessionCreateParams.payment_method_types[]`
* Change type of `Refund.reason` from `string` to `enum('duplicate'|'expired_uncaptured_charge'|'fraudulent'|'requested_by_customer')`


types/2020-08-27/Checkout/Sessions.d.ts
```diff
@@ -1619,6 +1615,7 @@ declare module 'stripe' {
           | 'acss_debit'
           | 'afterpay_clearpay'
           | 'alipay'
+          | 'au_becs_debit'
```
dcr-stripe, line 1618: Let's do the other languages as well for this change - looks like we missed it!

dcr-stripe (approved):
LGTM - thanks Yejia!

PR #1358, opened 2022-02-25
Update request reasons

kamesstory:
Include `authorization_controls` as a Reason, which is returned in legacy API versions.


richardm-stripe:
Hi @kamesstory, unfortunately this file is code-generated from our API Definition, which only reflects the fields contained in latest, official version of the Stripe API. If we merged this pull request your changes would be wiped the next time we generated the library.

I'm sorry that this is the answer - it totally makes sense that you would want a library that reflects the Stripe API that you actually experience. This problem is top of mind for us, but we don't have a solution for now other than to recommend you bypass typescript in this instance.

PR #1364, opened 2022-03-04
Update search pagination to use page param instead of next_page.

dcr-stripe:
r? @richardm-stripe 

## Summary

Updates the search pagination behavior to account for two changes:

1. The `next_page` request parameter has been renamed to `page`.
2. The `next_page` response property is always provided, but will return null when there is no page. Previously it would not be provided in the return object at all.

This is not considered a breaking change as no currently released API uses this object and is thus purely an infrastructural change.

### DIFF
--- a/test/autoPagination.spec.js
+++ b/test/autoPagination.spec.js
@@ -574,6 +574,25 @@ describe('auto pagination', function() {
     });
   });
 
+  const testCase = (
+    mockPaginationFn,
+    {pages, limit, expectedIds, expectedParamsLog, initialArgs}
+  ) => {
+    const {paginator, paramsLog} = mockPaginationFn(pages, initialArgs);
+
+    return expect(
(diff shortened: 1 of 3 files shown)


test/autoPagination.spec.js
```diff
@@ -708,12 +708,11 @@ describe('auto pagination', function() {
...
-        let nextPageProperties = {};
-        if (props.has_more) {
-          nextPageProperties = {
-            next_page: `${props.data[props.data.length - 1]}-encoded`,
-          };
-        }
+        const nextPageProperties = {
+          next_page: props.has_more
+            ? `${props.data[props.data.length - 1]}-encoded`
+            : nul,
```
richardm-stripe, line 714:
```suggestion
            : null,
```

dcr-stripe (commented):
PTAL @richardm-stripe now that the testing is fixed.

richardm-stripe (approved):

PR #1366, opened 2022-03-09
API Updates

dcr-stripe:
Codegen for openapi d65ebba.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `test_clock` on `CustomerListParams`
* Change `Invoice.test_clock`, `InvoiceItem.test_clock`, `Quote.test_clock`, `Subscription.test_clock`, and `SubscriptionSchedule.test_clock` to be required

### DIFF
--- a/types/2020-08-27/Customers.d.ts
+++ b/types/2020-08-27/Customers.d.ts
@@ -671,6 +671,11 @@ declare module 'stripe' {
        * Specifies which fields in the response should be expanded.
        */
       expand?: Array<string>;
+
+      /**
+       * Provides a list of customers that are associated with the specified test clock. If no list is provided, the response won't include customers with test clocks.
+       */
+      test_clock?: string;
     }
 
     interface CustomerDeleteParams {}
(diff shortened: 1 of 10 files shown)


types/2020-08-27/Customers.d.ts
```diff
@@ -671,6 +671,11 @@ declare module 'stripe' {
        * Specifies which fields in the response should be expanded.
        */
       expand?: Array<string>;
+
+      /**
+       * Provides a list of customers that are associated with the specified test clock. If no list is provided, the response won't include customers with test clocks.
```
kamil-stripe, line 676: Is that doc correct?

richardm-stripe (approved):
LGTM aside from docstring

types/2020-08-27/Customers.d.ts
dcr-stripe: Fixed - good catch!

richardm-stripe (approved):

PR #1370, opened 2022-03-15
if no content-length, do not add the header

CraigglesO:
Fixes errors thrown by Cloudflare. If content-length is 0, the request will error out.

### DIFF
--- a/lib/StripeResource.js
+++ b/lib/StripeResource.js
@@ -369,7 +369,6 @@ StripeResource.prototype = {
       Authorization: auth ? `Bearer ${auth}` : this._stripe.getApiField('auth'),
       Accept: 'application/json',
       'Content-Type': 'application/x-www-form-urlencoded',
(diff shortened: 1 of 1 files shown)


dcr-stripe:
Thanks for the PR @CraigglesO 

I think we'd actually want to do this on a per-method basis.

As per https://tools.ietf.org/html/rfc7230#section-3.3.2,

> A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.
   
 We shouldn't be setting this header for verbs that don't anticipate a body.
 
 However, for POST, we should keep sending Content-Length even if it is 0.
 
 > Any Content-Length field value greater than or equal to zero is
   valid

dcr-stripe:
I put together https://github.com/stripe/stripe-node/pull/1388 to do this on a per-method basis.

Issue #1374, opened 2022-03-19
StripeError is undefined


vinnymac:
## Summary

The issue is that when developing with `stripe-node` and typescript, it appears to the developer and to the typescript compiler, that `Stripe.StripeError` will exist at runtime. However, in reality this is `undefined` at runtime. I've confirmed this in a reproduction below. I ran into this issue recently when following the pattern recommended in the wiki for error handling with Typescript [here](https://github.com/stripe/stripe-node/wiki/Error-Handling#typescript). That link indicates that `err instanceof Stripe.StripeError` is a valid way to test that your error is in fact an instance of a Stripe Error. Instead you get a runtime error that says 

```
TypeError: Right-hand side of 'instanceof' is not an object
```

I recommend the types be updated so they indicate that Stripe.StripeErrors does not exist, and the wiki gets updated to recommend using `Stripe.errors.StripeError`, but for all I know this should exist, when it does not 🤷🏼‍♂️.

## Workaround
Just use `err instanceof Stripe.errors.StripeError` until this is fixed.

## Reproduction
https://codesandbox.io/s/stripeerror-is-undefined-kdzkmh?file=/src/index.ts

## Screenshot of issue in case of stale reproduction
<img width="1029" alt="Screen Shot 2022-03-19 at 1 45 00 PM" src="https://user-images.githubusercontent.com/1832781/159132462-455450da-d1dd-414d-afeb-4d5eecb064ec.png">

## Versions
I tested this with the following software

v8.129.0 of `stripe-node`
v10.19.1 of NodeJS
v4.5.0 of Typescript

as well as

v8.210.0 of `stripe-node`
v10.20.1 of NodeJS
v4.4.4 of Typescript

pakrym-stripe:
@vinnymac, for backwards compatibility reasons we can't remove `Stripe.StripeError` and other error definitions. We adjusted the namespaces and deprecated `Stripe.Stripe**Error` classes.

PR #1375, opened 2022-03-21
Update error types to be namespaced under Stripe.error

pakrym-stripe:
Fixes: https://github.com/stripe/stripe-node/issues/1374

This fixes a mismatch between the types and the code. All errors classes are exported through Stripe.errors and are not accessible at the top-level (eg. Stripe.StripeError is not valid - it is exported as Stripe.errors.StripeError). The types however showed these classes at the top-level. This PR brings the types in line with the underlying code while still offering the top-level types with a deprecation notice to avoid breaking existing Typescript implementations.

Looks like this in the IDE:

image

### DIFF
--- a/types/2020-08-27/index.d.ts
+++ b/types/2020-08-27/index.d.ts
@@ -217,8 +217,7 @@ declare module 'stripe' {
     /**
      * API Errors
      */
-    static errors: Stripe.Errors;
-    errors: Stripe.Errors;
+    errors: typeof Stripe.errors;
 
     on(event: 'request', handler: (event: Stripe.RequestEvent) => void): void;
     on(event: 'response', handler: (event: Stripe.ResponseEvent) => void): void;
--- a/types/2020-08-27/index.d.ts
+++ b/types/2020-08-27/index.d.ts
@@ -217,8 +217,8 @@ declare module 'stripe' {
     /**
      * API Errors
      */
-    static errors: Stripe.Errors;
-    errors: Stripe.Errors;
+    static errors: Stripe.errors.Errors;
+    errors: Stripe.errors.Errors;
 
     on(event: 'request', handler: (event: Stripe.RequestEvent) => void): void;
     on(event: 'response', handler: (event: Stripe.ResponseEvent) => void): void;
--- a/types/2020-08-27/index.d.ts
+++ b/types/2020-08-27/index.d.ts
@@ -217,6 +217,7 @@ declare module 'stripe' {
     /**
      * API Errors
      */
+    static errors: Stripe.Errors;
     errors: Stripe.Errors;
 
     on(event: 'request', handler: (event: Stripe.RequestEvent) => void): void;
--- a/types/2020-08-27/index.d.ts
+++ b/types/2020-08-27/index.d.ts
@@ -217,8 +217,7 @@ declare module 'stripe' {
     /**
      * API Errors
      */
-    static errors: Stripe.errors.Errors;
-    errors: Stripe.errors.Errors;
+    errors: Stripe.Errors;
 
     on(event: 'request', handler: (event: Stripe.RequestEvent) => void): void;
     on(event: 'response', handler: (event: Stripe.ResponseEvent) => void): void;
--- a/types/Errors.d.ts
+++ b/types/Errors.d.ts
@@ -1,18 +1,21 @@
 declare module 'stripe' {
   namespace Stripe {
+    /**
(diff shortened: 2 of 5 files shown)


richardm-stripe (changes requested):
Can you add some more detail to the pull request description?

Will this break type references to `Stripe.Errors.*`? If so, I think the blast radius of this is too big to release outside a major version, and maybe we should consider copying either the runtime type or the Typescript type to both places? Does that make sense?

pakrym-stripe:
I updated the PR so old types are still accessible but are marked as deprecated and don't work as a parameter to instanceof call.

dcr-stripe (commented):
This looks good Pavel!

For the sake of communication, I would:
1. Change PR title to be something like "Update error types to be namespaced under Stripe.error"
2. Update the PR description to make it clear that this is not a breaking change, and describing the current underlying implementation. Could be something like

> This fixes a mismatch between the types and the code. All errors classes are exported through `Stripe.errors` and are not accessible at the top-level (eg. `Stripe.StripeError` is not valid - it is exported as `Stripe.errors.StripeError`). The types however showed these classes at the top-level. This PR brings the types in line with the underlying code while still offering the top-level types with a deprecation notice to avoid breaking existing Typescript implementations.

3. On next release, let's make sure to include a similar sentence directly in the CHANGELOG so that users are aware of this.
4. Maybe we could add a comment in the types explaining that the deprecated type doesn't actually exist + why we're keeping it?

--

re. the instanceof support - I wonder if we want to keep that but also in a deprecated state?  It would be nice to not break existing compilations for either reason.  In an ideal world we could just re-export everything at the top-level to make the JS match the types. I also wonder if we used to do that and then something accidentally changed in the JS?

pakrym-stripe:
Done, duplicated types so instanceof still compiles added deprecated description to everything.

types/Errors.d.ts
```diff
@@ -1,5 +1,8 @@
 declare module 'stripe' {
   namespace Stripe {
+    /**
+     * @deprecated The {@link Stripe.Errors} type is deprecated, please use {@link Stripe.errors} namespace directly.}
```
pakrym-stripe, line 4: This deprecation only matters for customers that for some reason used Stripe.Errors type directly, which I expect to be very rare.

dcr-stripe, line 4:
```suggestion
     * @deprecated The {@link Stripe.Errors} type is deprecated, please use the {@link Stripe.errors} namespace directly.}
```

types/Errors.d.ts
```diff
@@ -46,6 +49,10 @@ declare module 'stripe' {
       source?: Stripe.Source;
     };
 
+    /**
+     * @deprecated The {@link Stripe.StripeError} type is deprecated, please use {@link Stripe.errors.StripeError}.}
```
dcr-stripe, line 53: This and most other @deprecated messages have a trailing `}` - let's remove.

eg.

```
     * @deprecated The {@link Stripe.StripeError} type is deprecated, please use {@link Stripe.errors.StripeError}.}
```

should be

```
     * @deprecated The {@link Stripe.StripeError} type is deprecated, please use {@link Stripe.errors.StripeError}.
```

dcr-stripe (commented):
Some small tweaks and then I think we're good to go! We should also do a pass internally on documentation to make sure we're suggesting `Stripe.errors.*` everywhere.

PTAL @pakrym-stripe ?

pakrym-stripe:
PTAL @dcr-stripe ?

types/2020-08-27/index.d.ts
```diff
@@ -217,8 +217,7 @@ declare module 'stripe' {
     /**
      * API Errors
      */
-    static errors: Stripe.Errors;
```
dcr-stripe, line 220: I guess we can't have this static anymore now that we have the `errors` namespace right?

dcr-stripe (approved):
LGTM! Thanks for your work on this Pavel!

Just one thing - let's remove:
>Also makes it so *** instanceof Stripe.StripeError; doesn't compile anymore but it seems OK because it never actually worked at runtime.

From the PR description, since now it does compile.

types/2020-08-27/index.d.ts
pakrym-stripe: It's not that we can't it's that we don't need to. 

`static errors: typeof Stripe.errors` is equivalent to not having a static and `Stripe.errors` handled by namespace.

dcr-stripe: Yep totally makes sense.

PR #1379, opened 2022-03-23
Update outdated packages

kamil-stripe:
* Prettier update brings the most changes so please check "ignore whitespace" option.


.github/workflows/main.yml
```diff
@@ -13,17 +11,50 @@ on:
...
+    strategy:
+      matrix:
+        os:
+          - "ubuntu-latest"
+        node:
+          - "16"
+          - "14"
+          - "12"
+          - "10"
+          - "8"
```
dcr-stripe, line 50: I don't think we need Node 8 here - it wasn't previously in our CI matrix. But if it passes then 🤷  can't hurt.

test/resources/Subscriptions.spec.js
```diff
@@ -126,7 +126,7 @@ describe('subscriptions Resource', () => {
     it('Sends the correct request', () => {
       stripe.subscriptions.update('test_sub', {
         items: {
-          '0': {
+          0: {
```
dcr-stripe, line 129: This seems like an actual functionality change - `'0'` isn't the same as `0`. Let's double check what the API returns here and fix this to match that. If it's actual number indices, should this be an array?

.eslintrc.json
```diff
@@ -0,0 +1,253 @@
+{
```
dcr-stripe, line 1: Can we tag this file and the other eslint as a `mv`? That way we can preserve the git history for this file.

dcr-stripe (commented):
Thanks Kamil for driving this! 

Some thoughts:

1. Can you expand on "Split CI into build/lint and test stage. Some of the updated packages dropped support for Node 8 and 10 and there is no need to see if it builds there, just if it works fine."? This was a bit unclear to me since it looks like we're still doing that for all Node versions and more importantly we do need to support Node 10, so we should make sure build is working in that environment.
2. Let's update the PR description to say "Update dev dependencies and CI template" to make it clear what we're upgrading.
3. Let's split out the `Subscriptions.spec.js` fix into a separate PR if it turns out we had an issue there. I want to make sure that gets fixed if we have to revert/change some of this PR and it's isolated from the rest of this PR anyways.

PTAL @kamil-stripe

kamil-stripe:
> 1. Can you expand on "Split CI into build/lint and test stage. Some of the updated packages dropped support for Node 8 and 10 and there is no need to see if it builds there, just if it works fine."? This was a bit unclear to me since it looks like we're still doing that for all Node versions and more importantly we do need to support Node 10, so we should make sure build is working in that environment.

We build on the latest node and then test on LTS Node 8 to 14. I added it because we claim to support it:
> Requirements
> Node 8, 10 or higher

.github/workflows/main.yml
kamil-stripe: Copying it here:
I added it because we claim to support it:

> Requirements
> Node 8, 10 or higher

We can drop it from the requirements if you prefer this route.

dcr-stripe: Whoops yes, great call. I don't know how I missed that 🤦

test/resources/Subscriptions.spec.js
kamil-stripe: Reverted changes to this file.

Issue #1360, opened 2022-02-26
Deletion commands do not work via fetch


CraigglesO:
NodeJS version: v16.13.2
Stripe version: ^8.197.0

I am using a cloudflare worker, so if you check the webpack config [here](https://github.com/stripe-samples/stripe-node-cloudflare-worker-template) you can see it is limiting stripe to fetch and building like so:

```js
const STRIPE = Stripe(STRIPE_API_CODE, { httpClient: Stripe.createFetchHttpClient() })
```

commands like `await STRIPE.customers.deleteTaxId` and `await STRIPE.customers.del` do not work. An example error is as follows:

```
Zo [Error]: An error occurred with our connection to Stripe.
    at /Users/ct/Documents/Projects/node_modules/.pnpm/stripe@8.202.0/node_modules/stripe/lib/StripeResource.js:519:15
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  type: 'Zo',
  raw: {
    message: 'An error occurred with our connection to Stripe.',
    detail: TypeError: fetch failed
        at Object.processResponse (/Users/ct/.npm/_npx/dce38835dbc726fc/node_modules/undici/lib/fetch/index.js:183:23)
        at Fetch.fetchFinale (/Users/ct/.npm/_npx/dce38835dbc726fc/node_modules/undici/lib/fetch/index.js:937:17)
        at Fetch.mainFetch (/Users/ct/.npm/_npx/dce38835dbc726fc/node_modules/undici/lib/fetch/index.js:744:17)
        at processTicksAndRejections (node:internal/process/task_queues:96:5) {
      cause: [RequestContentLengthMismatchError]
    }
  },
  rawType: undefined,
  code: undefined,
  doc_url: undefined,
  param: undefined,
  detail: TypeError: fetch failed
      at Object.processResponse (/Users/ct/.npm/_npx/dce38835dbc726fc/node_modules/undici/lib/fetch/index.js:183:23)
      at Fetch.fetchFinale (/Users/ct/.npm/_npx/dce38835dbc726fc/node_modules/undici/lib/fetch/index.js:937:17)
      at Fetch.mainFetch (/Users/ct/.npm/_npx/dce38835dbc726fc/node_modules/undici/lib/fetch/index.js:744:17)
      at processTicksAndRejections (node:internal/process/task_queues:96:5) {
    cause: RequestContentLengthMismatchError: Request body length does not match content-length header
[...]

PR #1388, opened 2022-03-30
Stop sending Content-Length header for verbs which don't have bodies.

dcr-stripe:
r? @pakrym-stripe 

## Summary

Stops sending a `Content-Length` header for requests which:
1. Use a verb which doesn't anticipate a body (eg. `GET`, `DELETE`) and
2. Actually has no body.

For requests that don't anticipate a body but end up having one, we'll continue setting the header and emit a warning so that we can track and fix these cases as they are not following proper semantics.

`POST`, `PUT` and `PATCH` request are unchanged.

## Motivation

Fixes https://github.com/stripe/stripe-node/issues/1360

### DIFF
--- a/test/StripeResource.spec.js
+++ b/test/StripeResource.spec.js
@@ -75,7 +75,13 @@ describe('StripeResource', () => {
           data,
         };
 
-        const scope = nock(`https://${options.host}`)
+        const scope = nock(
+          `https://${options.host}`,
+          // No Content-Length should be present for GET requests.
+          {
+            badheaders: ['Content-Length'],
+          }
+        )
           .get(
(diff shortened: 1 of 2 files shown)


test/StripeResource.spec.js
```diff
@@ -75,7 +75,13 @@ describe('StripeResource', () => {
           data,
         };
 
-        const scope = nock(`https://${options.host}`)
+        const scope = nock(
+          `https://${options.host}`,
+          // No Content-Length should be present for GET requests.
+          {
+            badheaders: ['Content-Length'],
```
pakrym-stripe, line 82: TIL 👍

pakrym-stripe (approved):
Beautiful!

dcr-stripe:
r? @yejia-stripe for calibration

Thanks!

PR #1398, opened 2022-04-21
API Updates

richardm-stripe:
Codegen for openapi e07f049.
r? @pakrym-stripe (test_helpers expertise)
cc @stripe/api-libraries

## Changelog
* Add support for `expire` test helper method on resource `Refund`
* Change type of `BillingPortal.Configuration.application` from `string` to `expandable($Application)`
* Change `IssuingDisputeCreateParams.transaction` to be optional

### DIFF
--- a/types/2020-08-27/BillingPortal/Configurations.d.ts
+++ b/types/2020-08-27/BillingPortal/Configurations.d.ts
@@ -25,7 +25,7 @@ declare module 'stripe' {
         /**
          * ID of the Connect Application that created the configuration.
          */
-        application: string | null;
+        application: string | Stripe.Application | null;
 
         business_profile: Configuration.BusinessProfile;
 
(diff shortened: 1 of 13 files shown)


types/2020-08-27/BillingPortal/Configurations.d.ts
```diff
@@ -25,7 +25,7 @@ declare module 'stripe' {
         /**
          * ID of the Connect Application that created the configuration.
          */
-        application: string | null;
+        application: string | Stripe.Application | null;
```
pakrym-stripe, line 28: This is technically breaking? Are we allowing this in TS?

richardm-stripe: Yes, that is the present policy.

pakrym-stripe (approved):

PR #1400, opened 2022-04-25
Split CI into two stages

kamil-stripe:
Split CI into build stage that builds and lints on the latest Node.js and test stage that tests on each Node.js version that is supported by us.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,12 +76,12 @@ jobs:
       - name: Start stripe-mock
         run: docker run -d -p 12111-12112:12111-12112 stripe/stripe-mock && sleep 5
 
-      - name: Install dependencies (8)
-        if: ${{ matrix.node_version == "8" }}
+      - name: Install dependencies (Node.js 8)
+        if: ${{ matrix.node_version == '8' }}
         run: yarn install --ignore-engines
 
       - name: Install dependencies
-        if: ${{ matrix.node_version != "8" }}
+        if: ${{ matrix.node_version != '8' }}
         run: yarn install
 
       - name: Test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,3 @@
-# This is a basic workflow to help you get started with Actions
-
 name: CI
 
 on:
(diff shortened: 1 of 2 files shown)


pakrym-stripe:
Related: https://github.com/stripe/stripe-node/pull/1399

.github/workflows/main.yml
```diff
@@ -13,17 +11,50 @@ on:
...
 
     steps:
       - uses: actions/checkout@v2
 
       - name: Setup node
         uses: actions/setup-node@v2
+
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
```
pakrym-stripe, line 23: Should we be sharing caches between test and build configurations?

.github/workflows/main.yml
```diff
@@ -13,17 +11,50 @@ on:
...
+      - uses: actions/cache@v2
+        id: yarn-cache
         with:
-          node-version: ${{ matrix.node }}
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Node check
```
pakrym-stripe, line 33: What does this check for?

pakrym-stripe:
I'm a bit surprised we can still run on node 8. Are we sure we pick up the right node executable in tests?

pakrym-stripe (approved):

.github/workflows/main.yml
kamil-stripe: If JavaScript is valid. This can be removed as lint will catch it anyway.

.github/workflows/main.yml
```diff
@@ -13,17 +11,50 @@ on:
...
+        run: find . -name "*.js" -type f -not -path "./node_modules/*" -not -path "./\.*" -exec node --check {} \;
+
+      - name: Lint
```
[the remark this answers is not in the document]
kamil-stripe: I'd be in favor of that.

.github/workflows/main.yml
kamil-stripe: I don't feel super strongly about it. I wanted to use same packages in both. Not sure how much does it matter with .lock file.

kamil-stripe: It doesn't cache node_modules, just the global cache so we save on not downloading packages from npm.

.github/workflows/main.yml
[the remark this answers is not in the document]
pakrym-stripe: Was there a reason you decided not to switch to yarn jest?

.github/workflows/main.yml
```diff
@@ -13,10 +11,45 @@ on:
...
+        run: yarn && yarn lint
```
pakrym-stripe, line 37: Does this actually build?

PR #1401, opened 2022-04-25
Upgrade outdated packages

kamil-stripe:
Moved ESLint configs from .js to .json to remove warnings regarding "module.exports" bit. Used `git mv` which was ignored.


types/test/typescriptTest.ts
```diff
@@ -12,17 +12,17 @@ let stripe = new Stripe('sk_test_123', {
...
-  apiVersion: '2019-11-05',
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  apiVersion: '2019-11-05' as any,
 });
 
 stripe = new Stripe('sk_test_123', {
-  // @ts-ignore ignore default apiVersion.
-  apiVersion: null,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  apiVersion: null as any,
```
imranbarbhuiya, line 25: these should use `// @ts-expect-error` instead of any cast

imranbarbhuiya (changes requested):

PR #1405, opened 2022-04-29
Turn on automatic publishing

pakrym-stripe:
I verified the the script works. Switch condition to trigger on pushed tags.


dcr-stripe (approved):
LGTM!

PR #1419, opened 2022-05-11
Made line1 address parameter optional

Chigozie-Gerald:
In customer creation, the line1 parameter of the address property is optional according to the documentation. Here, it is currently acting as a required field. 
The change was made to make it optional


pakrym-stripe:
I apologize for the extreme delay,
this was implemented in https://github.com/stripe/stripe-node/pull/1505

PR #1445, opened 2022-06-08
Support updating pre-release versions

pakrym-stripe:
Fix the regex to match the pre-release versions. Also flip the current version to a non-existent 9.7.0-beta.1 so the release script doesn't fail when comparing pre-release with a non-pre-release version.


dcr-stripe (approved):
LGTM. Let's make sure to merge the Makefile change to master as well once we've validated this works?

PR #1450, opened 2022-06-13
Only publish stable version to the latest tag

pakrym-stripe:
We don't want beta version to be installed by default when someone calls `npm install stripe`. Publish beta versions to the beta tag instead.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -99,7 +99,14 @@ jobs:
           npm config set //registry.npmjs.org/:_authToken $NPM_AUTH_TOKEN
           # print the NPM user name for validation
           npm whoami
-          npm publish --otp="$(oathtool -b --totp $NPM_OTP)"
+          VERSION=$(npm -s run env echo '$npm_package_version')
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -99,7 +99,14 @@ jobs:
...
           npm whoami
-          npm publish --otp="$(oathtool -b --totp $NPM_OTP)"
+          VERSION=$(npm -s run env echo '$npm_package_version')
+          # Only publish stable versions to the latest tag
+          if [[ "$VERSION" =~ ^[^-]+$ ]]; then
+            NPM_TAG="latest"
+          else
+            NPM_TAG="beta"
+          fi
+          npm publish --otp="$(oathtool -b --totp $NPM_OTP)" --tag $NPM_TAG
```
dcr-stripe, line 109: Nit: Let's print the tag we're publishing to so we can debug from CI logs if needed?

dcr-stripe (approved):
LGTM! Just one nit and then ship!

PR #1455, opened 2022-06-17
fix: Stop using path.join to create URLs.

dcr-stripe:
r? @kamil-stripe @pakrym-stripe 

## Summary

Stop using `path.join` to create URLs and instead just join them. `path.join` is meant for file system paths, and introduces extra surprises. eg. passing `.`/`..` to a path parameter would update the entire path (`path.join('a, '..', 'b' => 'b'` and `path.join('a', '.', 'b') => a/b`). This also eliminates the need for transforming the Windows path separator.

Unfortunately because this is part of the public API we can't just do `.join('/')` and call it quits. `path.join` transforms `//` to `/`, and users can technically pass in values which lead to `//` since we don't document the expectations. So to be safe, apply the same transformation.

## Motivation

Remove unnecessary dependency. We unfortunately still use path in tests though.


lib/StripeResource.js
```diff
@@ -71,22 +69,26 @@ StripeResource.prototype = {
...
   // createFullPath which takes some data to replace them with). For example it
   // might produce: /invoices/{id}
   createResourcePathWithSymbols(pathWithSymbols) {
-    return `/${path
-      .join(this.resourcePath, pathWithSymbols || '')
-      .replace(/\\/g, '/')}`; // ugly workaround for Windows
+    return `/${this._joinUrlParts([this.resourcePath, pathWithSymbols || ''])}`;
+  },
+
+  _joinUrlParts(parts) {
```
pakrym-stripe, line 86: Do we have a test for this?

pakrym-stripe (approved):
🎉 !!

lib/StripeResource.js
pakrym-stripe: And by this I mean 

```
Unfortuantely we need to do this
    // as the functions for creating paths are technically part of the public
    // interface and so we need to preserve backwards compatibility.
```

lib/StripeResource.js
```diff
@@ -71,22 +69,26 @@ StripeResource.prototype = {
...
+    // Replace any accidentally doubled up parenthesis. This previously used
+    // path.join, which would do this as well. Unfortuantely we need to do this
```
kamil-stripe, line 88:
```suggestion
    // path.join, which would do this as well. Unfortunately we need to do this
```
;)

lib/StripeResource.js
dcr-stripe: Added tests!

pakrym-stripe (approved):

kamil-stripe (approved):

PR #1457, opened 2022-06-21
Use the generated API version

pakrym-stripe:
To support generating beta header values, use the generated ApiVersion value instead of the hardcoded one.

r? @dcr-stripe


dcr-stripe (commented):
One question and then good to go!

r? @pakrym-stripe

dcr-stripe (approved):
LGTM!

PR #1473, opened 2022-07-12
API Updates

yejia-stripe:
Codegen for openapi v164.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `customer_details` on `CheckoutSessionListParams`
* Change `LineItem.amount_discount` and `LineItem.amount_tax` to be required
* Change `Transfer.source_type` to be optional and not nullable


dcr-stripe (approved):
LGTM!

PR #1475, opened 2022-07-13
Move @types/node to devDependencies

### LINKED ISSUE #1474: Can @types/node be installed only on development mode?
### Describe the bug

I see `@types/node` package being installed on yarn install --prod, which does take space on an AWS lambda package zip file (on OSX I see it to be 1.8 MB unzipped, 360 KB zipped).

### To Reproduce

1. `rm -fr node_modules`
2. `yarn install --prod`
3. Check node_modules directory

### Expected behavior

Expected behavior? I don't expect dev dependencies to make its way into production deployments.

Does it matter? since it's just 360 KB?
Well, maybe not the highest priority, but "function" services like AWS lambda is sensitive to deployment package size since it increases cold start. So trimming it down is a good thing.

### Code snippets

_No response_

### OS

macOS, linux

### Node version

v14.19.1

### Library version

8.222.0

### API version

NA

### Additional context

_No response_

yejia-stripe:
Closes #1474 

r? @dcr-stripe

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "eslint-plugin-chai-friendly": "^0.4.0",
     "eslint-plugin-prettier": "^3.0.1",
     "mocha": "^8.3.2",
+    "mocha-junit-reporter": "^1.23.1",
     "nock": "^13.1.1",
     "node-fetch": "^2.6.2",
     "nyc": "^15.1.0",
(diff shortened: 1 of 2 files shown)


dcr-stripe (approved):
LGTM!

Did we used to use that mocha reporter package for CI and now no longer? Seems like it from https://github.com/stripe/stripe-node/pull/730 but I can't see where it was in CI...

yejia-stripe:
Oh hm I'm actually not sure either, it just didn't look like it was being used. I'll revert that part!

PR #1489, opened 2022-07-25
Remove deprecated error definitions

pakrym-stripe:
r? @kamil-stripe 

Removes error definitions that we kept for back-compat.

## Changelog

* Remove deprecated `Stripe.Stripe***Error` types. Prefer the `Stripe.errors.Stripe***Error` instead.


kamil-stripe (approved):
LGTM. Add changelog entries?

PR #1513, opened 2022-08-12
fix: update Webhooks.d.ts

lpj145:
fix: types of webhooks, Webhooks is not a class is a module, and is write and long use with 'w' not with 'W'

Sorry i did this directly from github, so i suppose something need to be fix with eslint.

### DIFF
--- a/types/Webhooks.d.ts
+++ b/types/Webhooks.d.ts
@@ -2,13 +2,8 @@
 
 declare module 'stripe' {
   namespace Stripe {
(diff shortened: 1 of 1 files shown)


pakrym-stripe:
Hi, does the current definition of `Webhooks` cause issue for your integration? Can you provide an example, please?

pakrym-stripe:
Closing for housekeeping purposes. Please free to reopen!

PR #1515, opened 2022-08-17
Add a support section to the readme

ramya-stripe:
We have user feedback on the lack of clarity around whether or not we maintain and improve older major versions of the SDK. This PR adds a support section clarifying this


dcr-stripe (approved):
Thanks Ramya! LGTM

PR #1539, opened 2022-08-30
Build Typescript and migrate StripeResource

anniel-stripe:
### Summary
r? @kamil-stripe 

This PR starts the process of migrating stripe-node to Typescript through the following steps:
1. Move files from `lib/` to `src/` directory
2. Configure TypeScript compilation through `tsconfig.json`, outputs into `lib/`
4. Add `build` and `prepack` script for Typescript compilation
5. Add step in CI/CD to build Typescript
6. Begin migrating existing stripe-node infrastructure to TypeScript with `StripeResource`, add linting overrides for .ts files

### Testing
Validated that build, lint, and test scripts pass locally. Compared previous `StripeResource.js` to compiled `lib/StripeResource.js` to make sure that there are no unexpected changes. 

With this change, the published NPM package will only contain the transpiled Javascript code in `lib/`. Our tests still import from`/lib/*.js`, and pass with this change.


package.json
```diff
@@ -24,9 +24,10 @@
   "engines": {
     "node": "^8.1 || >=10.*"
   },
-  "main": "lib/stripe.js",
+  "main": "src/stripe.js",
```
kamil-stripe, line 27: I don't think we want to change that. That's the entry point of the shipped library which should still be in "lib" directory.

anniel-stripe: Oops thanks for catching! Just updated

kamil-stripe (approved):
LGTM. @dcr-stripe for quick double check?

tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
dcr-stripe, line 1: Does this need to be aware of the types folder?

src/StripeResource.ts
```diff
@@ -356,15 +366,19 @@ StripeResource.prototype = {
     sleepSeconds = Math.max(initialNetworkRetryDelay, sleepSeconds);
 
     // And never sleep less than the time the API asks us to wait, assuming it's a reasonable ask.
-    if (Number.isInteger(retryAfter) && retryAfter <= MAX_RETRY_AFTER_WAIT) {
+    if (
+      retryAfter &&
```
dcr-stripe, line 370: Why did we introduce this first condition here? This is subtly different:

For `retryAfter = 0`:
- `Number.isInteger(retryAfter) == true`
- `retryAfter && Number.isInteger(retryAfter) == false`

dcr-stripe (commented):
LGTM overall - great work Annie! Just left some minor comments.

Let's:

1. Set up the corresponding PR to start code-generating at `src/` instead of `lib/` and re-run codegen in this branch to make sure it's a no-op before merging this.
2. Document our testing plan in this PR ie. how do we make sure the artifact we publish to NPM after this is still valid JS and compatible with existing integrations (both for JS and Typescript)

r? @anniel-stripe

src/StripeResource.ts
anniel-stripe: That was to make sure `retryAfter` is not null. It actually seems like that was not necessary (there was an error there earlier but it doesn't show up when I remove the check). Thank you for flagging!

anniel-stripe:
> LGTM overall - great work Annie! Just left some minor comments.
> 
> Let's:
> 
> 1. Set up the corresponding PR to start code-generating at `src/` instead of `lib/` and re-run codegen in this branch to make sure it's a no-op before merging this.
> 2. Document our testing plan in this PR ie. how do we make sure the artifact we publish to NPM after this is still valid JS and compatible with existing integrations (both for JS and Typescript)
> 
> r? @anniel-stripe

1. Just submitted the PR!
2. The NPM package will only contain the transpiled Javascript code in `lib/`, which is what is being used in the tests (imports from `/lib/*.js`. All of these tests pass with this change. Will update the PR description.

dcr-stripe (approved):
LGTM!

PR #1543, opened 2022-09-02
Migrate Stripe infrastructure to Typescript

anniel-stripe:
### Summary
r? @kamil-stripe 

Convert the following files to Typescript:
* StripeMethod
* StripeResource
* autoPagination
* makeRequest
* stripe
* Error

### Testing
Validated that changes build and pass existing tests. Checked that transpiled JS files look functionally identical to the original JS. No observed changes in behavior when the library is imported and used in an external project.

### DIFF
--- a/src/Error.ts
+++ b/src/Error.ts
@@ -1,11 +1,63 @@
 'use strict';
 
+type RawErrorType =
+  | 'card_error'
+  | 'invalid_request_error'
+  | 'api_error'
+  | 'idempotency_error'
+  | 'rate_limit_error'
+  | 'authentication_error'
+  | 'invalid_grant';
+
+type StripeRawError = {
+  message?: string;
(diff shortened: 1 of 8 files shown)


src/Error.ts
```diff
@@ -60,83 +112,69 @@ class StripeError extends Error {
  * CardError is raised when a user enters a card that can't be charged for
  * some reason.
  */
-class StripeCardError extends StripeError {}
+export class StripeCardError extends StripeError {}
```
kamil-stripe, line 115: how does the compiled code looks like now for this export?

kamil-stripe (approved):

src/Error.ts
anniel-stripe: It now looks like this:

```
exports.StripeCardError = StripeCardError;
```

PR #1546, opened 2022-09-07
fix: `customer` on PaymentMethod is either customer id (`string`) or `null`

elitan:


elitan:
never mind. It can be the full customer object using Expanding Responses: https://stripe.com/docs/api/expanding_objects

PR #1547, opened 2022-09-07
Revert Typescript changes

anniel-stripe:
## Summary
r? @kamil-stripe 
cc @yejia-stripe

Reset master to https://github.com/stripe/stripe-node/commit/18f88e487723cba994afd38dc645a709211e33b3 .


kamil-stripe (commented):
I think we should remove `src` directory too, right?

kamil-stripe (approved):

PR #1551, opened 2022-09-13
Re-introduce Typescript changes

anniel-stripe:
### Summary
r? @kamil-stripe 

Re-introduce the changes reverted in https://github.com/stripe/stripe-node/pull/1547 , with additional changes to minimize differences between the original and TypeScript-compiled JavaScript code.

The first 3 commits of this PR are separated as follows:

1. ["Configure Typescript build and CI" ](https://github.com/stripe/stripe-node/pull/1551/commits/dcba64af199cada1f8091e2c0a0280e8030ef307) corresponds to the changes in https://github.com/stripe/stripe-node/pull/1539 (also adds yarn install to prepack script so that dependencies are correctly installed before building, as in https://github.com/stripe/stripe-node/pull/1545)
2. ["Migrate stripe, StripeResource, StripeMethod, autoPagination, makeRequest, and Error"](https://github.com/stripe/stripe-node/pull/1551/commits/b5231441b665071c45fd936b0945519f1ae0bfbf) corresponds to the changes in   https://github.com/stripe/stripe-node/pull/1543
3. ["Enable no-var-requires, update imports/exports"](https://github.com/stripe/stripe-node/pull/1551/commits/649e6fb0c551a945b6d3322ec4fabe2261537a19) removes the no-var-requires rule, and updates imports/exports in the migrated TS files so that the compiled JS code is functionally identical to the originals, and does not include the line `Object.defineProperty(exports, "__esModule", { value: true });`

The remaining commits add test projects importing `stripe` with both CommonJS- and ES-style imports, and also updates imports / exports to follow this pattern:
1. Convert all `module.exports` to `export =` syntax in migrated TS files, export for in `stripe.ts` to preserve named and default exports.
2. Convert all `const = require` to `import = require` syntax for modules that have been migrated to TypeScript to get type checking


src/autoPagination.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const makeRequest = require('./makeRequest');
-const utils = require('./utils');
+const utils_1 = require('./utils');
```
pakrym-stripe, line 4: why do we need the rename?

src/stripe.ts
```diff
@@ -483,7 +486,7 @@ Stripe.prototype = {
    */
   getClientUserAgentSeeded(seed, cb) {
     this.getUname((uname) => {
-      const userAgent = {};
+      const userAgent: any = {};
```
pakrym-stripe, line 489:
```suggestion
      const userAgent: Record<...> = {};
```

package.json
```diff
@@ -27,6 +27,7 @@
   "main": "lib/stripe.js",
   "types": "types/2022-08-01/index.d.ts",
   "devDependencies": {
+    "@istanbuljs/nyc-config-typescript": "^1.0.2",
```
pakrym-stripe, line 30: why this particular config?

src/StripeMethod.ts
```diff
@@ -1,7 +1,5 @@
-'use strict';
-
-const utils = require('./utils');
-const makeRequest = require('./makeRequest');
+import * as utils from './utils';
```
pakrym-stripe, line 1: How to we decide which types to rename to `.ts` and which to keep as `.js`?

src/autoPagination.ts
anniel-stripe: I renamed this to address the error `Cannot redeclare block-scoped variable 'utils'.ts(2451)` (due to `utils` already being declared in stripe.js).

src/StripeMethod.ts
anniel-stripe: The goal is to eventually change everything to `.ts` - this PR starts by migrating the most important Stripe infrastructure files. The remaining infrastructure will be migrated next, followed by generated files and tests.

package.json
anniel-stripe: Ah good catch, I preemptively added that when I was thinking about test coverage on the TS files, but it's unnecessary because our tests still use the compiled JS in `lib/` Will remove!

src/autoPagination.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const makeRequest = require('./makeRequest');
```
pakrym-stripe, line 3: sometimes we use `import makeRequest = require` and sometimes `const makeRequest = require` is it intentional?

src/makeRequest.ts
```diff
@@ -1,6 +1,4 @@
-'use strict';
-
-const utils = require('./utils');
+import * as utils from './utils';
```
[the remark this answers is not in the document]
anniel-stripe: (From offline discussion)

We decided to do the following:
1. Convert all `module.exports` to `export = ` syntax
2. Convert all `const  = require` to `import = require` syntax for modules that have been migrated to TypeScript to get type checking

src/autoPagination.ts
anniel-stripe: (Copied from offline discussion)

This was intentional to avoid emitting the `Object.defineProperty(exports, "__esModule", { value: true });` line.

lib/Error.js
```diff
@@ -1,142 +1,138 @@
-'use strict';
```
pakrym-stripe, line 1: Let's mark these files as generated code in .gitattributes : https://github.com/stripe-dev/api-codegen-next/blob/master/.gitattributes#L1


This will allow github to collapse changes in them by default.

anniel-stripe: Done!

lib/Error.js
```diff
@@ -1,142 +1,138 @@
...
-        return new StripeAuthenticationError(rawStripeError);
-      case 'rate_limit_error':
-        return new StripeRateLimitError(rawStripeError);
-      case 'idempotency_error':
-        return new StripeIdempotencyError(rawStripeError);
-      case 'invalid_grant':
-        return new StripeInvalidGrantError(rawStripeError);
-      default:
-        return new StripeUnknownError(rawStripeError);
+    constructor(raw) {
```
kamil-stripe, line 7: This is a little different. Please add default value to `raw`. Currently, the lines like `this.code = raw.code;` will fail with `Cannot read properties of undefined`.

lib/StripeResource.js
```diff
@@ -1,619 +1,434 @@
...
-
-            return jsonResponse;
-          },
-          (e) => {
-            throw new StripeAPIError({
-              message: 'Invalid JSON received from the Stripe API',
-              exception: e,
-              requestId: headers['request-id'],
+        }
+        else {
```
kamil-stripe, line 57: Not your fault but weird formatting

lib/StripeResource.js
```diff
@@ -1,619 +1,434 @@
...
+        };
+        const prepareAndMakeRequest = (error, data) => {
+            if (error) {
+                return callback(error);
+            }
+            requestData = data;
+            this._stripe.getClientUserAgent((clientUserAgent) => {
+                const apiVersion = this._stripe.getApiField('version');
+                const headers = this._makeHeaders(auth, requestData.length, apiVersion, clientUserAgent, method, options.headers, options.settings);
+                makeRequest(apiVersion, headers, 0);
```
kamil-stripe, line 423: Note to self: get back to that `0` at the end.

lib/Webhooks.js
```diff
@@ -1,277 +1,174 @@
 'use strict';
-
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
kamil-stripe, line 2: Why we used async/await targeting ES6? 🤷🏻

lib/Webhooks.js
```diff
@@ -1,277 +1,174 @@
...
-
-    return generatedHeader;
-  },
+    DEFAULT_TOLERANCE: 300,
+    constructEvent(payload, header, secret, tolerance, cryptoProvider) {
+        this.signature.verifyHeader(payload, header, secret, tolerance || Webhook.DEFAULT_TOLERANCE, cryptoProvider);
+        const jsonPayload = JSON.parse(payload);
+        return jsonPayload;
+    },
+    constructEventAsync(payload, header, secret, tolerance, cryptoProvider) {
```
kamil-stripe, line 20: I think we may want to bump the spec to ES2017 to not include that __awaiter helper method if we have already used async/await.

kamil-stripe: Everything we use from ES2017 is supported by Node 8 so I think the change is pretty safe: https://node.green/#ES2017

lib/Webhooks.js
```diff
@@ -1,277 +1,174 @@
 'use strict';
```
kamil-stripe, line 1: Note to self: this file to be re-reviewed after adjusting target framework.

lib/crypto/NodeCryptoProvider.js
```diff
@@ -1,26 +1,32 @@
 'use strict';
-
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
kamil-stripe, line 2: __awaiter again: to be revisited

lib/crypto/SubtleCryptoProvider.js
```diff
@@ -1,69 +1,56 @@
 'use strict';
-
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
kamil-stripe, line 2: and one more

lib/stripe.js
```diff
@@ -1,147 +1,104 @@
...
-  bindings_version: Stripe.PACKAGE_VERSION,
-  lang: 'node',
-  publisher: 'stripe',
-  uname: null,
-  typescript: false,
-  ...determineProcessUserAgentProperties(),
-};
-
+const { determineProcessUserAgentProperties, emitWarning } = utils;
+Stripe.USER_AGENT = Object.assign({ bindings_version: Stripe.PACKAGE_VERSION, lang: 'node', publisher: 'stripe', uname: null, typescript: false }, determineProcessUserAgentProperties());
```
kamil-stripe, line 12: Why this uses `Object.assign` instead of a regular assignment?

lib/stripe.js
```diff
@@ -172,439 +127,363 @@ Stripe.createNodeCryptoProvider = () => {
...
-module.exports = Stripe;
-
 // expose constructor as a named property to enable mocking with Sinon.JS
-module.exports.Stripe = Stripe;
-
+Stripe.Stripe = Stripe;
 // Allow use with the TypeScript compiler without `esModuleInterop`.
 // We may also want to add `Object.defineProperty(exports, "__esModule", {value: true});` in the future, so that Babel users will use the `default` version.
-module.exports.default = Stripe;
+Stripe.default = Stripe;
```
kamil-stripe, line 488: This LGTM but double checking if you can import it the same way

lib/stripe.js
```diff
@@ -172,439 +127,363 @@ Stripe.createNodeCryptoProvider = () => {
...
+        return config;
+    },
 };
-
-module.exports = Stripe;
-
 // expose constructor as a named property to enable mocking with Sinon.JS
-module.exports.Stripe = Stripe;
-
+Stripe.Stripe = Stripe;
```
kamil-stripe, line 485: This is an interesting line ;)

testProjects/mjs/index.js
```diff
@@ -0,0 +1,7 @@
+import Stripe from 'stripe';
+
+const stripe = new Stripe(process.argv[2]);
+
+stripe.customers.create({
```
kamil-stripe, line 5: Can we import all infrastructure classes that are exposed to customers? i.e. StripeError.

testProjects/mjs/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "mjs",
+  "type": "module",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "dependencies": {
+    "stripe": "file:../../"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
kamil-stripe, line 11: I wouldn't actually call a resource method right now. I'd focus on on importing things like StripeError and later running it (maybe with `npx`) and check for runtime errors.

kamil-stripe (commented):
Few comments but this looks promising!

lib/stripe.js
anniel-stripe: (also discussed over Slack) It looks like Object spread was included starting in ES2018. The `Stripe.USER_AGENT = Object.assign...` syntax gets changed back to the original `Stripe.USER_AGENT = {...}` if we update the target to ES2018, but it wouldn't be safe to bump the spec to ES2018 since older versions of Node that we support would error: https://node.green/#ES2018

lib/StripeResource.js
anniel-stripe: Resolving for now, feel free to open this back up if you want me to change something!

lib/Webhooks.js
anniel-stripe: Same as above, resolving for now but feel free to open this back up if you want me to change something!

package.json
```diff
@@ -55,7 +55,7 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "tsc -p tsconfig.json",
+    "build": "tsc -p tsconfig.json; sed -i.bak '/Object.defineProperty(exports, \"__esModule\", { value: true });/d' ./lib/stripe.js; rm ./lib/stripe.js.bak",
```
kamil-stripe, line 58: Can we extract it to its own command?

kamil-stripe (approved):

kamil-stripe:
LGTM. Let's talk with @xavdid-stripe tomorrow.

package.json
anniel-stripe: Done!

xavdid-stripe (approved):
I gave the following files a fairly thorough read:

- StripeMethod
- StripeResource
- autoPagination.js
- makeRequest.js
- stripe.js
- Error.js

I spot checked a few more, but stopped when they mostly had whitespace lines removed.

Nothing jumped out as being functionally different in those files - most changes were formatting-only (or explicitly commented on as a change).

If you're looking for a more thorough comparison, one option for surfacing any non-superficial changes would be to take `src/x.js` and `lib/x.js` either:

- run them both through a formatter like `prettier` and then diff that result
- diff the ASTs of each, so we can better compare structural changes vs formatting ones

Happy to run with that a bit if it's something you're interested in! Or we can probably call this "good enough" 😁

lib/StripeResource.js
```diff
@@ -263,33 +223,29 @@ StripeResource.prototype = {
...
     }`;
   },
-
   _errorHandler(req, requestRetries, callback) {
     return (message, detail) => {
       callback.call(
         this,
         new StripeConnectionError({
           message: this._generateConnectionErrorMessage(requestRetries),
-          detail: error,
```
xavdid-stripe, line 286: Looks like this line changed. Was that intentional?

lib/StripeResource.js
```diff
@@ -498,16 +429,15 @@ StripeResource.prototype = {
         requestRetries + 1
       );
     };
-
     const makeRequest = (apiVersion, headers, numRetries) => {
       // timeout can be set on a per-request basis. Favor that over the global setting
       const timeout =
         options.settings &&
+        options.settings.timeout &&
```
xavdid-stripe, line 436: this is TS being extra defensive, no problem there.

lib/StripeResource.js
```diff
@@ -598,11 +518,9 @@ StripeResource.prototype = {
           options.headers,
           options.settings
         );
-
-        makeRequest(apiVersion, headers);
+        makeRequest(apiVersion, headers, 0);
```
xavdid-stripe, line 521: This 0 is an addition- I remember a comment from a previous revision, but it seems to have been wiped from GH.

lib/utils.js
```diff
@@ -150,12 +133,10 @@ const utils = (module.exports = {
       if (typeof arg === 'string') {
         opts.auth = args.pop();
       } else if (utils.isOptionsHash(arg)) {
-        const params = {...args.pop()};
-
+        const params = Object.assign({}, args.pop());
```
xavdid-stripe, line 136: this is functionally equivalent

lib/stripe.js
```diff
@@ -116,32 +98,24 @@ function Stripe(key, config = {}) {
...
-
   this._prevRequestMetrics = [];
   this._enableTelemetry = props.telemetry !== false;
-
   // Expose StripeResource on the instance too
   this.StripeResource = Stripe.StripeResource;
 }
-
-Stripe.errors = require('./Error');
+Stripe.errors = _Error;
```
xavdid-stripe, line 113: These are just repoints, also fine

xavdid-stripe (commented):
I know this is merged, but wanted to quickly comment on the changes that popped out. Thanks for running prettier, really helped find changes!

PR #1554, opened 2022-09-15
Fixing an issue with metadata's type

### LINKED ISSUE #1553: `Stripe.Metadata` has incorrect type
### Describe the bug

The type for `Stripe.Metadata` is `{ [name: string]: string; }` which is incorrect; this specifies that all keys have a value of `string` when in reality, any key could be `string` OR `undefined`.

### To Reproduce

``` typescript
const { wat } = {} as Stripe.Metadata;
```

image

### Expected behavior

Any valued pulled from a metadata object should have the type `string | undefined`

### Code snippets

```JavaScript
typescript
const { wat } = {} as Stripe.Metadata;
```
```


### OS

macOS

### Node version

Node v16.13.0

### Library version

stripe 9.16.0

### API version

'2020-08-27'

### Additional context

Fixing this will be a breaking change for users who consume this type in typescript

SpencerKaiser:
### ⚠️ THIS IS A BREAKING CHANGE FOR TYPESCRIPT USERS
Only merge this into a `minor` or `major` release

Fixes #1553


pakrym-stripe:
Unfortunately, as mentioned in the issue, we don't think this is the right change to make.

PR #1564, opened 2022-09-28
API Updates

yejia-stripe:
Codegen for openapi v197.
cc @stripe/api-libraries

## Changelog
* Change type of `Charge.payment_method_details.card_present.incremental_authorization_supported` and `Charge.payment_method_details.card_present.overcapture_supported` from `boolean | null` to `boolean`
* Add support for `created` on `Checkout.Session`
* Add support for `setup_future_usage` on `PaymentIntent.payment_method_options.pix`, `PaymentIntentConfirmParams.payment_method_options.pix`, `PaymentIntentCreateParams.payment_method_options.pix`, and `PaymentIntentUpdateParams.payment_method_options.pix`
* Deprecate `CheckoutSessionCreateParams.subscription_data.items` (use the `line_items` param instead). This will be removed in the next major version.


types/2022-08-01/Checkout/Sessions.d.ts
```diff
@@ -3146,7 +3151,7 @@ declare module 'stripe' {
           description?: string;
 
           /**
-           * A list of items, each with an attached plan, that the customer is subscribing to. Prefer using `line_items`.
+           * This parameter is deprecated. Use the line_items parameter on the Session instead.
```
anniel-stripe, line 3154: This deprecation was mentioned in other changelogs, but not in the one for Node. Is that intentional?

anniel-stripe (approved):

PR #1568, opened 2022-09-30
Upgrade dependencies

kamil-stripe:


pakrym-stripe:
Can you add some context on why are we updating?

package.json
```diff
@@ -27,31 +27,30 @@
...
   "resolutions": {
-    "ansi-regex": "5.0.1",
     "minimist": "1.2.6",
-    "nanoid": "3.2.0"
+    "nanoid": "^3.2.0"
   },
   "dependencies": {
-    "@types/node": ">=8.1.0",
-    "qs": "^6.10.3"
+    "@types/node": "^18.7.23",
```
pakrym-stripe, line 52: If we are the only package that brings `types/node` dependency to customer's project this change can be breaking.

kamil-stripe:
> Can you add some context on why are we updating?

I was doing an analysis on which packages can't be upgraded and decided to update outdated packages. Except `@types/node` I don't see why not. I am going to revert the @types/node change, I was supposed to do that on Friday.

types/net/net.d.ts
```diff
@@ -76,7 +76,8 @@ declare module 'stripe' {
      */
     export const createFetchHttpClient: (
       /** When specified, interface should match the Web Fetch API function. */
-      fetchFn?: Function
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
pakrym-stripe, line 79: Is this correct? Isn't this a user-facing type?

PR #1572, opened 2022-10-06
API Updates for beta branch

kamil-stripe:
Codegen for openapi v201.
cc @stripe/api-libraries

## Changelog


richardm-stripe (approved):
This looks peaceful but is substantially more than the latest codegen?

PR #1574, opened 2022-10-07
Split StripeSignatureVerificationError detail field into header and payload

anniel-stripe:
## Summary
Splits detail field in `Webhooks.ts` to match the [convention for the same error in stripe-php](https://github.com/stripe/stripe-php/blob/c7cfc8bdf68f063cdd835f1fa9d6faa44b5cc8c3/lib/Exception/SignatureVerificationException.php#L23).

## Motivation
Remove `ts-expect-error` because `detail` is of type string.

### DIFF
--- a/lib/Error.js
+++ b/lib/Error.js
@@ -93,7 +93,13 @@ class StripeConnectionError extends StripeError {}
  * SignatureVerificationError is raised when the signature verification for a
  * webhook fails
  */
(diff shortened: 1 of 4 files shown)


kamil-stripe:
That should be staged into sdk-release/next-major. It's a breaking change.

kamil-stripe (approved):
🚀

Issue #1101, opened 2020-12-29
Math.random() is not a cryptographically secure way to generate UUIDv4s


gen0cide:
https://github.com/stripe/stripe-node/blob/ec960ead66f9942d68804f1a91dce8988d8d9660/lib/utils.js#L409

The function uses an outdated method of generating UUIDv4s that does not follow current PRNG guidance for UUIDv4 generation.

The referenced comment above the function (https://stackoverflow.com/a/2117523) has an updated implemention:

```
function uuidv4() {
  return ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(/[018]/g, c =>
    (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
  );
}
```

Given that Stripe is used for payment processing, it seems like sound cryptographic PRNG usage should be the default.

brandur-stripe:
The new implementation would be an improvement, but just to be clear: idempotency keys are scoped _entirely_ to single user accounts so the worst that could happen is that a user accidentally reuses one of their own idempotency keys, and even with a non-crypto random number generation, that's very unlikely to ever happen, especially since it would have to occur with a 24 hour window.

Exciting to see on the linked SO page that there's the UUID module is coming down the pipeline for inclusion in the core language.

tjfontaine-stripe:
For tracking purposes, upstream is considering an implementation for inclusion as well https://github.com/nodejs/node/pull/36729

PR #1585, opened 2022-10-17
use native UUID method if available

xavdid-stripe:
Fixes https://github.com/stripe/stripe-node/issues/1101 by using the native UUID implementation, if available. Added a (very) basic test as a starter.

### DIFF
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -616,19 +616,27 @@ describe('utils', () => {
       let randomUUID$;
       let called;
       beforeEach(() => {
-        called = false;
-        randomUUID$ = crypto.randomUUID;
-        crypto.randomUUID = () => {
-          called = true;
-          return 'no, YOU you id';
-        };
+        // if it's available, mock it and ensure it's called
+        // otherwise, skip this whole operation
+        if (crypto.randomUUID) {
+          called = false;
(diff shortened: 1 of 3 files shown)


richardm-stripe:
LGTM, I pushed a commit and wrote one of the tests you described

test/utils.spec.js
```diff
@@ -609,6 +609,37 @@ describe('utils', () => {
...
+
+  describe('uuid', () => {
+    describe('crypto.randomUUID', () => {
+      const crypto = require('crypto');
+      let randomUUID$;
+      let called;
+      beforeEach(() => {
+        called = false;
+        randomUUID$ = crypto.randomUUID;
+        crypto.randomUUID = () => {
```
xavdid-stripe, line 621: We might want to only run this if the function exists otherwise. As it stands, old nodes will call this rather than running the fallback code. But it's great otherwise!

xavdid-stripe:
@richardm-stripe  Ok, I pushed a tweak to your test so we only mock when the function was otherwise available.

Mind giving it a quick review when you have a chance?

xavdid-stripe:
Ope! Turns out auto-merge doesn't wait for reviewers 🤔 Probably still take a peek and lmk if you see anything we should tweak. Thank you!

PR #1591, opened 2022-10-24
Check if stripe-signature header exists

slackr:
Since stripe-signature could be undefined, we check it first

"Argument of type 'string | string[] | undefined' is not assignable to parameter of type 'string | string[] | Buffer'.\n  Type 'undefined' is not assignable to type 'string | string[] | Buffer'.",

### DIFF
--- a/examples/webhook-signing/typescript-node-express/express-ts.ts
+++ b/examples/webhook-signing/typescript-node-express/express-ts.ts
@@ -32,6 +32,11 @@ app.post(
   // Stripe requires the raw body to construct the event
   express.raw({type: 'application/json'}),
   (req: express.Request, res: express.Response): void => {
+    if (!req.headers['stripe-signature']) {
+      console.log('❌ Error message: missing stripe-signature header');
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


examples/webhook-signing/typescript-node-express/express-ts.ts
```diff
@@ -32,6 +32,11 @@ app.post(
   // Stripe requires the raw body to construct the event
   express.raw({type: 'application/json'}),
   (req: express.Request, res: express.Response): void => {
+    if (!req.headers['stripe-signature']) {
+      console.log('❌ Error message: missing stripe-signature header');
```
pakrym-stripe, line 36: Do you mind adding a `res.sendStatus(400);` call too?

pakrym-stripe:
Hi! Sorry for such a delayed response. Do you mind merging `master`  and adding a `res.sendStatus(400);` call to return a 400 status code?

pakrym-stripe:
Closing for housekeeping purposes. Please feel free to reopen!

PR #1597, opened 2022-11-01
Remove deprecated configuration setter methods

anniel-stripe:
## Summary
Remove deprecated configuration setter methods. Update tests that use these methods to set the corresponding property on the `config` object instead.

## Changelog
* Remove deprecated configuration setter methods (`setHost`, `setProtocol`, `setPort`, `setApiVersion`, `setApiKey`, `setTimeout`, `setAppInfo`, `setHttpAgent`, `setMaxNetworkRetries`, and `setTelemetryEnabled`). Use the config object to set these options instead, for example:
  ```typescript
    apiVersion: '2019-08-08',
    maxNetworkRetries: 1,
    httpAgent: new ProxyAgent(process.env.http_proxy),
    timeout: 1000,
    host: 'api.example.com',
    port: 123,
    telemetry: true,
  });
  ```


test/stripe.spec.js
```diff
@@ -276,25 +276,28 @@ describe('Stripe Module', function() {
...
     it('Should allow me to set a custom timeout', () => {
-      stripe.setTimeout(900);
+      const stripe = require('../lib/stripe')('sk_test', {
+        timeout: 900,
+      });
       expect(stripe.getApiField('timeout')).to.equal(900);
     });
     it('Should allow me to set null, to reset to the default', () => {
-      stripe.setTimeout(null);
+      const stripe = require('../lib/stripe')('sk_test', {
```
pakrym-stripe, line 291:
```suggestion
      const stripe = Stripe('sk_test', {
```

here and elsewhere.

types/2022-08-01/index.d.ts
```diff
@@ -281,27 +279,6 @@ declare module 'stripe' {
       event: 'response',
       handler: (event: Stripe.ResponseEvent) => void
     ): void;
-
-    setProtocol(protocol: string): void;
```
pakrym-stripe, line 285: this wasn't deprecated here.

pakrym-stripe: but we are removing everything else so not sure if keeping the protocol around is worth it.

pakrym-stripe:
Do you mind adding a code snippet to the changelog message that illustrates the new way?

pakrym-stripe (approved):

types/2022-08-01/index.d.ts
pakrym-stripe: We never shipped definitions from that file, right?

anniel-stripe: Oh sorry no we didn't. I wonder why setProtocol wasn't marked deprecated in [this commit](https://github.com/stripe/stripe-node/commit/5c5b557bcb69eec2e786b652c0c16bea17626355) but the rest were...

pakrym-stripe: I assume we just missed it.

test/resources/EphemeralKeys.spec.js
```diff
@@ -3,15 +3,15 @@
 const stripe = require('../../testUtils').getSpyableStripe();
```
pakrym-stripe, line 3: can we remove this global? it's confusing when helpers shadow the global with the parameter.

test/StripeResource.spec.js
```diff
@@ -1023,8 +1023,6 @@ describe('StripeResource', () => {
 
   describe('Request Timeout', () => {
     it('should allow the setting of a request timeout on a per-request basis', (done) => {
-      stripe.setTimeout(1000);
```
pakrym-stripe, line 1026: Why didn't this test start failing?

test/stripe.spec.js
```diff
@@ -308,61 +311,77 @@ describe('Stripe Module', function() {
...
+            name: 'MyAwesomeApp',
+            version: '1.2.345',
+          },
         });
         expect(stripe._appInfo).to.eql({
           name: 'MyAwesomeApp',
           version: '1.2.345',
         });
 
-        stripe.setAppInfo({
```
pakrym-stripe, line 354: Let's try to remove the global from this file as well.

test/StripeResource.spec.js
anniel-stripe: This tests that the request is sent with a timeout of 10 when configured as a per-request option

test/stripe.spec.js
anniel-stripe: Decided offline to follow convention in other existing tests in this file of using the name `newStripe` for local stripe instances.

test/resources/EphemeralKeys.spec.js
```diff
@@ -51,43 +52,35 @@ describe('EphemeralKey Resource', () => {
...
 
       it('Sends the correct stripe-version', () => {
-        sendsCorrectStripeVersion();
+        sendsCorrectStripeVersion(stripe);
       });
     });
   });
 
   describe('delete', () => {
+    const stripe = require('../../testUtils').getSpyableStripe();
```
pakrym-stripe, line 82: `const testUtils = require('../../testUtils');` once?

pakrym-stripe (approved):

PR #1600, opened 2022-11-04
Remove more deprecated items

anniel-stripe:
## Summary
Remove remaining items marked as deprecated for upcoming major.

## Changelog
* Remove deprecated basic method definitions in StripeMethod.basic.js. Use basic methods defined on a resource instead.
* Remove deprecated option names. Use the following option names instead (`OLD`->`NEW`): `api_key`->`apiKey`, `idempotency_key`->`idempotencyKey`, `stripe_account`->`stripeAccount`, `stripe_version`->`apiVersion`, `stripeVersion`->`apiVersion`.

PR #1608, opened 2022-11-07
Next major release changes

anniel-stripe:
# Changelog
## 11.0.0 - 2022-11-15

"⚠️" symbol highlights breaking changes.

### ⚠️ Changed
* Drop support for Node.js 8 and 10. We now support Node.js 12+. ((#1579)
* Change `StripeSignatureVerificationError` to have `header` and `payload` fields instead of `detail`. To access these properties, use `err.header` and `err.payload` instead of `err.detail.header` and `err.detail.payload`. (#1574)

### ⚠️ Removed
* Remove `Orders` resource. (#1580)
* Remove `SKU` resource (#1583)
* Remove deprecated `CheckoutSessionCreateParams.subscription_data.items`. (#1580)
* Remove deprecated configuration setter methods (`setHost`, `setProtocol`, `setPort`, `setApiVersion`, `setApiKey`, `setTimeout`, `setAppInfo`, `setHttpAgent`, `setMaxNetworkRetries`, and `setTelemetryEnabled`). (#1597)

  Use the config object to set these options instead, for example:
  ```typescript
    apiVersion: '2019-08-08',
    maxNetworkRetries: 1,
    httpAgent: new ProxyAgent(process.env.http_proxy),
    timeout: 1000,
    host: 'api.example.com',
    port: 123,
    telemetry: true,
  });
  ```
* Remove deprecated basic method definitions. (#1600)
  Use basic methods defined on the resource instead.
  ```typescript
  // Before
  basicMethods: true

  // After
  create: stripeMethod({
    method: 'POST',
    fullPath: '/v1/resource',
  }),
  list: stripeMethod({
    method: 'GET',
    methodType: 'list',
    fullPath: '/v1/resource',
  }),
  retrieve: stripeMethod({
    method: 'GET',
    fullPath: '/v1/resource/{id}',
  }),
  update: stripeMethod({
    method: 'POST',
    fullPath: '/v1/resource/{id}',
  }),
  // Avoid 'delete' keyword in JS
  del: stripeMethod({
    method: 'DELETE',
    fullPath: '/v1/resource/{id}',
  }),
  ```
* Remove deprecated option names. Use the following option names instead (`OLD`->`NEW`): `api_key`->`apiKey`, `idempotency_key`->`idempotencyKey`, `stripe_account`->`stripeAccount`, `stripe_version`->`apiVersion`, `stripeVersion`->`apiVersion`. (#1600)
* Remove `charges` field on `PaymentIntent` and replace it with `latest_charge`. (#1614 )
* Remove deprecated `amount` field on `Checkout.Session.LineItem`. (#1614 )
* Remove support for `tos_shown_and_accepted` on `Checkout.Session.PaymentMethodOptions.Paynow`. (#1614 )


xavdid-stripe (commented):
These look good! My only suggestion is adding small blurb about what (if anything) users will need to change to handle the `StripeSignatureVerificationError` changes

PR #1609, opened 2022-11-07
Update LatestApiVersion and types reference

anniel-stripe:


anniel-stripe:
Closing, actually need to fix [API_VERSION file](https://github.com/stripe/stripe-node/blob/master/API_VERSION)

PR #1619, opened 2022-11-11
Annotate prototypes with types

pakrym-stripe:
I've noticed prototypes are not getting typed that hid a few issues.


anniel-stripe (approved):
Thanks for adding more types / cleaning up!

PR #1624, opened 2022-11-21
Split resource and service types

pakrym-stripe:
Splits resource object definitions from the resource implementation that contains methods and puts them into separate files.

Adds a typescript test project.

Sanity checked by running:

```
git checkout master
grep -ohEi 'interface \w+' -R types/2022-11-15 | sort > /tmp/before.txt
git checkout pakrym/Split_resource_and_service_types
grep -ohEi 'interface \w+' -R types/2022-11-15 | sort > /tmp/after.txt
diff /tmp/before.txt /tmp/after.txt
```


richardm-stripe:
LGTM. As a sanity check, I did
```
git diff master types/ | grep '^[-] ' | cut -c 2- | sort > /tmp/removed
git diff master types/ | grep '^[+] ' | cut -c 2- | sort > /tmp/added
diff /tmp/removed /tmp/added
```
and confirmed that nothing is being added or deleted except extra `namespace {`, `}`, which is expected, and one docstring change

PR #1630, opened 2022-12-02
Remove BASIC_METHODS from TS definitions

pakrym-stripe:
We removed the feature itself in the major.


anniel-stripe (approved):

pakrym-stripe:
@anniel-stripe, removed a few more usages, do you mind taking a look?

anniel-stripe (approved):

PR #1633, opened 2022-12-06
API Updates for beta branch

richardm-stripe:
Codegen for openapi v212.
cc @stripe/api-libraries

## Changelog


anniel-stripe (approved):
Looks like the Coveralls error fixed itself by rerunning failed jobs.

PR #1640, opened 2022-12-15
API Updates for beta branch

anniel-stripe:
Codegen for openapi v214.
cc @stripe/api-libraries

## Changelog


richardm-stripe:
LGTM! Don't quite understand why Orders.d.ts wasn't deleted in a previous update but definitely a good thing we're nuking it now!

PR #1650, opened 2023-01-10
Refactor usages of Node.js Buffer

anniel-stripe:
## Summary
As part of our efforts to remove usages of Node-specific APIs and globals to support alternative runtimes, this PR refactors our library to not depend on the Node.js Buffer class.

* In `multipart.ts`, I use`Uint8Array` instead, since instances of Buffer are also instances of `Uint8Array`.
* In `Webhooks.ts`, I replace `Buffer.isBuffer` with a [polyfill implementation](https://github.com/feross/is-buffer), since we should still support when `stripe.webhooks.constructEvent(...)` is called with a `Buffer` type.
* I also refactor `secureCompare` since the arguments we pass in are of type `string`, not `Uint8Array`.

## Testing
I've validated locally that this implementation can be used to upload a test file, and that it constructs webhook events as expected when the payload/header are raw strings or a `Buffer`s.

Also added a Webhook test that checks it can still accept payloads and headers of type `Buffer`.


test/Webhook.spec.js
```diff
@@ -48,6 +48,21 @@ describe('Webhooks', () => {
         expect(event.id).to.equal(EVENT_PAYLOAD.id);
       });
 
+      it('should return an Event instance from a payload and header with type Buffer', async () => {
+        const header = stripe.webhooks.generateTestHeaderString({
```
pakrym-stripe, line 52: let's also add tests that create and pass instances of Uint8Array

src/utils.ts
```diff
@@ -367,6 +370,18 @@ const utils = {
     return (type === 'function' || type === 'object') && !!obj;
   },
 
+  // taken from https://github.com/feross/is-buffer/blob/master/index.js
+  isBuffer: (obj: unknown): boolean => {
```
pakrym-stripe, line 374: should we just do `instanceof Uint8Array` instead of calling this function?

src/utils.ts
```diff
@@ -216,27 +216,30 @@ const utils = {
...
-    if (a.length !== b.length) {
+    if (aEncoded.length !== bEncoded.length) {
       return false;
     }
 
     // use crypto.timingSafeEqual if available (since Node.js v6.6.0),
     // otherwise use our own scmp-internal function.
     if (crypto.timingSafeEqual) {
-      return crypto.timingSafeEqual(a, b);
+      return crypto.timingSafeEqual(aEncoded, bEncoded);
```
pakrym-stripe, line 235: how did this call work before? does it allow strings?

src/multipart.ts
```diff
@@ -71,15 +76,15 @@ const streamProcessor = (
...
+    .on('data', (line: Uint8Array) => {
       bufferArray.push(line);
     })
     .once('end', () => {
-      // @ts-ignore
-      const bufferData: BufferedFile = Object.assign({}, data);
-      bufferData.file.data = Buffer.concat(bufferArray);
+      const bufferData: StreamingFile = Object.assign({}, data);
+      // @ts-ignore - bufferArray will not store Uint8Array's nested greater than depth 1
+      bufferData.file.data = bufferArray.flatMap((e) => e);
```
pakrym-stripe, line 87: Will the result here be Uint8Array? It might be Array<number> instead which is not what we want.

src/multipart.ts
```diff
@@ -15,20 +15,25 @@ const multipartDataGenerator = (
...
     Math.round(Math.random() * 1e16) + Math.round(Math.random() * 1e16)
   ).toString();
   headers['Content-Type'] = `multipart/form-data; boundary=${segno}`;
-  let buffer = Buffer.alloc(0);
+  const textEncoder = new TextEncoder();
+
+  let buffer = new Uint8Array(0);
+  const endBuffer = textEncoder.encode('\r\n');
 
   function push(l: any): void {
```
pakrym-stripe, line 28: let's type this argument

pakrym-stripe (changes requested):
Some questions. Let's also add tests for explicit passing of Uint8Array to our APIs.

src/utils.ts
anniel-stripe: `crypto.timingSafeEqual` doesn't allow strings. Before we would convert the arguments to Buffers like `a = Buffer.from(a)`. Removing these conversions makes throw the error `TypeError: The "buf1" argument must be an instance of Buffer, TypedArray, or DataView.`

src/utils.ts
anniel-stripe: I did that for the `multipart`-related usages, but for webhooks we rely on `Buffer.toString`, which behaves differently from `Uint8Array.toString`. There seem to be use cases where users rely on this behavior and pass in a `Buffer`: https://github.com/vendure-ecommerce/vendure/blob/master/packages/payments-plugin/src/stripe/stripe.service.ts#L72

src/multipart.ts
anniel-stripe: Ah you're right, will fix

src/utils.ts
pakrym-stripe: My bad, missed those lines. Let's move 
```
    const aEncoded: Uint8Array = new TextEncoder().encode(a);
    const bEncoded: Uint8Array = new TextEncoder().encode(b);
```
calls inside this if statement. We can rewrite the loop to operate on chars using `a.charCodeAt() ^ b. charCodeAt()`.

src/utils.ts
pakrym-stripe: Can we use `TextDecoder` to mimic the behavior?

pakrym-stripe: Higher order principle here is that non-node customer should be able to use all Stripe Node SDK APIs so we can't have things that work only when `Buffer` is available.

anniel-stripe: Ah I didn't realize that we could use `TextDecoder` on `Buffer` as well, updated!

src/Webhooks.ts
```diff
@@ -78,8 +78,10 @@ const Webhook: WebhookObject = {
       cryptoProvider
     );
 
-    // @ts-ignore
-    const jsonPayload = JSON.parse(payload);
+    const jsonPayload =
+      payload instanceof Uint8Array
+        ? JSON.parse(new TextDecoder().decode(payload))
```
anniel-stripe, line 83: I added this condition with the Uint8Array tests since `JSON.parse` does not work as intended on `Uint8Array`

src/multipart.ts
anniel-stripe: Typed as discussed

src/Webhooks.ts
```diff
@@ -99,7 +101,10 @@ const Webhook: WebhookObject = {
     );
 
     // @ts-ignore
```
pakrym-stripe, line 103: do we need this ts-ignore?

src/Webhooks.ts
```diff
@@ -218,9 +223,11 @@ function parseEventDetails(
   encodedHeader: WebhookHeader,
   expectedScheme: string
 ): WebhookParsedEvent {
-  const decodedPayload = Buffer.isBuffer(encodedPayload)
-    ? encodedPayload.toString('utf8')
-    : encodedPayload;
+  const textDecoder = new TextDecoder('utf8');
```
pakrym-stripe, line 226: nit: inconsistent usage of `textDecoder` as a variable vs `new TextDecoder`. Also inconsistent 'utf8' parameter usage.

test/Webhook.spec.js
```diff
@@ -48,6 +48,37 @@ describe('Webhooks', () => {
...
+
+      it('should return an Event instance from a payload and header with type Uint8Array', async () => {
+        const header = stripe.webhooks.generateTestHeaderString({
+          payload: EVENT_PAYLOAD_STRING,
+          secret: SECRET,
+        });
+
+        const textEncoder = new TextEncoder();
+        const event = await constructEventFn(
+          new Uint8Array(textEncoder.encode(EVENT_PAYLOAD_STRING)),
```
pakrym-stripe, line 74: textEncoder.encode returns a Uint8Array judging by the docs. Do we need extra construction?

test/Webhook.spec.js
pakrym-stripe: Do we have file upload tests? Can we also add Uint8Array variants there?

src/utils.ts
```diff
@@ -439,6 +444,22 @@ const utils = {
           platform: process.platform,
         };
   },
+
+  /**
+   * Joins an array of Uint8Arrays into a single Uint8Array
+   */
+  concat: (arrays: Array<Uint8Array>): Uint8Array => {
```
pakrym-stripe, line 451: Let's add a few tests for this method.

src/Error.ts
```diff
@@ -128,10 +128,14 @@ class StripeConnectionError extends StripeError {}
  * webhook fails
  */
 class StripeSignatureVerificationError extends StripeError {
-  header: string;
-  payload: string;
-
-  constructor(header: string, payload: string, raw: StripeRawError = {}) {
+  header: string | Buffer;
```
pakrym-stripe, line 131: Uint8Array?

pakrym-stripe (changes requested):
Getting very close! A few more minor comments.

src/Webhooks.ts
anniel-stripe: Ah good point, it defaults to 'utf8' but we should be explicit.

src/Webhooks.ts
anniel-stripe: No we don't :)

test/Webhook.spec.js
anniel-stripe: Oh good catch, we don't need the extra construction. Fixed!

src/utils.ts
anniel-stripe: Added a few in utils.spec.js

test/Webhook.spec.js
anniel-stripe: Added a test for passing in file data as `Uint8Array` in flows.spec.js

src/multipart.ts
```diff
@@ -43,7 +48,7 @@ const multipartDataGenerator = (
     if (Object.prototype.hasOwnProperty.call(v, 'data')) {
       const typedEntry: {
         name: string;
-        data: BufferedFile;
+        data: string | Uint8Array;
```
pakrym-stripe, line 51: nit: i think `BufferedFile` was correct here.

pakrym-stripe (approved):

src/multipart.ts
anniel-stripe: Based on log statements, it seems like `typedEntry.data` gets converted to `string | Uint8Array` because it comes from `v` , which is the result of `utils.flattenAndStringify(data)`.

PR #1652, opened 2023-01-12
Add koa example

### LINKED ISSUE #1435: Koa JS exemple for using webhook and parsing raw body

I'm using koa js framework and i don't find solution for parsing body in raw mode.

You can do examples with koa js, like the one done with express.

_No response_

### Additional context

_No response_

richardm-stripe:
## Notify

## Summary
Adds an example that uses koa.js to our `examples/` folder.

I originally thought this would be trivial. The result looked simple but I unfortunately struggled for *hours* trying to get a handle on the raw bodyusing standard-ish koa node modules. The `koa-bodyparser` README contains [pretty clear instructions](https://github.com/koajs/bodyparser#raw-body), but I did not find those at first, because I tried to copy and paste the express example and just modify it for Koa. In our express example, we skip the body-parsing middleware in the case of webhooks, which is not the right approach for Koa, or at least I couldn't figure out how to access the raw body that way.

## Motivation
Fixes #1435.


examples/webhook-signing/node-koa/koa.js
```diff
@@ -0,0 +1,65 @@
+const Stripe = require('stripe');
```
pakrym-stripe, line 1: this is a js file but you have a tsconfig is this intended?

pakrym-stripe (approved):

pakrym-stripe:
nit: `r? https://github.com/orgs/stripe/teams/api-library-reviewers` doesn't work on github.com so no one got assigned to the PR.

examples/webhook-signing/node-koa/koa.js
richardm-stripe: no

PR #1654, opened 2023-01-12
Move request-sending functionality into a common place

pakrym-stripe:
Currently all the request making logic in stripe-node resides in `StripeResource`. We'd like to be able to make requests from arbitrary places without a resource object. 

This PR moves the request-making logic into `RequestSender` class and place an instance of it onto the main Stripe client.

This follows the pattern we have in Java with https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/net/StripeResponseGetter.java#L8


src/RequestSender.ts
```diff
@@ -0,0 +1,547 @@
...
+
+  _request(
+    method: string,
+    host: string | null,
+    path: string,
+    data: RequestData,
+    auth: string | null,
+    options: RequestOptions = {},
+    callback: RequestCallback,
+    requestDataProcessor: RequestDataProcessor | null = null
```
anniel-stripe, line 399: Is there a reason why we don't use `requestDataProcessor?: RequestDataProcessor` instead here?

src/Types.d.ts
```diff
@@ -2,15 +2,14 @@
...
-type HttpHeaderValue = string | number | string[];
 type MethodSpec = {
   method: string;
   methodType: string;
   urlParams: Array<string>;
   path?: string;
   fullPath?: string;
   encode: (data: RequestData) => RequestData;
-  validator: (data: RequestData, headers: RequestHeaders) => void;
+  validator: (data: RequestData, headers: {headers: RequestHeaders}) => void;
```
anniel-stripe, line 12: Why do we have this nested `headers` field? Is this something we could refactor?

test/makeRequest.spec.js
```diff
@@ -1,68 +0,0 @@
...
-      makeRequest(mockSelf, args, {}, {});
-      expect(args).to.deep.equal([
-        {
-          stripe_account: 'bad',
-        },
-      ]);
-    });
-  });
-
-  describe('makeRequest with fullPath', () => {
```
anniel-stripe, line 31: Do we still want some of these tests in RequestSender.spec.js?

lib/makeRequest.js
```diff
@@ -1,104 +1,458 @@
 'use strict';
```
anniel-stripe, line 1: We can get rid of this file :)

anniel-stripe (changes requested):

anniel-stripe:
Mostly questions!

src/RequestSender.ts
anniel-stripe: Nevermind, I see we don't want it to be `'undefined'`

src/Types.d.ts
pakrym-stripe: Ah, the parameter should be called options: https://github.com/stripe/stripe-node/blob/3e80143129af5046c42cbec338c9049aa525face/src/resources/EphemeralKeys.js#L12

test/makeRequest.spec.js
pakrym-stripe: I copied the `does not mutate user-supplied deprecated opts` to StripeResource tests. 

The `makeRequest with fullPath` is covered by `method with fullPath -> interpolates in parameters`.

anniel-stripe (approved):

PR #1657, opened 2023-01-13
Remove __esImport suppression and pretty-printing of compiled code

pakrym-stripe:
We've added the pretty-printing to simplify code comparison during initial TS migration. We can now remove it and compare non formatted files when making changes.

We've added the `Object.defineProperty(exports, \"__esModule\", { value: true })` line suppression during the initial migration from a fear of breaking Electron customers that might not have (exports) defined. After thinking about this more I realized that we reference `module.exports` in stripe.js which won't be defined as well in that case. So it's not a new problem we are creating.

Also turned a few `const ... = require()` into `import ... = require()` and fixed type errors.


anniel-stripe (approved):
LGTM. Thanks for making these changes and adding more type-checking!

PR #1660, opened 2023-01-17
Introduce separate entry point for worker environments

anniel-stripe:
## Summary
Introduces an [exports field](https://nodejs.org/api/packages.html#conditional-exports) to the package.json that defines a separate entry point for environments where Node.js built-in modules are not supported.

Major changes in this PR:
* Define two entrypoint files: `stripe.ts` and `stripe.worker.ts`, move common code into `stripe.common.ts`
* Separate functions defined in `utils.ts` that depend on Node.js built-ins to the `platform` directory in a `NodePlatformFunctions` class that inherits from `DefaultPlatformFunctions`
* Add private `_platformFunctions` property onto Stripe object. Use this to override functions that depend on Node-specific modules in `stripe.node.ts`
* Initialize `Stripe.createNodeHttpClient` and `Stripe.createNodeCryptoProvider` to throw errors and override them in `stripe.node.ts` (won't be available for workers)
* Create generic `Stripe.createHttpClient` that defaults to `Stripe.createNodeHttpClient` normally, and `Stripe.createFetchHttpClient` in workers
* Convert affected tests to TypeScript

## Testing
I wasn't sure of a good way to add an automated test that mimics a worker environment. I tested this locally, and validated that the expected build errors have been resolved (see next section 'Remaining Work' for remaining errors).

The original build errors:
Screen Shot 2023-01-17 at 12 46 53 PM

Now:
Screen Shot 2023-01-19 at 10 52 44 AM

## Remaining Work
* Refactor  `Stripe._emitter` and `utils.checkForStream` to not use `EventEmitter` for worker environments


src/stripe.common.ts
```diff
@@ -0,0 +1,489 @@
...
+  // Expose StripeResource on the instance too
+  // @ts-ignore
+  this.StripeResource = Stripe.StripeResource;
+}
+
+Stripe.errors = _Error;
+Stripe.webhooks = require('./Webhooks');
+
+// @ts-ignore
+Stripe.createNodeHttpClient = null;
```
anniel-stripe, line 164: Setting `createNodeHttpClient` to null, overriding in `stripe.default.ts`

src/stripe.common.ts
```diff
@@ -0,0 +1,489 @@
...
+  const {FetchHttpClient} = require('./net/FetchHttpClient');
+  return new FetchHttpClient(fetchFn);
+};
+
+/**
+ * Create a CryptoProvider which uses the built-in Node crypto libraries for
+ * its crypto operations.
+ */
+// @ts-ignore
+Stripe.createNodeCryptoProvider = null;
```
anniel-stripe, line 183: Setting `createNodeCryptoProvider ` to null, overriding in `stripe.default.ts`

package.json
```diff
@@ -69,5 +69,10 @@
     "fix": "yarn lint --fix && ./scripts/updateAPIVersion.js",
     "report": "nyc -r text -r lcov report",
     "coveralls": "cat coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js"
+  },
+  "exports": {
+    "browser": "./lib/stripe.worker.js",
+    "worker": "./lib/stripe.worker.js",
+    "default": "./lib/stripe.js"
```
anniel-stripe, line 76: Alternatively, we can have this be:
```suggestion
    "node": "./lib/stripe.js",
    "default": "./lib/stripe.worker.js" // stripe.default.js ?
```
This is [what is suggested in the Node docs](https://nodejs.org/api/packages.html#conditional-exports:~:text=For%20this%20reason%2C%20using%20%22node%22%20and%20%22default%22%20condition%20branches%20is%20usually%20preferable%20to%20using%20%22node%22%20and%20%22browser%22%20condition%20branches.), but I'd prefer to test the waters with just "worker" and "browser" environments first.

src/stripe.common.ts
pakrym-stripe: what do you think about throwing a descriptive error message instead? If we mess up and customers that were using `createNodeCryptoProvider` successfully before get the wrong version of the library at least we'll give them some inclination of what happened and where to complain.

pakrym-stripe: Same for httpClient above.

src/stripe.default.ts
```diff
@@ -0,0 +1,25 @@
...
+
+Stripe.createNodeHttpClient = (agent: http.Agent): typeof HttpClient => {
+  const {NodeHttpClient} = require('./net/NodeHttpClient');
+  return new NodeHttpClient(agent);
+};
+
+Stripe.createNodeCryptoProvider = (): StripeCryptoProvider => {
+  const NodeCryptoProvider = require('./crypto/NodeCryptoProvider');
+  return new NodeCryptoProvider();
+};
```
anniel-stripe, line 16: This is where we introduce Node-specific dependencies

src/nodeUtils.ts
```diff
@@ -0,0 +1,68 @@
+const crypto = require('crypto');
+import utils = require('./utils');
+
+/**
+ * Utility functions that rely on Node.js built-in libraries.
+ */
+const nodeUtils = {
```
pakrym-stripe, line 7: Let's call this something more descriptive. `PlatformFeatures`/`PlatformFunction`/... .

We should also separate the ones that depend on platform from other general purpose utils.

Check out how HttpClient and HttpClient tests are structured. We define an interface and two implementations. Then we define a set of tests and run the same set of tests for both implementations.

package.json
pakrym-stripe: We should keep the same default as we have now.

pakrym-stripe:
Why is node 12 on fire?

pakrym-stripe (changes requested):
Mostly code structure related comments.

pakrym-stripe:
>I wasn't sure of a good way to add an automated test that mimics a worker environment. 

Create a test project similar to ones we have and invoke the cloudflare build CLI on it from tests.

src/nodeUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+      return crypto.randomUUID();
+    }
+
+    return utils.uuid4();
+  },
+
+  /**
+   * Secure compare, from https://github.com/freewil/scmp
+   */
+  secureCompare: (a: string, b: string): boolean => {
```
pakrym-stripe, line 44: This is used in webhooks, how are they getting access to it?

pakrym-stripe, line 44: Random context: https://github.com/w3c/webcrypto/issues/270#issuecomment-1184792734

anniel-stripe: Addressed in this commit: https://github.com/stripe/stripe-node/pull/1660/commits/d81bb115ba77c5d3cacb0cd907c879de5d7fc3e4

I added a reference to the Stripe instance in `Webhooks`, from which it can access `Stripe._utils.secureCompare` (also updated the call to `uuid4` in `RequestSender.ts` because it had a similar issue)

src/Webhooks.ts
```diff
@@ -57,6 +57,7 @@ type WebhookObject = {
     cryptoProvider: StripeCryptoProvider
   ) => Promise<WebhookEvent>;
   generateTestHeaderString: (opts: WebhookTestHeaderOptions) => string;
+  _stripe: StripeObject | null;
```
pakrym-stripe, line 60: Did you have to reference the entire `_stripe` object because _platformFunctions were not set when webhooks were constructed?

src/platform/DefaultPlatformFunctions.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Interface encapsulating various utility functions whose
+ * implementations depend on the platform / JS runtime.
+ */
+class DefaultPlatformFunctions {
+  safeExec(
+    cmd: string,
+    cb: (error: unknown, stdout: string | null) => void
+  ): void {
+    throw new Error('safeExec not implemented.');
```
pakrym-stripe, line 10: When will this surface? Should we abstract our the entire getUname function? I think it's the only consumer of exec that we have.

src/stripe.common.ts
```diff
@@ -0,0 +1,498 @@
...
+import HttpClient = require('./net/HttpClient');
+Stripe.HttpClient = HttpClient.HttpClient;
+Stripe.HttpClientResponse = HttpClient.HttpClientResponse;
+
+import CryptoProvider = require('./crypto/CryptoProvider');
+import EventEmitter = require('events');
+Stripe.CryptoProvider = CryptoProvider;
+
+import DefaultPlatformFunctions = require('./platform/DefaultPlatformFunctions');
+Stripe._platformFunctions = new DefaultPlatformFunctions();
```
pakrym-stripe, line 69: How about setting this below in prototype definition?

pakrym-stripe: Oh, this is a "global" and we are copying the value to an instance on every creation. Interesting.

src/stripe.common.ts
```diff
@@ -0,0 +1,498 @@
...
+  this.StripeResource = Stripe.StripeResource;
+
+  this._platformFunctions = Stripe._platformFunctions;
+}
+
+Stripe.errors = _Error;
+Stripe.webhooks = require('./Webhooks');
+
+Stripe.createNodeHttpClient = (agent: http.Agent): void => {
+  throw new Error(
```
pakrym-stripe, line 166: Should we also change which client we create by default for workers? It's a bit strange to make customers call a method when we already know the right thing to do.

pakrym-stripe: I'm thinking about the existing customers that use polyfils. If we keep calling createNodeHttpClient code that was working before will start breaking.

pakrym-stripe (changes requested):
Few more comments. This is getting better and better with every commit!

src/stripe.common.ts
anniel-stripe: That's a good point, I agree defaulting to `createFetchHttpClient` in worker environments is a better solution to avoid breaking customers. I wasn't sure if we wanted to quietly transition to fetch, or throw this error so customers know what changed and can provide their own implementation if they choose. If we call `createFetchHttpClient` by default, should we emit a warning too to let customers know their http client changed?

src/Webhooks.ts
anniel-stripe: Oh that's a good callout, I didn't need to reference the entire `_stripe` object. I think this would be best resolved if we abstracted the entire getUname function and set that on Webhooks directly, since that's the only place where a platform-specific function is necessary.

src/stripe.node.ts
```diff
@@ -0,0 +1,25 @@
+import Stripe = require('./stripe.common');
+import NodePlatformFunctions = require('./platform/NodePlatformFunctions');
+import http = require('http');
+import HttpClient = require('./net/HttpClient');
+
+Stripe._platformFunctions = new NodePlatformFunctions();
+
+Stripe.createNodeHttpClient = (agent: http.Agent): typeof HttpClient => {
```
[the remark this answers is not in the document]
anniel-stripe: I'm hesitant to do this since I thought we previously exposed `createFetchHttpClient` and `createNodeHttpClient` as public APIs to be passed into the config as `httpClient`.

src/stripe.common.ts
anniel-stripe: Since I updated webhooks to hold a reference to `_platformFunctions`, I'd like to keep this as a singleton that can be reused for webhooks and across multiple Stripe instances.

src/RequestSender.ts
```diff
@@ -1,6 +1,5 @@
-import utils = require('./utils');
-
 import _Error = require('./Error');
+import utils = require('./utils');
```
pakrym-stripe, line 2: nit: remove?

src/Webhooks.ts
```diff
@@ -57,6 +57,7 @@ type WebhookObject = {
     cryptoProvider: StripeCryptoProvider
   ) => Promise<WebhookEvent>;
   generateTestHeaderString: (opts: WebhookTestHeaderOptions) => string;
+  _platformFunctions: import('./platform/DefaultPlatformFunctions') | null;
```
pakrym-stripe, line 60: let's import this at the top of the file

src/platform/DefaultPlatformFunctions.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  constructor() {
+    this._UNAME_CACHE = null as Promise<string> | null;
+  }
+
+  /**
+   * Calls Node's built-in `exec` function.
+   * If the `child_process` module is not available, return an error.
+   */
+  safeExec(
```
pakrym-stripe, line 16: do we need this public/separate?

src/platform/DefaultPlatformFunctions.ts
```diff
@@ -0,0 +1,64 @@
...
+    cmd: string,
+    cb: (error: unknown, stdout: string | null) => void
+  ): void {
+    cb(new Error('exec not available'), null);
+  }
+
+  getUname(cb: (uname: string) => void): void {
```
pakrym-stripe, line 23: let's move the entire fancy implementation into `Node**`

src/stripe.common.ts
```diff
@@ -0,0 +1,481 @@
...
+const DEFAULT_PORT = '443';
+const DEFAULT_BASE_PATH = '/v1/';
+const DEFAULT_API_VERSION = (null as unknown) as string;
+
+const DEFAULT_TIMEOUT = 80000;
+
+Stripe.PACKAGE_VERSION = require('../package.json').version;
+
+const determineProcessUserAgentProperties = (): Record<string, string> => {
+  return typeof process === 'undefined'
```
pakrym-stripe, line 16: Looks shady. Worth extracting?

src/stripe.common.ts
```diff
@@ -0,0 +1,481 @@
...
+  const {FetchHttpClient} = require('./net/FetchHttpClient');
+  return new FetchHttpClient(fetchFn);
+};
+
+/**
+ * Create a CryptoProvider which uses the built-in Node crypto libraries for
+ * its crypto operations.
+ */
+Stripe.createNodeCryptoProvider = (): void => {
+  throw new Error(
```
pakrym-stripe, line 188: We'll need a corresponding fix in `webhooks` to create a different default crypto provider.

src/stripe.node.ts
```diff
@@ -0,0 +1,28 @@
+import Stripe = require('./stripe.common');
+import NodePlatformFunctions = require('./platform/NodePlatformFunctions');
+import http = require('http');
+import HttpClient = require('./net/HttpClient');
+
+Stripe._platformFunctions = new NodePlatformFunctions();
+Stripe.webhooks._platformFunctions = Stripe._platformFunctions;
+
+const createNodeHttpClient = (agent: http.Agent): typeof HttpClient => {
```
pakrym-stripe, line 9: Think about moving this logic into `PlatformFunctions`

src/utils.ts
```diff
@@ -339,33 +293,7 @@ const utils = {
...
-    try {
-      utils._exec(cmd, cb);
-    } catch (e) {
-      cb(e, null);
-    }
-  },
-
-  // For mocking in tests.
-  _exec: exec,
+  _exec: null,
```
pakrym-stripe, line 296: rem?

src/utils.ts
```diff
@@ -437,15 +348,6 @@ const utils = {
     return n as number;
   },
 
-  determineProcessUserAgentProperties: (): Record<string, string> => {
```
pakrym-stripe, line 440: nit: check if this change is still needed.

test/PlatformFunctions.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    });
+
+    describe('secureCompare without crypto.timingSafeEqual', () => {
+      const crypto = require('crypto');
+      let timingSafeEqual$;
+      beforeEach(() => {
+        // if it's available, mock it and ensure it's called
+        // otherwise, skip this whole operation
+        if (isNodeEnvironment && crypto.timingSafeEqual) {
+          timingSafeEqual$ = crypto.timingSafeEqual;
```
pakrym-stripe, line 100: Do we still need this logic?

test/stripe.spec.ts
```diff
@@ -239,17 +265,17 @@ describe('Stripe Module', function() {
     describe('uname', () => {
```
pakrym-stripe, line 265: Should this be moved to PlatformFunctions tests?

test/PlatformFunctions.spec.ts
anniel-stripe: This logic is still necessary, since `crypto.timingSafeEqual` is used in `secureCompare` in Node environments and we want to ensure that secure compare still works when the `crypto` operation is not available.

src/platform/DefaultPlatformFunctions.ts
anniel-stripe: Moved this functionality to `getUname` since that's the only callsite

src/platform/DefaultPlatformFunctions.ts
anniel-stripe: Moved to `NodePlatformFunctions`. I updated this function to return a promise for better error handling.

src/stripe.common.ts
anniel-stripe: Agreed, added to list of things to investigate after this PR

src/stripe.common.ts
anniel-stripe: Done in https://github.com/stripe/stripe-node/pull/1660/commits/ddb5e1562d6460a89f8d4102a2f63ded4bb91026

test/stripe.spec.ts
anniel-stripe: I don't think it should, since it tests `getClientUserAgentSeeded` which is a method on the Stripe object that's indirectly (in `getClientUserAgent`) used in `RequestSender`.

src/stripe.node.ts
anniel-stripe: Agreed, it would be nice to combine this logic all into one place. I'd prefer to do this in a later PR.

src/Webhooks.ts
```diff
@@ -145,6 +149,10 @@ const Webhook: WebhookObject = {
 
     return generatedHeader;
   },
+
+  createCryptoProvider: () => null,
```
pakrym-stripe, line 153: was this public before? If not let's prefix it with `_`.

test/stripe.spec.ts
pakrym-stripe: Ok, then instead of messing with `_platformFunctions` internals we can replace it with a new implementation of PlatformFunctions that will return whatever we'd like from uname. [Something like this](https://github.com/stripe/stripe-node/blob/3c9e553ded16c8bc39f18c3264e43ffb59b38896/testUtils/index.js#L62)

anniel-stripe: Done in https://github.com/stripe/stripe-node/pull/1660/commits/c258657a20f97081e9498a4fe568a2d6b0a33056

pakrym-stripe (approved):
Great work! :tada:

PR #1667, opened 2023-01-26
fix: added support for TypeScript "NodeNext" module resolution

GauBen:
Hey there!

Small fix for people using `{ "moduleResolution": "NodeNext" }` in their tsconfig


anniel-stripe:
Thanks so much for catching this!

anniel-stripe (approved):

PR #1668, opened 2023-01-31
Use EventTarget in worker / browser runtimes

anniel-stripe:
## Summary
1. Refactor `_emitter` property on the `Stripe` object to use `EventTarget` instead of `EventEmitter`  in worker environments.
2. Add a base class `PlatformFunctions` that is implemented by `WebPlatformFunctions` and `NodePlatformFunctions`
3. Consolidate checking for streaming data  and buffering logic into a `tryBufferData` platform function, which throws an error in workers if a `ReadableStream` is detected.

## Workarounds because of Node <15
`Event` and `EventTarget` are only available in the global scope [starting with Node 14](https://nodejs.org/api/events.html#class-event:~:text=Changes-,v15.0.0,-The%20Event%20class). Since `_StripeEvent ` needs to extend `Event` (to add the `data` object used by listener functions), I ended up creating a base class for `PlatformFunctions` and a separate `WebPlatformFunctions` that is the only place where `StripeEmitter` is imported.

In the tests I added to `PlatformFunctions.spec.ts`, I also added a check for the Node version so that `WebPlatformFunctions` test do not run in Node <= 14.

## Testing
This gets rid of the last of the build errors when importing stripe-node in Cloudflare Workers without polyfills. I tested this new implementation in Cloudflare Workers / Pages and both emit events as expected.

### DIFF
--- a/src/StripeEmitter.ts
+++ b/src/StripeEmitter.ts
@@ -11,35 +11,37 @@ class _StripeEvent extends Event {
   }
 }
 
+type Listener = (...args: any[]) => any;
+
 /** Minimal EventEmitter wrapper around EventTarget. */
 class StripeEmitter {
   eventTarget: EventTarget;
-  listenerMapping: Record<string, (...args: any[]) => any>;
+  listenerMapping: Map<Listener, Listener>;
 
-  constructor(eventTarget: EventTarget) {
-    this.eventTarget = eventTarget;
-    this.listenerMapping = {};
+  constructor() {
+    this.eventTarget = new EventTarget();
+    this.listenerMapping = new Map();
   }
 
-  on(eventName: string, listener: (...args: any[]) => void): void {
+  on(eventName: string, listener: Listener): void {
     const listenerWrapper = (event: _StripeEvent): void => {
       listener(event.data);
     };
-    this.listenerMapping[listener.toString()] = listenerWrapper;
+    this.listenerMapping.set(listener, listenerWrapper);
     return this.eventTarget.addEventListener(eventName, listenerWrapper);
   }
 
-  removeListener(eventName: string, listener: (...args: any[]) => void): void {
-    const listenerWrapper = this.listenerMapping[listener.toString()];
-    delete this.listenerMapping[listener.toString()];
-    return this.eventTarget.removeEventListener(eventName, listenerWrapper);
+  removeListener(eventName: string, listener: Listener): void {
+    const listenerWrapper = this.listenerMapping.get(listener);
+    this.listenerMapping.delete(listener);
+    return this.eventTarget.removeEventListener(eventName, listenerWrapper!);
   }
 
-  once(eventName: string, listener: (...args: any[]) => void): void {
+  once(eventName: string, listener: Listener): void {
     const listenerWrapper = (event: _StripeEvent): void => {
       listener(event.data);
     };
-    this.listenerMapping[listener.toString()] = listenerWrapper;
+    this.listenerMapping.set(listener, listenerWrapper);
     return this.eventTarget.addEventListener(eventName, listenerWrapper, {
       once: true,
     });
--- a/src/StripeEmitter.ts
+++ b/src/StripeEmitter.ts
@@ -1,0 +1,51 @@
+/** @private
+ * (For internal use in stripe-node.)
+ * Wrapper around the Event Web API. */
+class _StripeEvent extends Event {
+  data: any;
+  constructor(eventName: string, data: any) {
+    super(eventName);
+    this.data = data;
+  }
+}
+
+/** Minimal EventEmitter wrapper around EventTarget. */
+class StripeEmitter {
+  eventTarget: EventTarget;
+  listenerMapping: Record<string, (...args: any[]) => any>;
+
+  constructor(eventTarget: EventTarget) {
+    this.eventTarget = eventTarget;
+    this.listenerMapping = {};
+  }
+
+  on(eventName: string, listener: (...args: any[]) => void): void {
+    const listenerWrapper = (event: any): void => {
+      listener(event.data);
+    };
+    this.listenerMapping[listener.toString()] = listenerWrapper;
+    return this.eventTarget.addEventListener(eventName, listenerWrapper);
+  }
+
+  removeListener(eventName: string, listener: (...args: any[]) => void): void {
+    const listenerWrapper = this.listenerMapping[listener.toString()];
+    delete this.listenerMapping[listener.toString()];
+    return this.eventTarget.removeEventListener(eventName, listenerWrapper);
+  }
+
+  once(eventName: string, listener: (...args: any[]) => void): void {
+    const listenerWrapper = (event: any): void => {
+      listener(event.data);
+    };
+    this.listenerMapping[listener.toString()] = listenerWrapper;
+    return this.eventTarget.addEventListener(eventName, listenerWrapper, {
+      once: true,
+    });
+  }
+
+  emit(eventName: string, data: any): boolean {
+    return this.eventTarget.dispatchEvent(new _StripeEvent(eventName, data));
+  }
+}
+
+export = StripeEmitter;
--- a/src/StripeEmitter.ts
+++ b/src/StripeEmitter.ts
@@ -1,50 +1,27 @@
-/** @private
- * (For internal use in stripe-node.)
- * Wrapper around the Event Web API. */
-class _StripeEvent extends Event {
-  data: any;
-  constructor(eventName: string, data: any) {
-    super(eventName);
-    this.data = data;
-  }
-}
-
 /** Minimal EventEmitter wrapper around EventTarget. */
 class StripeEmitter {
   eventTarget: EventTarget;
-  listenerMapping: Record<string, (...args: any[]) => any>;
 
   constructor(eventTarget: EventTarget) {
     this.eventTarget = eventTarget;
-    this.listenerMapping = {};
   }
 
   on(eventName: string, listener: (...args: any[]) => void): void {
-    const listenerWrapper = (event: any): void => {
-      listener(event.data);
-    };
-    this.listenerMapping[listener.toString()] = listenerWrapper;
-    return this.eventTarget.addEventListener(eventName, listenerWrapper);
+    return this.eventTarget.addEventListener(eventName, listener);
   }
 
   removeListener(eventName: string, listener: (...args: any[]) => void): void {
-    const listenerWrapper = this.listenerMapping[listener.toString()];
-    delete this.listenerMapping[listener.toString()];
-    return this.eventTarget.removeEventListener(eventName, listenerWrapper);
+    return this.eventTarget.removeEventListener(eventName, listener);
   }
 
   once(eventName: string, listener: (...args: any[]) => void): void {
-    const listenerWrapper = (event: any): void => {
-      listener(event.data);
-    };
-    this.listenerMapping[listener.toString()] = listenerWrapper;
-    return this.eventTarget.addEventListener(eventName, listenerWrapper, {
+    return this.eventTarget.addEventListener(eventName, listener, {
       once: true,
     });
   }
 
   emit(eventName: string, data: any): boolean {
-    return this.eventTarget.dispatchEvent(new _StripeEvent(eventName, data));
+    return this.eventTarget.dispatchEvent(new Event(eventName, data));
   }
 }
 
--- a/src/StripeEmitter.ts
+++ b/src/StripeEmitter.ts
@@ -12,11 +12,12 @@ class _StripeEvent extends Event {
 }
 
 type Listener = (...args: any[]) => any;
+type ListenerWrapper = (event: _StripeEvent) => void;
 
 /** Minimal EventEmitter wrapper around EventTarget. */
 class StripeEmitter {
   eventTarget: EventTarget;
-  listenerMapping: Map<Listener, Listener>;
+  listenerMapping: Map<Listener, ListenerWrapper>;
 
   constructor() {
     this.eventTarget = new EventTarget();
@@ -24,7 +25,7 @@ class StripeEmitter {
   }
 
   on(eventName: string, listener: Listener): void {
-    const listenerWrapper = (event: _StripeEvent): void => {
+    const listenerWrapper: ListenerWrapper = (event: _StripeEvent): void => {
       listener(event.data);
     };
     this.listenerMapping.set(listener, listenerWrapper);
@@ -38,7 +39,7 @@ class StripeEmitter {
   }
 
   once(eventName: string, listener: Listener): void {
-    const listenerWrapper = (event: _StripeEvent): void => {
+    const listenerWrapper: ListenerWrapper = (event: _StripeEvent): void => {
       listener(event.data);
     };
     this.listenerMapping.set(listener, listenerWrapper);
--- a/src/StripeEmitter.ts
+++ b/src/StripeEmitter.ts
@@ -1,27 +1,50 @@
+/** @private
+ * (For internal use in stripe-node.)
+ * Wrapper around the Event Web API. */
+class _StripeEvent extends Event {
+  data: any;
+  constructor(eventName: string, data: any) {
+    super(eventName);
+    this.data = data;
+  }
+}
+
 /** Minimal EventEmitter wrapper around EventTarget. */
 class StripeEmitter {
   eventTarget: EventTarget;
+  listenerMapping: Record<string, (...args: any[]) => any>;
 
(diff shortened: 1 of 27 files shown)


test/PlatformFunctions.spec.ts
```diff
@@ -181,5 +184,59 @@ for (const platform in platforms) {
...
+
+        if (!isNodeEnvironment) {
+          before(() => {
+            oldCreateEmitter = platformFunctions.createEmitter;
+
+            // StripeEmitter uses the Event Web API. `createEmitter` needs
+            // to be defined here to avoid reference errors in Node versions <15.
+            platformFunctions.createEmitter = (): StripeEmitter => {
+              return new StripeEmitter(new EventTarget());
+            };
```
anniel-stripe, line 201: Is there a better way to share this `createEmitter` implementation with `stripe.worker.ts` without defining it in its own module? Should I just create a new module?

src/platform/DefaultPlatformFunctions.ts
```diff
@@ -38,6 +41,10 @@ class DefaultPlatformFunctions {
     }
     return result === 0;
   }
+
+  createEmitter(): StripeEmitter | EventEmitter {
+    throw new Error('createEmitter not implemented');
```
pakrym-stripe, line 46: Feels like this should return a new instance of StripeEmitter

src/stripe.common.ts
```diff
@@ -203,6 +202,7 @@ Stripe.prototype = {
   on: null!,
   off: null!,
   once: null!,
+  emit: null!,
```
pakrym-stripe, line 205: ?

src/StripeEmitter.ts
```diff
@@ -0,0 +1,51 @@
...
+class _StripeEvent extends Event {
+  data: any;
+  constructor(eventName: string, data: any) {
+    super(eventName);
+    this.data = data;
+  }
+}
+
+/** Minimal EventEmitter wrapper around EventTarget. */
+class StripeEmitter {
```
pakrym-stripe, line 13: Let's write some tests that run both on `StripeEmitter` and `EventEmiter`. We need to confirm that `StripeEmitter` and has the same behavior as `EventEmiter`.

src/StripeEmitter.ts
```diff
@@ -0,0 +1,51 @@
...
+  eventTarget: EventTarget;
+  listenerMapping: Record<string, (...args: any[]) => any>;
+
+  constructor(eventTarget: EventTarget) {
+    this.eventTarget = eventTarget;
+    this.listenerMapping = {};
+  }
+
+  on(eventName: string, listener: (...args: any[]) => void): void {
+    const listenerWrapper = (event: any): void => {
```
pakrym-stripe, line 23: can we have a better type than any?

src/stripe.worker.ts
```diff
@@ -1,8 +1,16 @@
...
+import StripeEmitter = require('./StripeEmitter');
 
 Stripe.createHttpClient = Stripe.createFetchHttpClient;
 Stripe.webhooks._createCryptoProvider = Stripe.createSubtleCryptoProvider;
 
+// StripeEmitter uses the Event Web API. `Event` is not available
+// in the global scope until Node 15, so we set `createEmitter` here
+// to avoid reference errors.
+Stripe._platformFunctions.createEmitter = (): StripeEmitter => {
+  return new StripeEmitter(new EventTarget());
```
pakrym-stripe, line 11: What fails if we move this logic to DefaultPlatformFunctions?

src/utils.ts
```diff
@@ -252,7 +251,13 @@ const utils = {
    */
   checkForStream: (obj: {file?: {data: unknown}}): boolean => {
     if (obj.file && obj.file.data) {
-      return obj.file.data instanceof EventEmitter;
+      return (
```
pakrym-stripe, line 254: Can streams of be represented in non-node environments?

src/stripe.worker.ts
pakrym-stripe: We might need a base type for platform functions and then 2 implementations: one for node and one for workers. In that way the node implementation won't have to inherit all dependencies of the worker implementation.

src/utils.ts
pakrym-stripe: Yeah, looks like there is something https://developer.mozilla.org/en-US/docs/Web/API/FileReader#events

pakrym-stripe (changes requested):
1. Let's refactor platform functions a bit so we don't have to push more logic into entry point.
2. Let's add some tests to confirm parity between event emitter implementations.
3. Let's decide what we want to do with file streams in non-node environments.

src/platform/DefaultPlatformFunctions.ts
anniel-stripe: Updated in latest commit - `createEmitter` returns `StripeEmitter` in `WebPlatformFunctions`

src/StripeEmitter.ts
anniel-stripe: Changed to `_StripeEvent` in https://github.com/stripe/stripe-node/pull/1668/commits/fafb866900d20f3001c12537d12305df140e8f20

src/stripe.worker.ts
anniel-stripe: > What fails if we move this logic to DefaultPlatformFunctions?
This used to fail because it imports `StripeEmitter`, where we extend `Event` for `_StripeEvent`. `Event` is not available globally until Node 15, so this failed for Node 12 and 14 tests.

I followed your suggestion and created a `PlatformFunctions` base type that has `NodePlatformFunctions` and `WebPlatformFunctions` implementations in https://github.com/stripe/stripe-node/pull/1668/commits/fafb866900d20f3001c12537d12305df140e8f20

src/StripeEmitter.ts
anniel-stripe: This is done in [`PlatformFunctions.spec.ts`](https://github.com/stripe/stripe-node/pull/1668/files#diff-f0d0bd6c151c9a6aff19b6a75d08af3bb6c01c410b4494b9dd1e7144481d0a45L14-L18)

src/platform/WebPlatformFunctions.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /** @override */
+  tryBufferData(data: any): Promise<any> {
+    if (
+      data.file.data.getReader &&
+      typeof data.file.data.getReader === 'function'
+    ) {
+      throw new Error(
+        'Uploading a file as a stream is not supported in non-Node environments. Please open or upvote an issue at github.com/stripe/stripe-node if you use this, detailing your use-case.'
```
anniel-stripe, line 50: Feel free to suggest a better error message!

src/StripeEmitter.ts
```diff
@@ -0,0 +1,53 @@
...
+  constructor(eventName: string, data: any) {
+    super(eventName);
+    this.data = data;
+  }
+}
+
+/** Minimal EventEmitter wrapper around EventTarget. */
+class StripeEmitter {
+  eventTarget: EventTarget;
+  listenerMapping: Record<string, (...args: any[]) => any>;
```
pakrym-stripe, line 17: nit: let's add an alias for `(...args: any[]) => any` and use it across the file.

pakrym-stripe, line 17: Instead of being a `Record<string` should it be a `Map<object, ...`? Calling toString on function objects seems a bit hacky.

src/Webhooks.ts
```diff
@@ -60,7 +60,7 @@ type WebhookObject = {
   ) => Promise<WebhookEvent>;
   generateTestHeaderString: (opts: WebhookTestHeaderOptions) => string;
   _createCryptoProvider: () => CryptoProvider | null;
-  _platformFunctions: DefaultPlatformFunctions | null;
+  _platformFunctions: WebPlatformFunctions | null;
```
pakrym-stripe, line 63:
```suggestion
  _platformFunctions: PlatformFunctions | null;
```

src/platform/PlatformFunctions.ts
```diff
@@ -1,13 +1,16 @@
...
  * implementations depend on the platform / JS runtime.
  */
-class DefaultPlatformFunctions {
+class PlatformFunctions {
   /**
    * Gets uname with Node's built-in `exec` function, if available.
    */
   getUname(): Promise<string | null> {
-    return Promise.resolve(null);
+    throw new Error('getUname not implemented.');
```
pakrym-stripe, line 13: why this change?

src/platform/WebPlatformFunctions.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Specializes WebPlatformFunctions using APIs available in Web workers.
+ */
+class WebPlatformFunctions extends PlatformFunctions {
+  /** @override */
+  getUname(): Promise<string | null> {
+    return Promise.resolve(null);
+  }
+
+  /** @override */
+  uuid4(): string {
```
pakrym-stripe, line 14: do we need a copy here?

src/platform/WebPlatformFunctions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    for (let i = 0; i < len; ++i) {
+      result |= a.charCodeAt(i) ^ b.charCodeAt(i);
+    }
+    return result === 0;
+  }
+
+  /** @override */
+  createEmitter(): StripeEmitter {
+    return new StripeEmitter(new EventTarget());
```
pakrym-stripe, line 40: Should we create EventTarget inside StripeEmitter ctor?

src/stripe.common.ts
```diff
@@ -53,11 +53,10 @@ Stripe.HttpClient = HttpClient.HttpClient;
 Stripe.HttpClientResponse = HttpClient.HttpClientResponse;
 
 import CryptoProvider = require('./crypto/CryptoProvider');
-import EventEmitter = require('events');
 Stripe.CryptoProvider = CryptoProvider;
 
-import DefaultPlatformFunctions = require('./platform/DefaultPlatformFunctions');
-Stripe._platformFunctions = new DefaultPlatformFunctions();
+import PlatformFunctions = require('./platform/PlatformFunctions');
+Stripe._platformFunctions = new PlatformFunctions();
```
pakrym-stripe, line 59: Do we need this? `PlatformFunctions` should never be instantiated. It's a "broken" implementation.

test/PlatformFunctions.spec.ts
```diff
@@ -2,20 +2,35 @@
...
+
+import {expect} from 'chai';
+
+let platforms: Record<string, PlatformFunctions>;
 
-const platforms = {
-  default: new DefaultPlatformFunctions(),
-  node: new NodePlatformFunctions(),
-};
+if (process.versions.node < '15') {
```
pakrym-stripe, line 14: Nice

test/PlatformFunctions.spec.ts
```diff
@@ -181,5 +196,158 @@ for (const platform in platforms) {
...
+        // WebPlatformFunctions
+        it('should throw an error in web environments if streaming data is provided', () => {
+          if (process.versions.node < '18') {
+            console.log(
+              `'ReadableStream' is not available in the global scope for ${process.version}, skipping test.`
+            );
+            return;
+          }
+
+          const f = new ReadableStream();
```
pakrym-stripe, line 293: This won't work on node as well. Should throw everywhere.

pakrym-stripe (changes requested):
few more nits

src/platform/WebPlatformFunctions.ts
anniel-stripe: Oh good point, I missed this during the refactor

src/StripeEmitter.ts
anniel-stripe: This is discouraged by TypeScript, see the following linting error:
```
Don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
Consider using `Record<string, unknown>` instead, as it allows you to more easily inspect and use the keys.eslint[@typescript-eslint/ban-types](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/ban-types.md)
```

According to [this SO post](https://stackoverflow.com/a/10858667), it looks like everything gets turned into a string anyways under the hood, so I'd rather be explicit here.

test/PlatformFunctions.spec.ts
anniel-stripe: This test doesn't pass in node, but actually I don't think it's worth checking / throwing if the user tries to pass in a ReadableStream. We haven't had any issues / user requests for it, and it would add an additional, unnecessary check in `tryBufferData`.

src/StripeEmitter.ts
pakrym-stripe: https://stackoverflow.com/questions/10858632/are-functions-valid-keys-for-javascript-object-properties talks about object properties, I was wondering if we should use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map 
>Any value (both objects and [primitive values](https://developer.mozilla.org/en-US/docs/Glossary/Primitive)) may be used as either a key or a value

pakrym-stripe:
```
const makeFunc = () => {
    const time = Date.now()
    return () => console.log(time);
}
const a = makeFunc();
const b = makeFunc();

const dict: Record<string, any> = {};
dict[a.toString()] = a;
dict[b.toString()] = b;

Object.values(dict).forEach(f=>f());
```
illustrates the problem with using function.toString()

anniel-stripe: Ah sorry, I misunderstood! Thanks for the example, `Map` is definitely what we want. :)

test/PlatformFunctions.spec.ts
pakrym-stripe: I'll leave this up to you but we both know what the behavior is when passing a `ReadableStream` on node and it's bad.

src/StripeEmitter.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.data = data;
+  }
+}
+
+type Listener = (...args: any[]) => any;
+
+/** Minimal EventEmitter wrapper around EventTarget. */
+class StripeEmitter {
+  eventTarget: EventTarget;
+  listenerMapping: Map<Listener, Listener>;
```
pakrym-stripe, line 19: the value type can be more precise `(event: _StripeEvent): void`

src/platform/NodePlatformFunctions.ts
```diff
@@ -80,6 +84,39 @@ class NodePlatformFunctions extends DefaultPlatformFunctions {
 
     return super.secureCompare(a, b);
   }
+
+  createEmitter(): EventEmitter {
+    return new EventEmitter();
+  }
+
+  /** @override */
+  tryBufferData(data: StreamingFile): Promise<any> {
```
pakrym-stripe, line 93: nit: Didn't notice it the first time. `Promise<any>`... Boooo!

src/stripe.common.ts
```diff
@@ -53,11 +53,10 @@ Stripe.HttpClient = HttpClient.HttpClient;
 Stripe.HttpClientResponse = HttpClient.HttpClientResponse;
 
 import CryptoProvider = require('./crypto/CryptoProvider');
-import EventEmitter = require('events');
 Stripe.CryptoProvider = CryptoProvider;
 
-import DefaultPlatformFunctions = require('./platform/DefaultPlatformFunctions');
-Stripe._platformFunctions = new DefaultPlatformFunctions();
+// @ts-ignore
+Stripe._platformFunctions = null;
```
pakrym-stripe, line 59: nit: do we need this setter at all?

pakrym-stripe (approved):

src/stripe.common.ts
anniel-stripe: It's just to appease TypeScript 😅 I think functionally it makes more sense to remove this and add a few more `@ts-ignore`'s

anniel-stripe: Do you have a preference?

pakrym-stripe: Let's keep it and try to clean up the entire pattern in a follow up

test/PlatformFunctions.spec.ts
anniel-stripe: I agree it's bad since the `ReadableStream` gets stringified, but I think this puts too many guard rails in our library for something users are unlikely to do. Also, `ReadableStream` is not available in Node <18, so we'd have to use a duck-typing check to see if the file data looks like a `ReadableStream`. 🦆

pakrym-stripe (approved):

PR #1672, opened 2023-02-03
Update main entrypoint on package.json

giannisp:
Fix for the `main` entrypoint on package.json.
Noticed it when ESLint was complaining about not being able to find the module on path since `11.9.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "engines": {
     "node": ">=12.*"
   },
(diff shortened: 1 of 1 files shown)


anniel-stripe (approved):
Thanks so much for catching this and the quick contribution!

anniel-stripe:
Released in https://github.com/stripe/stripe-node/releases/tag/v11.9.1

PR #1679, opened 2023-02-07
Move all platform-specific implementations into PlatformFunctions

anniel-stripe:
## Summary
1. Creates a `createStripe` and `createWebhooks` function that takes in platform functions as a parameter, removing the need for a global `_platformFunctions` field.
2. Move all platform-dependent logic (including `create...CryptoProvider` and `create...HttpClient` into `PlatformFunctions`).
   a. l removed `createHttpClient` from Stripe and `_createCryptoProvider` from `Webhook`, since we can't have a "default" implementation as `create...CryptoProvider` and `create...HttpClient` take different types of arguments. The user will have to keep specifying `httpClient` on the stripe config and `cryptoProvider` in `constructEvent`/ `constructEventAsync`.

## Testing
Added tests in `PlatformFunctions.spec.ts` and removed no-op tests from `stripe.spec.ts`. Also ran manual end-to-end tests in a Cloudflare Pages Functions and Cloudflare Worker project.


src/Webhooks.ts
```diff
@@ -58,297 +58,311 @@ type WebhookObject = {
...
-  const details = parseHeader(decodedHeader, expectedScheme);
+  function validateComputedSignature(
+    payload: WebhookPayload,
+    header: WebhookHeader,
+    details: WebhookParsedHeader,
+    expectedSignature: string,
+    tolerance: number
+  ): boolean {
+    const signatureFound = !!details.signatures.filter(
+      platformFunctions!.secureCompare.bind(
```
pakrym-stripe, line 293: why do you still need the `!` here?

src/platform/NodePlatformFunctions.ts
```diff
@@ -119,6 +125,17 @@ class NodePlatformFunctions extends PlatformFunctions {
...
   }
+
+  /** @override */
+  createNodeHttpClient(agent: http.Agent): typeof HttpClient {
+    // @ts-ignore
+    return new NodeHttpClient(agent);
+  }
+
+  /** @override */
+  createNodeCryptoProvider(): StripeCryptoProvider {
```
pakrym-stripe, line 136: nit: Isn't StripeCryptoProvider a class can we reference it the same way we do HttpClient?

pakrym-stripe (approved):
Mostly nits + let's try and make the library work without customers providing extra parameters.

pakrym-stripe (changes requested):
...

src/platform/NodePlatformFunctions.ts
anniel-stripe: `StripeCryptoProvider` is the type defined in https://github.com/stripe/stripe-node/blob/5c444308d744bb806b50f915abd1255c0bd17434/src/Types.d.ts#L91, but after seeing this I want to try defining a `StripeHttpClient` type so we don't have to import so many things

src/Webhooks.ts
anniel-stripe: that's a great question 😬

test/stripe.spec.ts
```diff
@@ -156,30 +156,6 @@ describe('Stripe Module', function() {
     });
   });
```
pakrym-stripe, line 157: Wasn't there a test where we reached into Stripe and set the _platformFunctions field directly?

src/platform/NodePlatformFunctions.ts
pakrym-stripe: We should do the opposite, use as much of existing types as possible. The goal is to remove Types.d.ts it's a temporary hack.

test/stripe.spec.ts
anniel-stripe: It's the uname tests: https://github.com/stripe/stripe-node/blob/663f7dbc1c67b968358d8226b78d8f19b45884d0/test/stripe.spec.ts#L271

But in `stripe.common` we still access this implementation through `_platformFunctions` on the instance: https://github.com/stripe/stripe-node/pull/1679/files/ef4162c1b220a166df91bd7151896128188cdb0f#diff-9ed8c83003262337d25a9d027e4591a600743fca72db01a1b1ef02ff084fae74R347

src/platform/NodePlatformFunctions.ts
anniel-stripe: Gotcha, sounds good!

test/stripe.spec.ts
pakrym-stripe: Right, but with your new factory function you can create a new instance of stripe and provide `getMockPlatformFunctions()..` as parameter.

anniel-stripe: Updated in https://github.com/stripe/stripe-node/pull/1679/commits/b43f1fb28dac6ae470e08feedc998e1aafbe1eee

src/platform/NodePlatformFunctions.ts
anniel-stripe: Replaced with `CryptoProvider` in https://github.com/stripe/stripe-node/pull/1679/commits/b43f1fb28dac6ae470e08feedc998e1aafbe1eee

src/platform/NodePlatformFunctions.ts
```diff
@@ -119,6 +126,28 @@ class NodePlatformFunctions extends PlatformFunctions {
         });
     });
   }
+
+  /** @override */
+  createNodeHttpClient(agent: http.Agent): typeof HttpClient {
+    // @ts-ignore
```
pakrym-stripe, line 132: why ignore?

src/platform/NodePlatformFunctions.ts
```diff
@@ -119,6 +126,28 @@ class NodePlatformFunctions extends PlatformFunctions {
...
+    return new NodeHttpClient(agent);
+  }
+
+  /** @override */
+  createHttpClient(agent?: http.Agent): typeof HttpClient {
```
pakrym-stripe, line 137: let's call this createDefaultHttpClient

pakrym-stripe, line 137: We shouldn't have the `http.Agent` parameter here. It's a node specific type.

src/platform/PlatformFunctions.ts
```diff
@@ -56,6 +71,53 @@ class PlatformFunctions {
   ): Promise<RequestData | BufferedFile> {
     throw new Error('tryBufferData not implemented.');
   }
+
+  /**
+   * Creates an HTTP client which uses the Node `http` and `https` packages
+   * to issue requests.
+   */
+  createNodeHttpClient(agent?: http.Agent | null): typeof HttpClient {
```
pakrym-stripe, line 79: I'm surprised cloudflare doesn't complain about this use of Node `http.Agent`

src/platform/WebPlatformFunctions.ts
```diff
@@ -26,6 +29,30 @@ class WebPlatformFunctions extends PlatformFunctions {
...
+
+  /** @override */
+  createNodeCryptoProvider(): CryptoProvider {
+    throw new Error(
+      'Stripe: `createNodeCryptoProvider()` is not available in non-Node environments. Please use `createSubtleCryptoProvider()` instead.'
+    );
+  }
+
+  /** @override */
+  createCryptoProvider(): CryptoProvider {
```
pakrym-stripe, line 53: Let's call this createDefaultCryptoProvider

src/stripe.node.ts
```diff
@@ -1,24 +1,7 @@
...
-Stripe.createNodeHttpClient = createNodeHttpClient;
-Stripe.createHttpClient = createNodeHttpClient;
-
-const createNodeCryptoProvider = (): StripeCryptoProvider => {
-  const NodeCryptoProvider = require('./crypto/NodeCryptoProvider');
-  return new NodeCryptoProvider();
-};
-Stripe.createNodeCryptoProvider = createNodeCryptoProvider;
-Stripe.webhooks._createCryptoProvider = createNodeCryptoProvider;
+const Stripe = _Stripe.createStripe(new NodePlatformFunctions());
```
pakrym-stripe, line 4: So clean!

src/platform/NodePlatformFunctions.ts
anniel-stripe: I get a this error:
```
Type 'NodeHttpClient' is missing the following properties from type 'typeof HttpClient': prototype, CONNECTION_CLOSED_ERROR_CODES, TIMEOUT_ERROR_CODE, makeTimeoutError
```
`CONNECTION_CLOSED_ERROR_CODES`, `TIMEOUT_ERROR_CODE`, and `makeTimeoutError` are defined as static properties / methods on `HttpClient`, but I tried and failed at diagnosing why they were't being detected by the linter.

src/platform/PlatformFunctions.ts
anniel-stripe: The type / import goes away when transpiled to JS

src/platform/NodePlatformFunctions.ts
anniel-stripe: This is in `NodePlatformFunctions` - is the concern because the web implementation also inherits from `PlatformFunctions` which takes `agent?: http.Agent | null`?

pakrym-stripe: For `createDefaultHttpClient` the concern is that `agent?: http.Agent | null` doesn't apply to all implementations. it's ok to keep it on the createNodeHttpClient method where it makes sense.

anniel-stripe: Fixed in latest commit as discussed!

src/platform/NodePlatformFunctions.ts
anniel-stripe: Since this is due to how we import `HttpClient` and `NodeHttpClient` and can be fixed with ES modules, I'm going to wait to fix this when we do the ESM refactor.

PR #1682, opened 2023-02-08
Don't auto update api version in samples

pakrym-stripe:
This fails because we re-arranged samples but we shouldn't update them automatically as they reference the package not the local code.

The misalignment will be caught by TS compiler when we update the package version used in tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@
 update-version:
 	@echo "$(VERSION)" > VERSION
 	@perl -pi -e 's|"version": "[.\-\d\w]+"|"version": "$(VERSION)"|' package.json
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,7 +2,6 @@
 update-version:
 	@echo "$(VERSION)" > VERSION
 	@perl -pi -e 's|"version": "[.\-\d\w]+"|"version": "$(VERSION)"|' package.json
-	@find . -name 'package.json' -exec perl -pi -e 's|"stripe": "[\^\d\w.]*?"|"stripe": "^$(VERSION)"|' {} +
```
pakrym-stripe, line 5: I realized this is a bad idea, the package won't be available between the time we update the version and release the actual package to NPM. If we update samples during that time we'll fail the build.

anniel-stripe (approved):

PR #1687, opened 2023-02-22
Convert TypeScript files to use ES modules

anniel-stripe:
## Summary
This PR updates our TypeScript source files to use ESM syntax to get better type safety. I opted to use named exports in our internal files to make it more obvious what is being exported / imported. This also updates CJS imports (`require(...)`) as needed in generated resource files and tests to use those named exports.

## Testing
Functionality of transpiled output (in `lib/`) should stay the same. Verified in manual test projects that named and default exports still work when `stripe` is imported using `import` or `require`.


src/StripeResource.ts
```diff
@@ -1,10 +1,14 @@
...
+  protoExtend,
+  stringifyRequestData,
+} from './utils.js';
+import {stripeMethod} from './StripeMethod.js';
 
 // Provide extension mechanism for Stripe Resource Sub-Classes
-StripeResource.extend = utils.protoExtend;
-
-// Expose method-creator
-StripeResource.method = require('./StripeMethod');
```
pakrym-stripe, line 7: Is this method visible through `stripe` instance? Like `Stripe.StripeResource.method`?

pakrym-stripe: Something doesn't have to be exported directly to be visible to customers and we should make sure things are not broken.

anniel-stripe: Yes that was visible on `StripeResource`, adding it back. Thanks for catching that!

src/RequestSender.ts
```diff
@@ -1,19 +1,22 @@
...
-} = _Error;
-
-const {HttpClient} = require('./net/HttpClient');
+} from './Error.js';
+import {
+  emitWarning,
+  normalizeHeaders,
+  removeNullish,
+  stringifyRequestData,
+} from './utils.js';
```
pakrym-stripe, line 14: Why is there `js` extension everywhere? Shouldn't it work without?

src/Types.d.ts
```diff
@@ -155,7 +155,7 @@ type StripeObject = {
   _requestSender: RequestSender;
   _getPropsFromConfig: (config: Record<string, unknown>) => UserProvidedConfig;
   _clientId?: string;
-  _platformFunctions: import('./platform/PlatformFunctions');
+  _platformFunctions: import('./platform/PlatformFunctions').PlatformFunctions;
```
pakrym-stripe, line 158: Should we do a real import on top?

src/platform/PlatformFunctions.ts
```diff
@@ -76,7 +74,7 @@ class PlatformFunctions {
    * Creates an HTTP client which uses the Node `http` and `https` packages
    * to issue requests.
    */
-  createNodeHttpClient(agent?: http.Agent | null): typeof HttpClient {
+  createNodeHttpClient(agent: http.Agent): HttpClient {
```
pakrym-stripe, line 77: Isn't this parameter optional?

src/platform/PlatformFunctions.ts
```diff
@@ -87,15 +85,15 @@ class PlatformFunctions {
    * A fetch function can optionally be passed in as a parameter. If none is
    * passed, will default to the default `fetch` function in the global scope.
    */
-  createFetchHttpClient(fetchFn?: typeof fetch | null): typeof HttpClient {
+  createFetchHttpClient(fetchFn?: typeof fetch | null): HttpClient {
     // @ts-ignore
```
pakrym-stripe, line 89: Still ts-ignore?

pakrym-stripe (changes requested):
Few comments

src/RequestSender.ts
anniel-stripe: It does work without at the moment, but when we add `type: "module"` to our package.json later we'll need this because [file extensions are mandatory in ESM](https://nodejs.org/api/esm.html#esm_mandatory_file_extensions), and the TypeScript compiler will also [need these explicit .js extensions](https://github.com/nodejs/node/issues/46006) when targeting ESM. However now that I think about it, we should add these extensions later in a separate PR when we actually need them.

src/Types.d.ts
anniel-stripe: Moving these to the top [turns this into a local module s](https://stackoverflow.com/a/51114250) so we'd have to export / import types where they're needed (which might be what we want because it'll be easier to eventually get rid of this file later). I think that's a bigger lift than should fit in this PR, but if you think that's a good idea I'm happy to log this as a follow-up!

pakrym-stripe: Fine to do it later. We do need to clean up this file. Ideally it should go away completely.

src/RequestSender.ts
pakrym-stripe: Can the extension be `.ts`?

src/platform/PlatformFunctions.ts
anniel-stripe: Yes it is 😅 Fixed

src/RequestSender.ts
pakrym-stripe: [No.](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937)

src/platform/PlatformFunctions.ts
anniel-stripe: Fixed

pakrym-stripe (approved):
Awesome!

PR #1695, opened 2023-02-28
Migrate generated files to ES module syntax

anniel-stripe:
## Summary
Migrates generated files ( `resources/*`, `resources`, `apiVersion`) to ESM syntax, and changes file endings to .ts

This PR also cleans up some remaining CJS syntax in our non-generated, infrastructure files (except for the `module.exports...` in our entrypoint files).

## Testing
Functionality of transpiled output (in `lib/`) should stay the same. Verified locally that test Node.js and Cloudflare worker projects still behave as expected.


src/Types.d.ts
```diff
@@ -241,3 +240,14 @@ type UserProvidedConfig = {
   telemetry?: boolean;
   appInfo?: AppInfo;
 };
+type OAuthAuthorizeUrlParams = {
```
pakrym-stripe, line 243: can we define these in the file where they are used?

pakrym-stripe (approved):
Awesome work, few nits

src/Types.d.ts
anniel-stripe: Done!

pakrym-stripe (approved):

PR #1699, opened 2023-03-03
Remove extra test directory

pakrym-stripe:
We don't really need another test*** directory. Move testUtils into tests and remove unused polyfils.


anniel-stripe (changes requested):
Looks like these polyfills are imported in autopagination tests, but since we now support Node 12+ we can get rid of those?

anniel-stripe (approved):

PR #1704, opened 2023-03-09
Configure 2 TypeScript compile targets for ESM and CJS

anniel-stripe:
## Summary
Convert stripe-node to a [dual package](https://nodejs.org/api/packages.html#dual-commonjses-module-packages) supporting both CommonJS and ES Modules.

In this PR:
* Configure 2 TypeScript builds for CJS and ESM, respectively, to `cjs` and `esm` (`lib` is removed)
* Update entry points to point to the CommonJS entry point defined by `src/stripe.cjs.*..ts`, will add ESM entry point in a follow-up PR
* Rename `stripe.common.ts` to `stripe.core.ts` because the naming is too similar to CommonJS
* Add `Stripe` and `default` properties to Stripe object in `stripe.core.ts` for backwards compatibility
* Update relative imports to specify `.js` file ending
* Set package version directly in `stripe.core.ts` and add rule to update that version during releases.
* Update paths in test files to point to `cjs/` output

## Testing
In addition to CI tests, verified that package works as expected in manual test projects


package.json
```diff
@@ -59,8 +59,11 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "tsc -p tsconfig.json",
-    "clean": "rm -rf ./.nyc_output ./node_modules/.cache ./coverage ./lib",
+    "build": "yarn build-esm && yarn build-cjs && yarn rename-entrypoints && yarn lint --fix",
+    "build-esm": "tsc -p tsconfig.esm.json && echo '{\"type\":\"commonjs\"}' > cjs/package.json",
+    "build-cjs": "tsc -p tsconfig.cjs.json && echo '{\"type\":\"module\"}' > esm/package.json",
+    "rename-entrypoints": "mv cjs/stripe.cjs.worker.js cjs/stripe.worker.js && mv cjs/stripe.cjs.node.js cjs/stripe.node.js && mv esm/stripe.esm.worker.js esm/stripe.worker.js && mv esm/stripe.esm.node.js esm/stripe.node.js",
```
anniel-stripe, line 65: Is this too hacky? I struggled to find a better alternative that works, unless we're OK with keeping the name `cjs/stripe.cjs.node.js` or something similarly verbose

package.json
```diff
@@ -59,8 +59,11 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "tsc -p tsconfig.json",
-    "clean": "rm -rf ./.nyc_output ./node_modules/.cache ./coverage ./lib",
+    "build": "yarn build-esm && yarn build-cjs && yarn rename-entrypoints && yarn lint --fix",
```
anniel-stripe, line 62: Added `yarn lint --fix` here because for some reason [prettier got really mad during the linting CI step](https://github.com/stripe/stripe-node/actions/runs/4379461291), even though it didn't before for `lib/`? Not sure what changed to cause it

src/createStripe.ts
```diff
@@ -1,16 +1,17 @@
-import * as _Error from './Error';
```
anniel-stripe, line 1: I renamed this file because stripe.common.ts compiles to stripe.common.js, which I think is confusing in the context of a dual package

src/createStripe.ts
```diff
@@ -61,6 +62,8 @@ export function createStripe(
   Stripe.HttpClient = HttpClient;
   Stripe.HttpClientResponse = HttpClientResponse;
   Stripe.CryptoProvider = CryptoProvider;
+  Stripe.Stripe = Stripe;
```
anniel-stripe, line 65: I added this to maintain backwards compatibility in the ESM entrypoint, because these properties existed on the object the user imported. The way we export right now (`export const Stripe = ...` and `export default Stripe`) in the ESM entrypoint no longer introduces those properties through `module.exports`.

test/PlatformFunctions.spec.ts
```diff
@@ -5,12 +5,12 @@ require('./testUtils.js');
 import * as fs from 'fs';
 import * as path from 'path';
 
-import {FetchHttpClient} from '../lib/net/FetchHttpClient';
-import {NodeCryptoProvider} from '../lib/crypto/NodeCryptoProvider';
-import {NodeHttpClient} from '../lib/net/NodeHttpClient';
-import {NodePlatformFunctions} from '../lib/platform/NodePlatformFunctions';
-import {PlatformFunctions} from '../lib/platform/PlatformFunctions';
-import {SubtleCryptoProvider} from '../lib/crypto/SubtleCryptoProvider';
+import {FetchHttpClient} from '../cjs/net/FetchHttpClient';
```
anniel-stripe, line 8: Do we want to duplicate the test suite for `esm/` files? I haven't done this yet since this would double the size of our tests when the underlying functionality is the same, and the main issues we're worried about with the additional ESM entry point should be captured in integration tests (which I'll add to as needed).

tsconfig.esm.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
```
pakrym-stripe, line 2: Let's use [config inheritance](https://www.typescriptlang.org/tsconfig#extends) here instead of copying all the settings

package.json
pakrym-stripe: > We could also get rid of this logic if we're OK with keeping the name cjs/stripe.cjs.node.js 

Repetitive names don't bother me.

package.json
pakrym-stripe: we shouldn't be linting or formatting generated code anymore. I think you missed some ignore/exclusion declaration somewhere.

test/PlatformFunctions.spec.ts
pakrym-stripe: > Do we want to duplicate the test suite for esm/ files? 

No. For the exact reason you outlined.

src/createStripe.ts
pakrym-stripe: Let's add a comment because it will look strange for a future reader. And maybe consider removing at least for ESM in the next major

test/utils.spec.ts
```diff
@@ -2,7 +2,7 @@
 
 require('./testUtils.js');
 
-import * as utils from '../lib/utils';
+import * as utils from '../cjs/utils';
```
pakrym-stripe, line 5: I wonder if we should start importing TS files instead of compiled output. Maybe as a follow up.

src/stripe.esm.worker.ts
```diff
@@ -0,0 +1,5 @@
+import {WebPlatformFunctions} from './platform/WebPlatformFunctions.js';
+import {createStripe} from './createStripe.js';
+
+export const Stripe = createStripe(new WebPlatformFunctions());
+export default Stripe;
```
pakrym-stripe, line 5: is this for back compat as well?

src/createStripe.ts
pakrym-stripe: Not a fan of the `createStripe` name tbh :) `stripe.core`? `stripeClient`? `stripeFactory`? Not a strong opinion.

test/PlatformFunctions.spec.ts
pakrym-stripe: We might want to covert tests to import from ts files. That way tests can be compiled to be ESM or CJS based.

pakrym-stripe (changes requested):
Nice work! Mostly nits and discussions.

package.json
anniel-stripe: Ah yep, I missed the .eslintignore 🤦‍♀️

src/createStripe.ts
anniel-stripe: Filed an issue to remove these in the next major!

src/createStripe.ts
anniel-stripe: I'm a fan of stripe.core

cjs/RequestSender.js
```diff
@@ -0,0 +1,461 @@
+'use strict';
+
+Object.defineProperty(exports, '__esModule', {value: true});
```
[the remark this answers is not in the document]
anniel-stripe: The only difference when I compare this with the old `lib/RequestSender.js` is that variables have been renamed to include the file ending, i.e. `Error_1` becomes `Error_js_1`

tsconfig.esm.json
anniel-stripe: I grouped them into a `configs/` - let me know if you'd prefer to organize these in a different way (an alternative is tsconfig.esm.json directly extends tsconfig.cjs.json)

test/utils.spec.ts
anniel-stripe: I agree this would be ideal! I've added a task to track this. When I tried to do this, I didn't have a lot of luck and ended up wrestling with mocha and a series of seemingly unrelated error messages for a few days 😓

test/PlatformFunctions.spec.ts
anniel-stripe: Spoke offline, this is out of scope for this PR and has been logged as a follow-up task

configs/tsconfig.json
```diff
@@ -1,9 +1,6 @@
 {
   "compilerOptions": {
```
pakrym-stripe, line 2: Does all the tooling know to look for config here?

.npmignore
```diff
@@ -9,5 +9,5 @@
 !/README.md
 !/VERSION
 !/package.json
-!/lib/**/*
```
pakrym-stripe, line 12: + esm?

.eslintrc.js
```diff
@@ -244,6 +244,12 @@ module.exports = {
   plugins: ['prettier'],
   extends: ['plugin:prettier/recommended'],
   overrides: [
+    {
+      files: ['esm/**/*.js'],
```
pakrym-stripe, line 248: why don't we need the same thing for esm?

pakrym-stripe: pretty sure we don't need this at all.

configs/tsconfig.json
pakrym-stripe: Do we need 3 configs? Can we make the esm one the base?

pakrym-stripe (approved):
nits

.npmignore
anniel-stripe: Updated! My original thought process was that since we're defining both entrypoints yet we should just expose cjs/, but it would be weird to have things referencing `esm/` in the package but not include it.

.eslintrc.js
anniel-stripe: You're right, that was left over from an earlier iteration

configs/tsconfig.json
anniel-stripe: Also you're right about tooling, I moved the tsconfigs to the root. I made the cjs one the base because mocha does not play well when module is set to "es2022" in the tsconfig, and because I think it makes sense that the base tsconfig is the same as what we had previously.

pakrym-stripe (approved):
Great work!

PR #1709, opened 2023-03-14
Add ES module package entry point

anniel-stripe:
## Summary
Convert stripe-node to a [dual package](https://nodejs.org/api/packages.html#dual-commonjses-module-packages) supporting both CommonJS and ES Modules.

Follow-up to https://github.com/stripe/stripe-node/pull/1704

## Testing
In addition to our test projects from CI, I've run manual test projects to verify they work as before, and that they use the correct entry point. Verified the following:
* Works with CommonJS require `const Stripe = require('stripe')` and `const {Stripe} = require('stripe')`)
* Works with ES module import (`import Stripe from 'stripe'` and `import {Stripe} from 'stripe'`)
  * TypeScript, plain JS
  * Node and worker targets
* CJS / ESM entrypoints bundle successfully with esbuild and webpack for node and browser targets
* Type hints still work in IDE

## Changelog
* Add support for ES modules by defining a separate ESM entry point. This updates stripe-node to be a [dual CommonJS / ES module package](https://nodejs.org/api/packages.html#dual-commonjses-module-packages).


pakrym-stripe (approved):
🥳

PR #1720, opened 2023-03-17
Require `.js` extension via linter rule

richardm-stripe:


anniel-stripe (approved):
Ooh thanks for adding this rule! I wish I'd discovered it earlier it would have made my life a whole lot easier 😅

PR #1739, opened 2023-03-31
Refactor src/autoPagination.ts

richardm-stripe:
## Summary
Refactors src/autoPagination.ts to create a clearer division between "define a .next() style iterator" step and "given an implementation of a .next()-style iterator, produce .autoPagingEach and .autoPagingToArray methods" step.

To me this makes this file more readable and easier to extend.

### DIFF
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -63,9 +63,7 @@ class StripeIterator<T> implements IStripeIterator<T> {
     this.stripeResource = stripeResource;
   }
 
-  async iterate(
-    pageResult: PageResult<T>
-  ): Promise<IterationResult<T>> {
+  async iterate(pageResult: PageResult<T>): Promise<IterationResult<T>> {
     if (
       !(
         pageResult &&
@@ -92,7 +90,7 @@ class StripeIterator<T> implements IStripeIterator<T> {
       this.index = 0;
       this.pagePromise = this.getNextPage(pageResult);
       const nextPageResult = await this.pagePromise;
-      return this.iterate(nextPageResult)
+      return this.iterate(nextPageResult);
     }
     // TODO (next major) stop returning explicit undefined
     return {value: undefined, done: true};
@@ -103,8 +101,8 @@ class StripeIterator<T> implements IStripeIterator<T> {
   }
 
   private async _next() {
-    return this.iterate(await this.pagePromise)
-  };
+    return this.iterate(await this.pagePromise);
+  }
 
   next(): Promise<IterationResult<T>> {
     /**
@@ -116,7 +114,9 @@ class StripeIterator<T> implements IStripeIterator<T> {
       return this.promiseCache.currentPromise;
     }
 
-    const nextPromise = this._next().then((x) => (this.promiseCache.currentPromise = null, x))
+    const nextPromise = this._next().then(
+      (x) => ((this.promiseCache.currentPromise = null), x)
+    );
     this.promiseCache.currentPromise = nextPromise;
 
     return nextPromise;
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -63,9 +63,9 @@ class StripeIterator<T> implements IStripeIterator<T> {
     this.stripeResource = stripeResource;
   }
 
-  iterate(
+  async iterate(
     pageResult: PageResult<T>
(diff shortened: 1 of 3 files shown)


src/autoPagination.ts
```diff
@@ -4,82 +4,68 @@ import {callbackifyPromiseWithTimeout, getDataFromArgs} from './utils.js';
...
-type IterationResult = {
-  done: boolean;
-  value?: any;
-};
+type IterationResult<T> =
+  | {
+      done: false;
+      value: T;
+    }
+  | {done: true; value?: undefined};
```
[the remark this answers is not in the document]
pakrym-stripe: or 
```suggestion
  | {done: true; };
```
?

src/autoPagination.ts
```diff
@@ -4,82 +4,68 @@ import {callbackifyPromiseWithTimeout, getDataFromArgs} from './utils.js';
...
-      });
-    };
-  } else {
-    getNextPagePromise = (pageResult): Promise<any> => {
-      const lastId = getLastId(pageResult, reverseIteration);
-      return self._makeRequest(requestArgs, spec, {
-        [reverseIteration ? 'ending_before' : 'starting_after']: lastId,
-      });
-    };
+interface IStripeIterator<T> {
```
pakrym-stripe, line 37: The interface is used in a few places. Do we need it or can we pass the `StripeIterator` around?

src/autoPagination.ts
```diff
@@ -4,82 +4,68 @@ import {callbackifyPromiseWithTimeout, getDataFromArgs} from './utils.js';
...
+  next: () => Promise<IterationResult<T>>;
+}
+type PageResult<T> = {
+  data: Array<T>;
+  has_more: boolean;
+  next_page: string | null;
+};
+class StripeIterator<T> implements IStripeIterator<T> {
+  private i: number;
```
pakrym-stripe, line 46: now that it's a field
```suggestion
  private index: number;
```
?

src/autoPagination.ts
```diff
@@ -92,39 +78,97 @@ export function makeAutoPaginationMethods(
...
 
       return {value, done: false};
     } else if (pageResult.has_more) {
       // Reset counter, request next page, and recurse.
-      i = 0;
-      pagePromise = getNextPagePromise(pageResult);
-      return pagePromise.then(iterate);
+      this.i = 0;
+      this.pagePromise = this.getNextPage(pageResult);
+      return this.pagePromise.then((pageResult) => this.iterate(pageResult));
```
pakrym-stripe, line 94: can we use `await` instead of callback chaining?

src/autoPagination.ts
```diff
@@ -92,39 +78,97 @@ export function makeAutoPaginationMethods(
...
-      return pagePromise
-        .then(iterate)
+  getNextPage(_pageResult: PageResult<T>): Promise<PageResult<T>> {
+    throw new Error('Unimplemented');
+  }
+
+  next(): Promise<IterationResult<T>> {
+    return memoizedPromise(this.promiseCache, (resolve, reject) => {
+      return this.pagePromise
+        .then((pageResult) => this.iterate(pageResult))
```
pakrym-stripe, line 106: Same question about async/await here.

pakrym-stripe (approved):

src/autoPagination.ts
richardm-stripe: Ultimately yes, but if I made `iterate` an async function then the behavior of `throw` on line 76 would change. It would probably change for the better (it should be rejecting a promise instead of throwing outside the promise chain) but I wanted to keep this PR a true refactor.

src/autoPagination.ts
richardm-stripe: I don't think so? It feels wrong to make the `(resolve, reject) => ...` callback async, because the type of that function (the callback accepted by memoizedPromise) is
```
(resolve: (value: T) => void, reject: (reason?: any) => void) => void
```
and I think you should only make functions async if their return type is a promise.

Higher-level, I suspect there might be a cleaner way to do all of whatever this memoizedPromise stuff is doing, but detangling that seems pretty involved and I didn't want that to be in scope for this PR.

src/autoPagination.ts
richardm-stripe: Done!

pakrym-stripe (approved):

PR #1745, opened 2023-04-05
Use standard Typescript AsyncIterator types

richardm-stripe:
This is a follow-up to https://github.com/stripe/stripe-node/pull/1739#discussion_r1156111671.

I was originally going to make the iterator return `{done: true}` instead of `{done: true, value: undefined}` (which would have been a breaking change), but it turns out that `value: undefined` is actually the standard:

```node
> [][Symbol.iterator]().next()
{ value: undefined, done: true }
```
And it turns out that we can use the standard Typescript types for iterators/async iterators instead of defining our own. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.iterable.d.ts#L11-L28 
Doing this is nonbreaking.

### DIFF
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -4,12 +4,6 @@ import {callbackifyPromiseWithTimeout, getDataFromArgs} from './utils.js';
 type PromiseCache = {
   currentPromise: Promise<any> | undefined | null;
 };
-type IterationResult<T> =
-  | {
-      done: false;
-      value: T;
-    }
-  | {done: true};
 type IterationDoneCallback = () => void;
 type IterationItemCallback<T> = (
   item: T,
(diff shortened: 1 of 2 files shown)


src/autoPagination.ts
```diff
@@ -362,7 +351,7 @@ function wrapAsyncIteratorWithCallback<T>(
         onItem(item, next);
       }).then((shouldContinue) => {
         if (shouldContinue === false) {
-          return handleIteration({done: true});
+          return handleIteration({done: true, value: undefined});
```
richardm-stripe, line 354: This is the only real code change. It's not breaking because in `handleIteration`, the argument is only used like
```
      const item = iterResult.value;
      return new Promise((next) => {
        onItem(item, next);
```
and iterResult.value will be 'undefined' regardless of whether .value is absent or .value is explicitly undefined.

pakrym-stripe (approved):

PR #1746, opened 2023-04-05
Remove deprecated types

anniel-stripe:
These types should have been removed in v11.0.0 when we removed these deprecated option names ([Changelog](https://github.com/stripe/stripe-node/blob/v11.0.0/CHANGELOG.md#:~:text=Remove%20deprecated%20option%20names)).

I set the base branch to be the next major branch, but let me know if you'd prefer to merge into master since these options no longer exist anyways so the types are wrong.


pakrym-stripe (approved):

richardm-stripe (approved):

richardm-stripe:
Let's do this into `master` instead.

PR #1752, opened 2023-04-07
Convert tests to typescript

pakrym-stripe:
Adds a new tsconfig that includes `test` directory in addition to src. Now our tests are actually typechecked.
Renames all tests to `.ts`.
Suppressed type checking in a lot of tests. We can unsupress them one by one.


package.json
```diff
@@ -62,7 +62,7 @@
   "scripts": {
     "build": "yarn build-esm && yarn build-cjs",
     "build-esm": "mkdir -p esm && tsc -p tsconfig.esm.json && echo '{\"type\":\"module\"}' > esm/package.json",
-    "build-cjs": "mkdir -p cjs && tsc -p tsconfig.json && echo '{\"type\":\"commonjs\"}' > cjs/package.json",
+    "build-cjs": "mkdir -p cjs && tsc -p tsconfig.cjs.json",
```
anniel-stripe, line 65: Is there a reason this needs to be deleted? If not, it seems like it's better to be explicit about the "type": (https://webpack.js.org/guides/package-exports/#providing-only-a-commonjs-version says it helps to statically detect CommonJS files)

anniel-stripe (commented):
Thanks for doing this! Just one comment about the change in `build-cjs`

package.json
pakrym-stripe: I'n not sure if anyone outside our tests consume that `package.json`. Isn't the article you linked talking about the main `package.json`?

anniel-stripe: Oh hmm, I did this because several sources I looked at recommended this convention for dual modules:

https://webreflection.medium.com/a-nodejs-dual-module-deep-dive-8f94ff56210e#:~:text=%7B%22type%22%3A%20%22commonjs%22%7D

https://evertpot.com/universal-commonjs-esm-typescript-packages/#:~:text=As%20you%20can%20also%20see%20that%20both%20our%20cjs%20and%20esm%20packages%20get%20a%201%2Dline%20package.json%20file%20with%20a%20type%20property.

The second source says this is because Node.js needs it to correctly interpret the .js files in the `cjs` folder.

pakrym-stripe, line 65:
```suggestion
    "build-cjs": "mkdir -p cjs && tsc -p tsconfig.cjs.json && echo '{\"type\":\"commonjs\"}' > cjs/package.json",
```

anniel-stripe (approved):

PR #1762, opened 2023-04-14
Add Deno webhook signing example

anniel-stripe:
## Summary
Add a Deno webhook signing example. Also updated our test script to actually exit with an error if the server never starts (the test didn't fail when Deno wasn't installed in CI).


examples/webhook-signing/deno/main.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import {serveListener} from 'https://deno.land/std@0.116.0/http/server.ts';
+import 'https://deno.land/x/dotenv/load.ts';
+
+import Stripe from 'npm:stripe@^11.16';
+
+const stripe = Stripe(Deno.env.get('STRIPE_API_KEY'));
+
+// Must specify hostname explicitly, see https://github.com/denoland/deno/issues/5144
+const server = Deno.listen({hostname: '127.0.0.1', port: 8080});
```
pakrym-stripe, line 11: can we bind to port 0 (dynamic) and retrieve the actual value from the server?

examples/webhook-signing/test/main.ts
```diff
@@ -112,7 +112,11 @@ const sendTestRequest = async (url: string): Promise<string> => {
     process.exit(0);
   } catch (e) {
     console.error(e);
+    throw e;
   } finally {
     server.kill();
   }
-})().catch(console.log);
+})().catch((e) => {
```
pakrym-stripe, line 119: Was a failing test not failing the CI?

pakrym-stripe (changes requested):
Can we get dynamic port binding working? It will make test less flaky and resistant to the situation where another process already uses the port.

examples/webhook-signing/test/main.ts
anniel-stripe: Before, if the server hadn't started (e.g., if main.ts wasn't executable or `deno` wasn't installed), then the test would still pass, for example in https://github.com/stripe/stripe-node/actions/runs/4696233056/jobs/8326148593 the test exits 1 with:
```
sh: 1: 
deno: not found
```
but CI does not fail.

examples/webhook-signing/deno/main.ts
anniel-stripe: Yes! Added in latest commit

pakrym-stripe (approved):
Nice work!

PR #1775, opened 2023-04-27
Introduce stripe.rawRequest as a canonical way to request APIs without definitions

anniel-stripe:


cjs/stripe.core.js
```diff
@@ -34,7 +34,7 @@ const ALLOWED_CONFIG_PROPERTIES = [
 ];
 const defaultRequestSenderFactory = (stripe) => new RequestSender_js_1.RequestSender(stripe, StripeResource_js_1.StripeResource.MAX_BUFFERED_REQUEST_METRICS);
 function createStripe(platformFunctions, requestSender = defaultRequestSenderFactory) {
-    Stripe.PACKAGE_VERSION = '12.2.0-beta.1';
+    Stripe.PACKAGE_VERSION = '12.3.0-beta.1';
```
anniel-stripe, line 37: Note to self: get rid of build output, no longer needed

src/RequestSender.ts
```diff
@@ -555,7 +567,12 @@ export class RequestSender {
         prepareAndMakeRequest
       );
     } else {
-      prepareAndMakeRequest(null, stringifyRequestData(data || {}));
+      prepareAndMakeRequest(
+        null,
+        options.apiMode === 'preview'
+          ? JSON.stringify(data || {}, dateTimeReplacer)
```
pakrym-stripe, line 573: Let's extract this + dateTimeReplacer into a separate function. Feels a bit untidy to have `dateTimeReplacer` floating around.

anniel-stripe: I extracted this to `jsonStringifyRequestData` defined in utils.ts, and I also renamed the existing `stringifyRequestData` -> `queryStringifyRequestData` to better differentiate between the two.

pakrym-stripe:
I think we also need `developer_message` handling here.

src/stripe.core.ts
```diff
@@ -202,6 +214,31 @@ export function createStripe(
...
     _platformFunctions: null!,
 
+    rawRequest(
+      method: string,
+      fullPath: string,
+      params?: RequestData,
+      options?: RequestOptions,
+      callback?: RequestCallback // for testing
+    ): Promise<any> {
+      const Resource = Stripe.StripeResource.extend({
```
pakrym-stripe, line 224: Do we need all the logic that happens in StripeResource/stripeMethod? 

Can we call into `_requestSender` directly?

src/RequestSender.ts
```diff
@@ -535,6 +537,9 @@ export class RequestSender {
         const apiVersion = this._stripe.getApiField('version');
         const headers = this._makeHeaders(
           auth,
+          options.apiMode === 'preview'
```
anniel-stripe, line 540: We should also generate the preview API version + set `Stripe-Version` header

src/stripe.core.ts
anniel-stripe: I think we do, there's a lot of functionality that lives in `StripeResource` and `stripeMethod` (especially the logic in [_makeRequest](https://github.com/stripe/stripe-node/blob/beta/src/StripeResource.ts#L196)) that's hard to extract out. The reason I'm using `StripeResource` is because the code example we used to offer for beta endpoints in stripe-node was like:
```typescript
stripe.setApiVersion('...');
const resource = Stripe.StripeResource.extend({
  request: Stripe.StripeResource.method({
    method: 'POST',
    path: '...',
  })
});

new resource(stripe).request({
  ...
},
  function(err, response) {
    // asynchronously called
  }
);
```

test/stripe.spec.ts
```diff
@@ -636,4 +636,186 @@ describe('Stripe Module', function() {
...
+      );
+    });
+
+    it('should make request successfully', async () => {
+      const response = await stripe.rawRequest('GET', '/v1/customers', {});
+
+      expect(response).to.have.property('object', 'list');
+    });
+
+    it('should allow passing in query params', (done) => {
```
pakrym-stripe, line 795: Do we have similar logic in other dynamic languages? Or sample generation definitely embeds query arguments into path.

src/stripe.core.ts
```diff
@@ -202,6 +209,22 @@ export function createStripe(
     _requestSender: null!,
     _platformFunctions: null!,
 
+    rawRequest(
+      method: string,
+      fullPath: string,
+      params?: RequestData,
+      options?: RequestOptions,
+      callback?: RequestCallback // for testing
```
pakrym-stripe, line 217: Is it only for testing?

src/StripeResource.ts
```diff
@@ -157,6 +157,8 @@ StripeResource.prototype = {
     const options = getOptionsFromArgs(args);
     const host = options.host || spec.host;
     const streaming = !!spec.streaming;
+    const apiMode = options.apiMode || 'standard';
```
pakrym-stripe, line 160: Should we support overriding apiMode for non raw_request calls?

src/RequestSender.ts
```diff
@@ -404,6 +412,88 @@ export class RequestSender {
...
+        const bodyData = dataInQuery ? {} : data;
+        const queryData = dataInQuery ? data : {};
+
+        opts = {
+          requestMethod,
+          requestPath: fullPath,
+          bodyData,
+          queryData,
+          auth: calculatedOptions.auth,
+          // @ts-ignore-next-line
```
pakrym-stripe, line 447: nit: can we fix this?

test/stripe.spec.ts
anniel-stripe: I can log this as a follow-up item to check that the other dynamic languages support this.

src/StripeResource.ts
anniel-stripe: Ah you're right, we shouldn't let this leak into non raw_request calls. Let me refactor this

src/Error.ts
```diff
@@ -50,7 +50,7 @@ export class StripeError extends Error {
   readonly source?: any;
 
   constructor(raw: StripeRawError = {}) {
-    super(raw.message);
+    super(raw.developer_message || raw.message);
```
pakrym-stripe, line 53: nit: iirc we preferred  `message` in other languages

src/Types.d.ts
```diff
@@ -66,6 +67,7 @@ export type RequestOpts = {
   host: string | null;
   streaming: boolean;
   settings: RequestSettings;
+  apiMode?: 'standard' | 'preview';
```
pakrym-stripe, line 70: nit: define a type?

src/utils.ts
```diff
@@ -43,11 +49,15 @@ export function isOptionsHash(o: unknown): boolean | unknown {
...
-export function stringifyRequestData(data: RequestData | string): string {
+export function queryStringifyRequestData(
+  data: RequestData | string,
+  apiMode?: APIMode
+): string {
   return (
     qs
       .stringify(data, {
         serializeDate: (d: Date) => Math.floor(d.getTime() / 1000).toString(),
+        arrayFormat: apiMode === 'preview' ? 'repeat' : 'indices',
```
pakrym-stripe, line 60: nit: is this used?

test/RequestSender.spec.ts
```diff
@@ -513,6 +517,26 @@ describe('RequestSender', () => {
...
+
+        nock(`https://${options.host}`)
+          .post('/v1/charges')
+          .reply(400, {
+            error,
+          });
+
+        try {
+          await realStripe.rawRequest('POST', '/v1/charges');
+          throw new Error('Should not reach here');
```
pakrym-stripe, line 533:
```
expect(() => {
        }).to.throw();
```
?

test/stripe.spec.ts
```diff
@@ -636,4 +636,167 @@ describe('Stripe Module', function() {
...
+  describe('rawRequest', () => {
+    const returnedCustomer = {
+      id: 'cus_123',
+    };
+
+    it('should make request with specified encoding FORM', (done) => {
+      return testUtils.getTestServerStripe(
+        {},
+        (req, res) => {
+          expect(req.headers['content-type']).to.equal(
```
pakrym-stripe, line 649: nit: assert API version here as well?

pakrym-stripe (approved):
nits

src/utils.ts
anniel-stripe: Nope, removed

PR #1782, opened 2023-05-01
Let user supply a timestamp when verifying webhooks

### LINKED ISSUE #1778: Allow specifying timestamp in webhook event signature verification

I am using the Stripe node API in a setting in which inbound events are placed in a queue before being processed by my code. As a result, there is a variable delay between when the webhook event is sent by Stripe and when it reaches my code. The only way to deal with this right now is to either increase tolerance in the call to `constructEvent` or to implement my own verification. 

I would prefer if `constructEvent` had an optional parameter that is the timestamp against which the webhook event timestamp will be compared. (Currently it always uses `now()` for comparison.) This would allow me to pass in the timestamp of when my queueing system received the message (rather than the time when I am processing the message). 

Writing my own signature verification

### Additional context

_No response_

richardm-stripe:
## Summary

Lets you supply an optional timestamp to Webhook.constructEvent (and all intermediate functions) to use instead of `Date.now()` when verifying webhook signatures.
Fixes #1778

### DIFF
--- a/types/Webhooks.d.ts
+++ b/types/Webhooks.d.ts
@@ -36,7 +36,12 @@ declare module 'stripe' {
         /**
          * Optional CryptoProvider to use for computing HMAC signatures.
          */
-        cryptoProvider?: CryptoProvider
+        cryptoProvider?: CryptoProvider,
+
+        /**
+         * Optional: timestamp to use when checking signature validity. Defaults to Date.now().
+         */
+        now?: number
       ): Stripe.Event;
 
       /**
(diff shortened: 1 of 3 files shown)


types/Webhooks.d.ts
```diff
@@ -36,7 +36,12 @@ declare module 'stripe' {
         /**
          * Optional CryptoProvider to use for computing HMAC signatures.
          */
-        cryptoProvider?: CryptoProvider
+        cryptoProvider?: CryptoProvider,
+
+        /**
+         * Optional: timestamp to use when checking signature validity. Defaults to Date.now().
+         */
+        now?: number
```
anniel-stripe, line 44: Should we name this something more descriptive than `now`? Maybe `currentTime` or or `comparisonTime`?

anniel-stripe (commented):
Thanks for fixing this so fast! Left a comment about the parameter name, but otherwise LGTM!

anniel-stripe (approved):

PR #1800, opened 2023-05-25
Update cloudflare-pages test project build

anniel-stripe:
The test has started leaving around a `_worker.bundle` file for me locally. Update the build script to remove all build artifacts after testing.

### DIFF
--- a/testProjects/cloudflare-pages/package.json
+++ b/testProjects/cloudflare-pages/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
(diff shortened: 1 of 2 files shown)


testProjects/cloudflare-pages/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "build": "wrangler pages functions build && rm _worker.js"
+    "build": "wrangler pages functions build && rm _worker.*"
```
pakrym-stripe, line 6: Should we add this file to gitignore instead?

pakrym-stripe (approved):

PR #1806, opened 2023-06-01
Allow request-level options to disable retries

richardm-stripe:
## Summary
Allows `maxNetworkRetries` on per-request options to *disable* retries. This seems to have been a bug: ever since per-request retries [were introduced](https://github.com/stripe/stripe-node/pull/703/files#diff-d5d3bf39549deedcf2b6f61e2f2f7f8398c71b21ba178612e0ff5c81a06d1ce5R288), code like
```javascript
stripe = new Stripe(..., {maxNetworkRetries: 1});
stripe.customers.create(..., {maxNetworkRetries: 0});
```
would still (incorrectly) perform a retry attempt for certain failures, although

```javascript
stripe = new Stripe(..., {maxNetworkRetries: 1});
stripe.customers.create(..., {maxNetworkRetries: 2});
```
would correctly perform up to 2 retry attempts.

## Release plan

I'm going to include this in the upcoming major release along with other retry-related changes, even though I think this is a non-disruptive bugfix that could go in a patch version.


src/RequestSender.ts
```diff
@@ -287,7 +287,7 @@ export class RequestSender {
 
   // Max retries can be set on a per request basis. Favor those over the global setting
   _getMaxNetworkRetries(settings: RequestSettings = {}): number {
-    return settings.maxNetworkRetries &&
+    return settings.maxNetworkRetries !== undefined &&
```
pakrym-stripe, line 290: Do we need this? I thought you said `Number.isInteger` is enough.

pakrym-stripe (approved):

src/RequestSender.ts
richardm-stripe: Typescript wants it.

Issue #1827, opened 2023-06-26
Clarify Intent of SubtleCryptoProvider Message


sundaycrafts:
### Is your feature request related to a problem? Please describe.

The current error message for the default SubtleCryptoProvider doesn't clearly indicate that a synchronous constructEvent cannot be used. The existing error message states, "SubtleCryptoProvider cannot be used in a synchronous context."

### Describe the solution you'd like

If there's only one context in which this error message appears, the message could be made more helpful. For example, it might suggest using 'constructEventAsync' as an alternative, or provide a link to the relevant section of the documentation.

### Describe alternatives you've considered

N/A

### Additional context

N/A

richardm-stripe:
Hello @sundaycrafts. Thank you for taking the time to report this and helping us improve our library.

I took a crack at this in https://github.com/stripe/stripe-node/pull/1828/files, let me know if you have any thought about it there.

richardm-stripe:
https://github.com/stripe/stripe-node/pull/1828 has been merged. It will be bundled with our release later this week.

sundaycrafts:
Thank you for your great work!

PR #1828, opened 2023-06-26
Better CryptoProvider error

richardm-stripe:
In response to #1827.

* I could have made a smaller change to just add the string to `SubtleCryptoProvider.computeHMACSignature` itself, but it felt wrong to me for `.computeHMACSignature` to make assumptions about how it was called, so instead I introduced a specialized error class for `constructEvent` can produce the message. Open to reversing this upon feedback.
* I didn't change the default implementation of `.computeHMACSignature` to throw this error. That means we're asking hypothetical implementors of additional async-only CryptoProviders will have to explicitly override `.computeHMACSignature` to throw this error instead of just only implementing `.computeHMACSignatureAsync`. This kind of is a bummer but I think it prevents misleading error messages in the hypothetical case where the implementer implements neither method. This almost certainly doesn't matter, but figured I'd mention.

### DIFF
--- a/src/Webhooks.ts
+++ b/src/Webhooks.ts
@@ -1,5 +1,8 @@
 import {StripeError, StripeSignatureVerificationError} from './Error.js';
-import {CryptoProvider} from './crypto/CryptoProvider.js';
+import {
+  CryptoProvider,
+  CryptoProviderOnlySupportsAsyncError,
+} from './crypto/CryptoProvider.js';
 import {PlatformFunctions} from './platform/PlatformFunctions.js';
 
 type WebhookHeader = string | Uint8Array;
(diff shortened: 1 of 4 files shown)


test/Webhook.spec.ts
```diff
@@ -152,13 +153,25 @@ describe('Webhooks', () => {
...
+        payload: 'payload',
+        secret: 'secret',
+      });
+      expect(() => {
+        stripe.webhooks.constructEvent('payload', header, 'secret', 0, {
+          computeHMACSignature() {
+            throw new CryptoProviderOnlySupportsAsyncError('foobar');
+          },
+        });
+      }).to.throw(/Use `await constructEventAsync/);
```
anniel-stripe, line 172: Does this actually check the message? Shouldn't the error be "/foobar\nUse `await constructEventAsync(...)` instead of `constructEvent(...)`/"?

src/Webhooks.ts
```diff
@@ -79,14 +82,22 @@ export function createWebhooks(
...
+          header,
+          secret,
+          tolerance || Webhook.DEFAULT_TOLERANCE,
+          cryptoProvider,
+          receivedAt
+        );
+      } catch (e) {
+        if (e instanceof CryptoProviderOnlySupportsAsyncError) {
+          e.message +=
+            '\nUse `await constructEventAsync(...)` instead of `constructEvent(...).`';
```
anniel-stripe, line 97:
```suggestion
            '\nUse `await constructEventAsync(...)` instead of `constructEvent(...)`';
```
nit: period inside of \`\`

anniel-stripe (approved):

test/Webhook.spec.ts
richardm-stripe: It runs the regex on the message. But I should include 'foobar' at least.

PR #1834, opened 2023-06-30
Allow for deleting customer's email address in type defintions

snoack:
You can delete a customer's email address via `stripe.customers.update(id, {email: null})`. However, this is currently causing a type error, since the type definitions don't allow the use of `null` here. But if you bypass the type check it works as expected. I wonder if other `CustomerUpdateParams` properties have the same issue?

### DIFF
--- a/types/CustomersResource.d.ts
+++ b/types/CustomersResource.d.ts
@@ -328,7 +328,7 @@ declare module 'stripe' {
       /**
        * Customer's email address. It's displayed alongside the customer in your dashboard and can be useful for searching and tracking. This may be up to *512 characters*.
        */
-      email?: string;
+      email?: string | null;
 
       /**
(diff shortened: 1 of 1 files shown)


richardm-stripe:
Hello @snoack, sorry for the delay on this.

You can pass the empty string `''` in order to trigger the emptying of the email string (stripe-node represents `null` values as empty strings when sending them over the wire anyway because the Stripe API is form-encoded).

I investigated and apparently there's an issue with several fields of type `string` that would be better represented as `Stripe.Emptyable<string>` to allow passing null. Many of them were fixed in [v12.17.0](https://github.com/stripe/stripe-node/releases/tag/v12.17.0) but the particular field in your report is waiting on a little extra work to fix the definitions upstream.

Thank you for reporting and thanks for your patience!

Issue #1794, opened 2023-05-18
TypeScript - Property 'plan' does not exist on type 'Subscription'


jonbarrow:
### Describe the bug

The property `plan` is not accessible on `Subscription` objects when used in TypeScript. The `plan` property is, however, present in the `Subscription` object

![Screenshot from 2023-05-18 16-16-57](https://github.com/stripe/stripe-node/assets/27011796/c02f7d65-d88e-4674-bfa4-302d679fb546)

![Screenshot from 2023-05-18 16-16-33](https://github.com/stripe/stripe-node/assets/27011796/a4dc5f6a-c034-471d-a921-f9523f43fca6)

### To Reproduce

1. Create an instance of `Stripe`
2. Request a `Subscription` object
3. Try to access the `plan` property

### Expected behavior

For the `plan` property to be accessible

### OS

Ubuntu 22.04

### Node version

Node v18.12.1

### Library version

stripe-node v12.5.0

### API version

2022-11-15

### Additional context

The documentation for the `Subscription` object also does not list the `plan` property, and several others <https://stripe.com/docs/api/subscriptions/object>

remi-stripe:
@jonbarrow The `plan` property on Subscription has been deprecated for many years now so it's expected that it's not in types and undocumented. See https://github.com/stripe/stripe-node/issues/1517#issuecomment-1221355535 for a longer explanation as to why.
We recommend using `items.data[xxx].price` instead!

jonbarrow:
> @jonbarrow The `plan` property on Subscription has been deprecated for many years now so it's expected that it's not in types and undocumented. See https://github.com/stripe/stripe-node/issues/1517#issuecomment-1221355535 for a longer explanation as to why.
> 
> We recommend using `items.data[xxx].price` instead!

Regardless on its deprecation, the data is still present and should honestly be documented as such. This property is still being included in data made even today, and it should still be accessible until it *actually* is removed. The documentation should also reflect that it exists but is deprecated. It's a bit nonsensical to have data which clearly still exists but is being arbitrarily withheld from developers. Improper documentation as to what the objects *actually* return will just lead to more issues like this

jonbarrow:
@remi-stripe I definitely understand deprecating parts of an API, but without proper documentation of deprecated features developers can run into footguns. The data is still available in the API, and so without proper documentation it's unclear as to what the data is or what the support plan is or anything like that. So long as the data is still returned by the current version of the API it should at the very least be documented as a deprecated feature

[...]

PR #1835, opened 2023-07-01
Add "plan" property of Subscription object to type definitions

snoack:
It looks like the `Subscription` object's `plan` property despite still being present in the latest API version (2022-11-15) has been left out of the type definitions.

### DIFF
--- a/types/Subscriptions.d.ts
+++ b/types/Subscriptions.d.ts
@@ -185,6 +185,11 @@ declare module 'stripe' {
       pending_update: Subscription.PendingUpdate | null;
 
       /**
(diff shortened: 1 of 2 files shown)


remi-stripe:
Hey @snoack ! Thanks a lot for the PR! Unfortunately, we won't merge this as the `plan` property has been deprecated for many years now and it's mostly here for old integrations. You can read more about this in https://github.com/stripe/stripe-node/issues/1794 where I went over why in details. As a developer, we recommend using `items.data[0].price` instead. I hope this helps!

Issue #1847, opened 2023-07-16
Changelog typo `CheckoutSessionCreateParams`


davidbielik:
### Describe the bug

I noticed that the changelog often mentions `CheckoutSessionCreateParams` yet I can't find this type in `stripe/types`.

Is this supposed to say `Checkout.SessionCreateParams`?

### To Reproduce

See: https://github.com/stripe/stripe-node/blob/master/CHANGELOG.md

### Expected behavior

`Checkout.SessionCreateParams` instead of `CheckoutSessionCreateParams`

### OS

macOS

### Node version

Node v16.14.2

### Library version

stripe-node v12.13.0

### API version

2022-08-01

PR #1848, opened 2023-07-16
Changelog: `CheckoutSessionCreateParams` -> `Checkout.SessionCreateParams`

davidbielik:
Fixes https://github.com/stripe/stripe-node/issues/1847


richardm-stripe:
Hello @davidbielik, thank you for the submission. This looks good to me. I suspect there are more changelog glitches after the same pattern and will attempt to follow up with a fix for these.

richardm-stripe (approved):

PR #1854, opened 2023-07-24
import * as http -> import http from 'http'

richardm-stripe:
Fixes #1844. TL;DR we will use the equivalent of `import {default as http} from 'http'` in an ES module context, to allow monkey patching, and in our tsconfig.json we have put "esModuleInterop" so in the CJS context we will still use `require('http')`

PR #1859, opened 2023-07-28
Revert "import * as http -> import http from 'http'"

richardm-stripe:
Reverts stripe/stripe-node#1854

We should not use `esModuleInterop` because this is a breaking change

> Accordingly, library authors should set both allowSyntheticDefaultImports and esModuleInterop to false. This allows consumers to opt into these semantics, but does not require them to do so. Consumers can always safely use alternative import syntaxes (including falling back to require() and import()), or can enable these flags and opt into this behavior themselves.

https://www.semver-ts.org/#module-interop

Thank you for catching this @anniel-stripe!


anniel-stripe (approved):
Thank you!

PR #1862, opened 2023-07-28
Remove wrong signature overload for stripe.invoice.retrieveUpcoming

snoack:
When calling `stripe.invoice.retrieveUpcoming` with a single argument, it is expected to be of the type `InvoiceRetrieveUpcomingParams`. This is already supported by the first signature, as the second argument is optional. However, the second overlaoded signature is picked up which incorrectly expects a `RequestOptions` object as first/only argument.

### DIFF
--- a/types/InvoicesResource.d.ts
+++ b/types/InvoicesResource.d.ts
@@ -2566,9 +2566,6 @@ declare module 'stripe' {
         params?: InvoiceRetrieveUpcomingParams,
         options?: RequestOptions
       ): Promise<Stripe.Response<Stripe.UpcomingInvoice>>;
-      retrieveUpcoming(
-        options?: RequestOptions
-      ): Promise<Stripe.Response<Stripe.UpcomingInvoice>>;
 
       /**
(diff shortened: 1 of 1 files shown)


snoack:
I'm not sure if passing an `options` object as first argument will work as expected. You can pass a `params` object but again this will cause an error in TypeScript if no second argument is specified. For example, try:

```js
stripe.invoice.retrieveUpcoming({customer: 'some_customer_id'})
```

It works fine from JavaScript but causes an error in TypeScript, unless you explicitly pass `undefined` (or an empty object) as second argument:

```js
stripe.invoice.retrieveUpcoming({customer: 'some_customer_id'}, undefined)
```

anniel-stripe:
Hmm that's strange, I'm still not able to repro the TypeScript error. Would you mind sharing the error that you're seeing, as well as your stripe-node + TypeScript version?

snoack:
Sorry, I can no longer reproduce it which is odd.

anniel-stripe:
OK, feel free to open a bug report if you run into this issue again! I'll close out this PR.

Issue #1844, opened 2023-07-14
Importing Stripe via ES Modules does not respect monkey-patching 'http'


janhesters:
### Describe the bug

When trying to mock responses made with stripe-node in a Node.js enviroment, the requests hit the real server instead of getting caught by [MSW](https://mswjs.io/).

### To Reproduce

You can copy and paste the code below. (Either write actual handlers for your `stripeHandlers`, or you can leave them out, as we did below. If you want to try it out with handlers, simply spread them into `setupServer`.)

Sine we configured `onUnhandledRequest: 'warn'` MSW should log out to the console when requests are uncaught. The manual method with `axios` works correctly and logs out this warning / reacts when you configure handlers.

However, the second (in the code blocked `skip`ped) test does NOT log out this warning and simply hits the servers.

```ts
// @vitest-environment node
import axios from 'axios';
import { rest } from 'msw';
import { setupServer } from 'msw/node';
import Stripe from 'stripe';
import invariant from 'tiny-invariant';

import { stripeHandlers } from '~/test/mocks/handlers/stripe';

invariant(
  process.env.STRIPE_SECRET_KEY,
  'STRIPE_SECRET_KEY environment variable is required',
);

const stripeAdmin = new Stripe(process.env.STRIPE_SECRET_KEY, {
  apiVersion: '2022-11-15',
});

const getProducts = async () => {
  try {
    const response = await axios.get('https://api.stripe.com/v1/products', {
      headers: {
        Authorization: `Bearer ${process.env.STRIPE_SECRET_KEY}`,
      },
      params: { active: true },
    });

    console.log(response.data);
  } catch (error) {
    console.error(error);
  }
};

const server = setupServer();

beforeAll(() => server.listen({ onUnhandledRequest: 'warn' }));
afterEach(() => server.resetHandlers());
afterAll(() => server.close());

describe('getProducts()', () => {
  test('it works', async () => {
    const actual = await getProducts();
    const expected: any = [];

    expect(actual).toEqual(expected);
  });
});

describe.skip('stripeAdmin', () => {
  test('getting products works', async () => {
    const actual = await stripeAdmin.products.list();
    const expected: any = [];

    expect(actual).toEqual(expected);
  });
});
```

```ts
import { rest } from 'msw';

import { mockStripePrices, mockStripeProducts } from '../fixtures/stripe';

export const stripeHandlers = [
  rest.get('https://api.stripe.com/v1/products', (request, response, context) =>
[...]

PR #1866, opened 2023-08-02
Allow monkey patching http / https

anniel-stripe:
Achieves a similar affect as https://github.com/stripe/stripe-node/pull/1854 (but without turning on the `esModuleInterop` option) to enable monkey patching of the 'http' / 'https' modules when importing stripe as an ES module.

Also adds a test to prevent against future regressions, and makes `esModuleInterop: false` explicit.

Motivation: https://github.com/stripe/stripe-node/issues/1844

### DIFF
--- a/src/net/NodeHttpClient.ts
+++ b/src/net/NodeHttpClient.ts
@@ -1,5 +1,15 @@
-import * as http from 'http';
-import * as https from 'https';
+// It's important to do a default import here, rather thane
+// `import * as http from 'http'` because the latter will
+// use a "Module Namespace Exotic Object" which is immune to
+// monkey-patching, whereas http.default (in an ES Module context)
+// will resolve to the same thing as require('http'), which is
+// monkey-patchable. We care about this because users in their test
+// suites might be using a library like "nock" which relies on the ability
+// to monkey-patch and intercept calls to http.request.
+
+// Get rid of this type error Module '"http"' has no default export.ts(1192)
+import * as http_ from 'http';
+import * as https_ from 'https';
 import {RequestHeaders, RequestData} from '../Types.js';
 import {
   HttpClient,
@@ -7,6 +17,9 @@ import {
   HttpClientResponseInterface,
 } from './HttpClient.js';
 
+const http = ((http_ as unknown) as {default: typeof http_}).default || http_;
+const https =
+  ((https_ as unknown) as {default: typeof https_}).default || https_;
 const defaultHttpAgent = new http.Agent({keepAlive: true});
 const defaultHttpsAgent = new https.Agent({keepAlive: true});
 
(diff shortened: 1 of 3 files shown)


src/net/NodeHttpClient.ts
```diff
@@ -1,12 +1,25 @@
...
+// It's important to do a default import here, rather thane
+// `import * as http from 'http'` because the latter will
+// use a "Module Namespace Exotic Object" which is immune to
+// monkey-patching, whereas http.default (in an ES Module context)
+// will resolve to the same thing as require('http'), which is
+// monkey-patchable. We care about this because users in their test
+// suites might be using a library like "nock" which relies on the ability
+// to monkey-patch and intercept calls to http.request.
+
+// Get rid of this type error Module '"http"' has no default export.ts(1192)
```
richardm-stripe: This comment is defunct. I think a comment lower down will be more appropriate.
```suggestion
// It's important to do a default import here, rather thane
// `import * as http from 'http'` because the latter will
// use a "Module Namespace Exotic Object" which is immune to
// monkey-patching, whereas http.default (in an ES Module context)
// will resolve to the same thing as require('http'), which is
// monkey-patchable. We care about this because users in their test
// suites might be using a library like "nock" which relies on the ability
// to monkey-patch and intercept calls to http.request.

// Get rid of this type error Module '"http"' has no default export.ts(1192)
```

src/net/NodeHttpClient.ts
```diff
@@ -1,12 +1,25 @@
...
+import * as http_ from 'http';
+import * as https_ from 'https';
 import {RequestHeaders, RequestData} from '../Types.js';
 import {
   HttpClient,
   HttpClientResponse,
   HttpClientResponseInterface,
 } from './HttpClient.js';
 
+const http = ((http_ as unknown) as {default: typeof http_}).default || http_;
```
richardm-stripe, line 16:
```suggestion
// In the CommonJS context, `http = require('http')` which is monkey-patchable
// and `http.default` will be undefined.
//
// In the ES Module context, `http` will be a "Module Namespace Exotic Object" which
// is not monkey-patchable. However, `http.default` will also be set and be equivalent to
// the CommonJS `module.exports` and can be monkey-patched.
const http = ((http_ as unknown) as {default: typeof http_}).default || http_;
```

tsconfig.cjs.json
```diff
@@ -9,7 +9,8 @@
     "noImplicitThis": true,
     "strict": true,
     "strictFunctionTypes": true,
-    "types": [ "node" ]
+    "types": [ "node" ],
+    "esModuleInterop": false // This is a viral option, do not enable https://www.semver-ts.org/#module-interop
```
richardm-stripe, line 13: 💯

richardm-stripe (approved):
LGTM, some suggestions about comments.

src/net/NodeHttpClient.ts
richardm-stripe: No, your comment is better.

PR #1877, opened 2023-08-17
Change API_VERSION

richardm-stripe:
Looks like `make codegen-format` reads from API_VERSION and does some regex substitutions.

This will prevent the new API version from being reverted.


README.md
```diff
@@ -71,7 +71,7 @@ and instantiate it as `new Stripe()` with the latest API version.
 ```ts
 import Stripe from 'stripe';
 const stripe = new Stripe('sk_test_...', {
```
pakrym-stripe, line 73: unrelated: do we still care about making customers do this?

pakrym-stripe (approved):

pakrym-stripe (changes requested):
Let's change the script to grep API version out of the generated file and fix it once and for all.

pakrym-stripe (approved):
Approving to unblock releases.

PR #1878, opened 2023-08-17
Use src/apiVersion.ts, not API_VERSION as source of truth

richardm-stripe:


scripts/updateAPIVersion.js
```diff
@@ -11,7 +11,13 @@ const edit = (file, cb) => write(file, cb(read(file)));
 const API_VERSION = '2[0-9][2-9][0-9]-[0-9]{2}-[0-9]{2}';
 
 const main = () => {
-  const apiVersion = read('API_VERSION').trim();
+  const matches = [...read('src/apiVersion.ts').matchAll(/'([^']*)'/g)];
+  if (matches.length !== 1) {
+    throw new Error(
+      `Expected src/apiVersion.ts to include 1 single-quote-delimited string, but found ${matches.length}`
```
pakrym-stripe, line 17: Beta includes two. https://github.com/stripe/stripe-node/blob/beta/src/apiVersion.ts

pakrym-stripe (changes requested):
Needs to be able to handle beta branch

scripts/updateAPIVersion.js
richardm-stripe: All fixed and ran the script manually: [Fix](https://github.com/stripe/stripe-node/pull/1878/commits/3d809e857e7c69222853302aba78d80e78f675db)

pakrym-stripe (approved):

PR #1889, opened 2023-08-30
Fixes Typescript example in readme by including missing property 'typescript'

PGuimarais:


anniel-stripe:
Hi @PGuimarais - apologies for the delay in response here. The `typescript` option has no effect except to add "TypeScript" to your user-agent, and we are considering removing this config option as it is not really used. I'll be closing out this PR because this option has little use, but thank you for bringing this to our attention!

PR #1921, opened 2023-10-10
Add event types

pakrym-stripe:
Renames the current `Event` type to `EventBase`. Adds generated event definitions for known events with correct `data.object` reference type. 

Defines a new `Event` type that's a union of all strongly-typed events. The `Event` union is equivalent in structure to the old `Event` type except the `type` property that becomes an enum of all known even types and not just a string.

```ts
const event = await stripe.webhooks.constructEventAsync(body, signature, secret);

// Allows narrowing of event type
if (event.type == 'customer.created') {
  const customer = event.data.object; // customer is typed as Stripe.Customer
  console.log(customer.email); // No error, typed property access
}

// TypeScript error
// This comparison appears to be unintentional because the types "account.application.authorized" | ... and  "customer.crated"  have no overlap.
if (event.type == 'customer.crated') {}

```


anniel-stripe (approved):
🔥🔥🔥

PR #1952, opened 2023-11-08
Use AbortController for native fetch cancellation when available

### LINKED ISSUE #1950: Use AbortSignal for timeout in FetchHttpClient implementation

The current implementation of `FetchHttpClient` uses `Promise.race()` to reject the promise while not actually cancelling the fetch operation or releasing the underlying resources.

https://github.com/stripe/stripe-node/blob/4dd332f53652b890893e156d24e0410b670d1636/src/net/FetchHttpClient.ts#L65-L94

`Request.signal` is the standard API for cancelling an in-flight fetch request. It is supported by browser implementations, `node-fetch` and Node.js built-in fetch.

_No response_

### Additional context

I suppose you'd have to rely on an `AbortController` global variable, which is available since Node.js `v14.17.0`, so you'd have to either increase the minimum supported version (currently 12.x) or more likely, fall back to `Promise.race()` if `AbortController` is not available in the global scope.

Happy to help with a PR if you wish.

brettwillis:
Closes #1950


src/net/FetchHttpClient.ts
```diff
@@ -15,19 +21,72 @@ import {
...
+    return (url, init, timeout: number): Promise<Response> => {
+      // Use AbortController because AbortSignal.timeout() was added later in Node v17.3.0, v16.14.0
+      const abort = new AbortController();
+      let timeoutId: NodeJS.Timeout | null = setTimeout(() => {
+        timeoutId = null;
+        abort.abort(HttpClient.makeTimeoutError());
+      }, timeout);
+      return fetchFn(url, {
+        ...init,
+        signal: AbortSignal.timeout(timeout),
```
pakrym-stripe, line 75: Considering your comment above, should we guard against AbortSignal absence?

pakrym-stripe:
There is also a test failure that seems related.

pakrym-stripe (changes requested):
Question and failing test.

src/net/FetchHttpClient.ts
brettwillis: Sorry mate that one was a bit embarassing! It's supposed to be using the `AbortController.signal` I set up, not `AbortSignal.timeout()`. Fixed that.

brettwillis:
Ok, so this one test is failing on Node.js 12 because fetch does not exist and no implementation is provided. I have made it throw an error with a helpful message in this case.

Previously I think the test passed on Node.js 12 because the fetch function was not retrieved until a request is actually executed, at which point it would with an obscure error.

I think the new behaviour (clear message, and failing early) is better, but it means we need to modify the bahaviour of that test for Node.js 12. Do you have any preference? Perhaps provide a stub function for that test?

pakrym-stripe:
I like your explicit error approach and I think it makes sense to reflect it in the test. Depending on the `process.versions.node` we should expect either a value or an exception.

There is another test failure on 19 in the `rejects with a timeout error` test:

```
       HttpClientTestSuite
         makeRequest
           rejects with a timeout error:
     AssertionError: expected undefined to equal 'ETIMEDOUT'
```

brettwillis:
Ok yeah sorry I'm having trouble running the tests on my machine so I'm not picking up these failures.

This one was interesting... `node-fetch` does not respect the "reason" passed to `AbortController.abort(reason)`, instead it always throws it's own `AbortError`. Whereas the built-in fetch does throw the provided error. So I added a clause to normalise `AbortError`s to our own `TimeoutError`. Let me know your thoughts.

src/net/FetchHttpClient.ts
```diff
@@ -15,19 +21,95 @@ import {
...
+        // Some implementations, like node-fetch, do not respect the reason passed to AbortController.abort()
+        // and instead it always throws an AbortError
+        // We catch this case to normalise all timeout errors
+        if ((err as any).name === 'AbortError') {
+          throw HttpClient.makeTimeoutError();
+        } else {
+          throw err;
+        }
+      } finally {
+        if (timeoutId) {
```
pakrym-stripe, line 100: nit: is this ever not true?

pakrym-stripe:
It looks like fetch was added in node 17 (https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch) so we need to adjust the test.

src/net/FetchHttpClient.ts
brettwillis: Yeah if the timeout occurs before the fetch resolves then it is unnecessary to clear the timeout, and `timeoutId` will have been set to `null` in the timout callback.

Normally I wouldn't bother with setting `timeoutId` to `null` and doing this check, as I'm pretty sure calling `clearTimeout` after it has already fired is just a no-op [^1]. But I was just following suit with how the `Promise.race()` timeout was originally implemented here.

[^1]: The web spec states that [it is guaranteed that a timeoutID value will never be reused](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value), but I can't find a similar assertion in the Node.js docs, so perhaps it is best to stick with this method.

brettwillis:
Ah yeah I see even up until Node.js 18.0.0 it was behind the `--experimental-global-fetch` CLI flag. I've added a `< 18.0.0` check.

pakrym-stripe (approved):
Awesome work! Thank you for all the iterations and writing such a clean and well-commented code.

chris-erickson:
> Ok, so this one test is failing on Node.js 12 because fetch does not exist and no implementation is provided. I have made it throw an error with a helpful message in this case.
> 
> Previously I think the test passed on Node.js 12 because the fetch function was not retrieved until a request is actually executed, at which point it would with an obscure error.
> 
> I think the new behaviour (clear message, and failing early) is better, but it means we need to modify the bahaviour of that test for Node.js 12. Do you have any preference? Perhaps provide a stub function for that test?

I suspect a lot of folks will be arriving soon to learn why their Jest tests are failing now..

```
    fetch() function not provided and is not defined in the global scope. You must provide a fetch implementation.
     
25 | const stripe = new Stripe(STRIPE_SECRET_TOKEN, {
```

Is there any early guidance on how we should actually fix this? Is it a test problem, or an implementation problem?

brettwillis:
TL;DR you need to update your test to provide a fetch implementation, for example `Stripe.createFetchHttpClient(require('node-fetch'))`.

`Stripe.createFetchHttpClient()` now throws when using `Stripe.createFetchHttpClient()` without providing a fetch implementation on a runtime where there is no global fetch to use as a default (i.e. pre-Node.js 18). This highlights an invalid state wherein the HTTP client would previously have failed anyway (probably with a TypeError of some sort) upon trying to perform an actual request.

When using the fetch HTTP client on a runtime with no default fetch implementation, you must provide your own. For example, you could use `undici`, `node-fetch`, or a mock for the purpose of testing.

```js
const { fetch } = require(`undici`);
const stripe = new Stripe(STRIPE_SECRET_TOKEN, {
  httpClient: Stripe.createFetchHttpClient(fetch), // provide your own fetch implementation when runtime has no default
});
```

I suspect your test was not exercising any actual HTTP requests otherwise it would have been failing anyway? So you could provide a mock function of any kind if you don't want to use a fetch library.

<details><summary>Example</summary>

```js
// Restore previous behaviour of delaying failure until a request is performed
function fetch() {
  throw new Error('fetch() not implemented');
}

const stripe = new Stripe(STRIPE_SECRET_TOKEN, {
  httpClient: Stripe.createFetchHttpClient(fetch),
});
```
</details>

pakrym-stripe:
>I suspect a lot of folks will be arriving soon to learn why their Jest tests are failing now..

How are you referencing stripe in tests? Can you give an example of a test that started failing?

Issue #1956, opened 2023-11-14
Make webhooks utilities available as a static property


brettwillis:
### Is your feature request related to a problem? Please describe.

In some scenarios for me (normally when using Stripe Connect), we don't have enough information to create a Stripe instance until we have parsed the webhook event.

Specifically, we may need the `Event.account` property to determine what credentials/API key to use to create a Stripe instance.

However, the `Stripe.Webhooks` utility is only available *after* creating a Stripe instance.

### Describe the solution you'd like

After inspecting the code, it seems the webhook utilities instance does not depend at all on a specific Stripe instance.

https://github.com/stripe/stripe-node/blob/12fe437bb94bb088a3725d79292979cbeacb3302/src/Webhooks.ts#L70-L72

This webhook utilites instance could be made available as a static property on the `Stripe` constructor, and that instance also shared on the individual instances.

Like so:

```js
const Stripe = require('stripe');

// Available as a static property
Stripe.webhooks.constructEventAsync(...);

// Remains available on the instance also
const client = Stripe('...');
client.webhooks === Stripe.webhooks // true
```

Happy to help with a PR if you agree with the logic.

### Describe alternatives you've considered

A workaround is using `new Stripe('').webhooks` to get an instance of the webhook utilities, but it feels like a hack, because it relies on the internal workings of the `Stripe()` constructor not attempting to validate the empty API key.

brettwillis:
Ah I just noticed that `Stripe.webhooks` is already assigned as the factory function.
https://github.com/stripe/stripe-node/blob/12fe437bb94bb088a3725d79292979cbeacb3302/src/stripe.core.ts#L158

I don't know if this would make it a breaking change or not. It doesn't seem to actually document that property in the typings?

The problem with this factory function is that it requires an instance of `platformFunctions` which is also not easily available.

### Alternative 1

Maybe an alterative is to make the `Stripe.webooks` factory function have the `platformFunctions` parameter default to the `platformFunctions` used by the current Stripe object. Like

```js
// Stripe.webhooks = createWebhooks --- change this
Stripe.webhooks = (fns) => createWebhooks(fns ?? platformFunctions);
``` 

While this is not as elegant as single instance that can be shared around Stripe instances, it's a step in the right direction.

### Alternative 2

Otherwise we change the name `Stripe.webhooksInstance` or something, I don't know if I'd call that elegant either.

### Alternative 3 ✅ 

Maintain compatibillity and incorporate both proposed improvements like this

```js
[...]

PR #1958, opened 2023-11-14
Move Webhooks instance to static field

brettwillis:
See #1956 

`Stripe.webooks` has the following changes
- The `platformFunctions` parameter is now optional, defaulting to the `platformFunctions` instance used by the Stripe object
- It is now also confirms to the `Stripe.Webhooks` signature so that `Stripe.webhooks` can be used as a static reference to the webhook utilities

Furthermore, all `Stripe` instances share the same reference to the static `Stripe.webhooks` object/function.

The function signature will be removed as a breaking change in the next major release.

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -156,7 +156,10 @@ export function createStripe(
     this._setApiKey(key);
 
     this.errors = _Error;
-    this.webhooks = Stripe.webhooks;
+
+    // Once Stripe.webhooks looses the factory function signature in a future release
+    // then this should become this.webhooks = Stripe.webhooks
+    this.webhooks = createWebhooksDefault();
 
     this._prevRequestMetrics = [];
     this._enableTelemetry = props.telemetry !== false;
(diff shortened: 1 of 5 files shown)


src/stripe.core.ts
```diff
@@ -142,7 +156,7 @@ export function createStripe(
     this._setApiKey(key);
 
     this.errors = _Error;
-    this.webhooks = createWebhooks(platformFunctions);
+    this.webhooks = Stripe.webhooks;
```
anniel-stripe, line 159: Any reason why we're changing `webhooks` on the `Stripe` instance as well? Feels a little wrong to let this one start also act as a factory function when we're removing that behavior later - we'd prefer to keep the breaking change surface minimal when the time comes.

brettwillis: Ah yes you're right I was getting ahead of myself 😉

anniel-stripe (approved):
Thanks so much for another great issue and high quality PR!

PR #2010, opened 2024-01-31
Include features attribute in Product types

huynhtk80:
Product Types is missing features attribute that can be included in the stripe product object.

### DIFF
--- a/types/Products.d.ts
+++ b/types/Products.d.ts
@@ -124,6 +124,11 @@ declare module 'stripe' {
       unit_label?: string | null;
 
       /**
+       * Features of a specific product to help users tell the differences between similar products.
+       */
+      features?: Array<string> | null;
+
+      /**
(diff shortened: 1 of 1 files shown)


remi-stripe:
@huynhtk80 We added support for `features` on Product back in September in version [13.5.0](https://github.com/stripe/stripe-node/blob/master/CHANGELOG.md#1350---2023-09-07). If you look at the types for the Product class, the `features` option is already [defined](https://github.com/stripe/stripe-node/blob/master/types/Products.d.ts#L71-L74). Same in your PR, you added `features` at the bottom instead of in the alphabetical order and if you look earlier in the file it's right there.
What's possible is that you are on an older version of stripe-node, before the feature shipped, and didn't realize it was already supported.

I'm going to close this out but let me know if you have any other question!

PR #2025, opened 2024-02-22
Standardize parameter interface names

pakrym-stripe:
Standardizes param class names to `<ResourceName><MethodName>Params`.

Adds type-forwards to preserve existing types.


anniel-stripe (approved):
Love this!

Issue #2042, opened 2024-03-19
error.type does not survive minification


sdsanders:
### Describe the bug

When bundling and minifying using a tool like esbuild, the `type` included in errors thrown by Stripe is also minified into a shorter string. Strings are usually safe from minification so this behavior is unexpected.

### To Reproduce

Minimal reproduction: https://github.com/sdsanders/stripe-minify-error
Run `npm run build && npm start` - incorrect `error.type` will be logged
Run `npm run build:unminified && npm start` - correct `error.type` will be logged

### Expected behavior

Error type string should remain intact after minification.

### OS

macOs

### Node version

Node v18.19.0

### Library version

stripe-node 

### API version

 2023-10-16

### Additional context

Minifying might not be a super common practice server-side but it can have some advantages. We run on AWS Lambda which is usually resource constrained and minification [can improve cold start times](https://aws.amazon.com/blogs/compute/optimizing-node-js-dependencies-in-aws-lambda/).

anniel-stripe:
Hi there @sdsanders ! Thanks for the bug report and for the minimal repro. This is happening because in the error object constructor, `type` is set to the constructor name: https://github.com/stripe/stripe-node/blob/master/src/Error.ts#L54

A fix on our side would be to explicitly specify the type as a string in the `StripeError` constructor. I'll have a PR up for this soon, as it seems like it would be relatively non-intrusive.

anniel-stripe:
The fix has been merged! The changes will be included in today's release.

sdsanders:
> The fix has been merged! The changes will be included in today's release.
> 
> Update: Included in v14.22.0

Thanks so much! Can confirm that this fixes the issue in our setup.

PR #2043, opened 2024-03-20
Don't mutate error.type during minification

anniel-stripe:
The `type` property on error objects was set to `this.constructor.name`, which gets mutated during minification. To fix this, explicitly set the `type` to a hardcoded string for known subclasses of `StripeError`. This also keeps the default behavior the same in the `StripeError` parent class for backwards compatibility.

I've added a test project to verify `error.type` does not get minified when bundling / minifying with esbuild.

Fixes https://github.com/stripe/stripe-node/issues/2042


test/Integration.spec.ts
```diff
@@ -55,6 +55,22 @@ describe('Integration test', function() {
     await runTestProject('mjs-ts');
   });
 
+  describe('esbuild', () => {
```
pakrym-stripe, line 58: It might be better to find a way of running all tests on a minified version of the library long-term.

pakrym-stripe (approved):

test/Integration.spec.ts
anniel-stripe: Agreed! Filed a ticket for this.

PR #2061, opened 2024-04-07
Make cloudflare package export

bitofbreeze:
Cloudflare's runtime uses the "workerd" conditional export key as described [here](https://developers.cloudflare.com/workers/wrangler/bundling/#conditional-exports) and [here](https://runtime-keys.proposal.wintercg.org/#workerd).


anniel-stripe (approved):
Thank you for the contribution @bitofbreeze!

PR #2074, opened 2024-04-29
Note on `limit` param in `autoPagingToArray`

tvanantwerp:
This change to the `README.md` file makes it more explicit that the `limit` value passed to `autoPagingToArray` is the limit for the total number of items that will be fetched from Stripe, and *not* a limit for the number of items retrieved per fetch during multiple rounds of fetching. The note about preventing runaway memory growth was suggestive of this behavior, but not entirely explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,7 +467,8 @@ stripe.customers
 
 This is a convenience for cases where you expect the number of items
 to be relatively small; accordingly, you must pass a `limit` option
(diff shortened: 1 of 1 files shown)


richardm-stripe (approved):
Looks good to me! Sorry for the confusion here.

Also added some comments to the code snippet below to hopefully clarify even further.

PR #2122, opened 2024-06-26
Remove TS test that cares about a specific API version

xavdid-stripe:
When releasing the most recent major, the types required a specific API version: 

https://github.com/stripe/stripe-node/blob/e219b05f02f09efb57b13e50977012d001a49223/types/lib.d.ts#L30

So this test was failing until we adjusted the string. A test that fails on a schedule isn't great and the test isn't verifying anything especially useful. I think it's more trouble than it's worth, so I pulled it.

### DIFF
--- a/testProjects/mjs-ts/index.ts
+++ b/testProjects/mjs-ts/index.ts
@@ -1,13 +1,15 @@
 import DefaultStripe, {Stripe} from 'stripe';
 
 const stripe = new Stripe(process.argv[2], {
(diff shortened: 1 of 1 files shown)


anniel-stripe (approved):
nice!

ramya-stripe:
Haha - I was just going to work on this - saw the merge conflicts when I pulled in master :)

Thank you for picking it up!

PR #2116, opened 2024-06-23
Removed qs

gkiely:
Based off this comment
https://github.com/stripe/stripe-node/issues/2038#issuecomment-2004751719

Includes a lightweight alternative to qs.stringify.
I have added 80% of the tests from the qs library, if you think this is a good approach I can add the remainder. I wanted to confirm it is viable before I put any more time into it :)

PR #2146, opened 2024-07-27
feat: migrate to picoquery

43081j:
Migrates from qs to picoquery for query strings, a faster and lighter alternative.

This is an alternative to #2116 - using pq to provide the logic rather than rolling our own

if you'd prefer to roll your own, feel free to close this

### DIFF
--- a/tsconfig.esm.json
+++ b/tsconfig.esm.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "outDir": "./esm",
     "module": "es2022",
+    "moduleResolution": "node",
   },
   "exclude": ["./src/stripe.cjs.node.ts", "./src/stripe.cjs.worker.ts"],
 }
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "@types/chai": "^4.3.4",
     "@types/chai-as-promised": "^7.1.5",
     "@types/mocha": "^10.0.1",
-    "@types/qs": "^6.9.7",
     "@typescript-eslint/eslint-plugin": "^4.33.0",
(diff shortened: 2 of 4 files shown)


tsconfig.esm.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "outDir": "./esm",
     "module": "es2022",
+    "moduleResolution": "node",
```
43081j, line 6: locally the esm tsc build failed without this. i'm not sure how it ever worked without it 🤔

helenye-stripe: We have been using the default and have not had issues, could you pull this out of this PR?

helenye-stripe:
Hi @43081j, thanks so much for this submission! Sorry for the delay. I think after reviewing the approaches for removing qs, we would like to investigate this further. Would you mind investigating the [test failure here](https://github.com/stripe/stripe-node/actions/runs/10122545091/job/28666083478?pr=2146#step:10:42)?

43081j:
unfortunately this is because node 12 doesn't support nullish coalescing

bumping the engine constraint here (i.e. dropping node <14 support) seems the sensible thing to do eventually but that seems like a bigger decision for you to make

until then i guess this PR is blocked

let me know what you think. if we have to hold off until node <14 support is dropped one day, im happy to come back at that point and revisit this

xavdid-stripe:
Yes, our version support policy is something we're working on. We'll put a pin in this and merge it when we can drop node 12 support.

Thank you for your contribution!

prathmesh-stripe:
Thanks for the PR. Closing this in favor of https://github.com/stripe/stripe-node/pull/2168

PR #2149, opened 2024-08-02
Metadata may be undefined

JamesJansson:
Metadata fields may be undefined, as such accessing a field should have a type of `string | undefined`

### DIFF
--- a/types/shared.d.ts
+++ b/types/shared.d.ts
@@ -4,7 +4,7 @@ declare module 'stripe' {
      * Set of key-value pairs that you can attach to an object. This can be useful for storing additional information about the object in a structured format.
      */
     interface Metadata {
-      [name: string]: string;
+      [name: string]: string | undefined;
     }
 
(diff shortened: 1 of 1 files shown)


xavdid-stripe:
@JamesJansson thank you for the contribution!

That said, the types are correct as written. The `Metadata` interface describes the shape of the present data, which is a mapping of `string -> string`. So while every key won't be present in the object, keys that _are_ present are `string`, not `string | undefined`. 

I realize that's a little counterintuitive, but it's just how objects in typescript work. The best approach is probably to check `if key in metadata` before doing a `metadata[key]`. Alternatively, checking that the value is truthy before use.

JamesJansson:
I have updated the Metadata type so it behaves as described:
- an object with arbitrary strings as keys (`Record`)
- does not have a value for every possible key (`Partial`)
- cannot have fields be set to `undefined` (per your comment above)

PR #2150, opened 2024-08-02
Remove a ts-ignore by using Object.entries

QuentinLemCode:
Hello

Very little PR : we can avoid a @ts-ignore by using Object.entries instead of Object.keys


xavdid-stripe (approved):
Awesome, thank you!

PR #2160, opened 2024-08-19
Add codeowners

xavdid-stripe:
It's nice when PRs will auto-request the reviewers instead of having to do it ourselves. We'll also get notified about external contributions. On the other hand, this might lead to a lot of noise from bot PRs (so we'll start with just this repo and see how it goes).

This PR is blocked while the team is private; going to see about making it public.


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# All files should be reviewed by a member of the SDKs team
+* @stripe/api-library-reviewers
```
prathmesh-stripe, line 2: There seems to be an error here. Do we have this group publicly accessible?

xavdid-stripe: yep 😁 in the PR description:

> This PR is blocked while the team is private

I'm figuring out who has the permissions

xavdid-stripe: ok, so it turns out teams are _never_ publicly accessible if you're not part of the team. But, the issue was the team didn't have "write" access (through the reviewers group directly; we got it through other teams we're all on). I've updated those permissions and it looks like we're good now!

prathmesh-stripe (approved):

PR #2161, opened 2024-08-22
Metadata may be undefined

JamesJansson:
Update to #2149


xavdid-stripe:
Hey James, appreciate the followup. 😄 Unfortunately, this has the same issue as your other PR - it allows for present-but-undefined values, which `metadata` doesn't have:

```ts
interface MetadataInterface { [name: string]: string }
const o: MetadataInterface = { key: undefined }; // correctly errors

type MetadataType = Partial<Record<string, string>>;
const o2: MetadataType = { key: undefined }; // doesn't error, but should
```

TS types for object don't pretend to have a key for every for every possible value- no JS object has every single possible string value as its set of keys. Instead, they describe the types for the present values. In our case, every present value is a `string`, so the types are correct as is.

Let me know if I'm not explaining this well and I'll do what I can to clear it up.

Thank you!

PR #2168, opened 2024-09-05
Remove qs package and implement URLSearchParams serializer

QuentinLemCode:
Hello

This pull request removes the qs package and replaces it with a custom implementation of URLSearchParams serializer.

The `stringifyRequestData` function now serializes an object or string into URL-encoded parameters, accommodating nested objects and arrays.

The goal of this change is to remove the `qs` dependency which was used for this single function. 
It will allow to reduce the size of the stripe-node library.

URLSearchParams is available in all evergreen browser, NodeJS since v10, Deno since v1, and Bun.

Source : https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams
https://bun.sh/docs/runtime/nodejs-apis#urlsearchparams

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,8 +55,7 @@
     "nanoid": "^3.2.0"
   },
   "dependencies": {
-    "@types/node": ">=8.1.0",
-    "qs": "^6.11.0"
+    "@types/node": ">=8.1.0"
   },
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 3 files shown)


prathmesh-stripe:
Thanks for the PR. Closing this in favor of https://github.com/stripe/stripe-node/pull/2168

vytskalt:
Closing #2168 in favor of #2168?

43081j:
#2558 for anyone wondering what the real one is

Issue #1274, opened 2021-10-18
@types/node in direct dependencies


berhalak:
Hi,
Is there any reason why package @types/node is included as a direct dependency and not as a devDependencies in package.json?

richardm-stripe:
Anyway I think I'm going to close this out as we are not going to take any action, but I wanted to ask: does the presence of `@types/node` in `dependencies` cause a problem for you?

PR #2169, opened 2024-09-05
Put @types/node as an optional peerDependency

QuentinLemCode:
Hello

This PR put @types/node as an optional peerDependency.

There was already previous attempt to move it to "devDependency" in order to avoid the size overhead of node types for Javascript project or even production builds.
https://github.com/stripe/stripe-node/issues/1274

However, this would cause issue because the Typescript compiler wouldn't provide correct error types.

In this PR, the `@types/node` library as been put in devDependencies and in optional peerDependencies.

This allow : 
- To have the node types available when developing on the stripe-node library
- To provide correct type error for typescript project using stripe-node (assuming project already have @types/node installed)
- To remove the @types/node dependency for javascript project (reducing size overhead)


QuentinLemCode:
This will probably conflict with #2168

xavdid-stripe:
Sorry it's been so long since we looked at this! I believe optional peer + dev is the correct path forward, so thank you for suggesting that.

It looks like the CI is failing on latest because the missing `undici-types`. If you've got time to debug that it would be great, or we can take a look soon(ish)

QuentinLemCode:
Hello @xavdid-stripe 
I have fixed the missing package, could relaunch the CI please ?

QuentinLemCode:
@xavdid-stripe Ready to merge :)

xavdid-stripe (approved):
perfect, thank you!

PR #2189, opened 2024-09-26
Add raw request

prathmesh-stripe:
## Changelog
Adds the ability to make requests to the Stripe API that are not directly supported in the SDK, by providing an HTTP method, url and relevant parameters

Example:
```node
import Stripe from 'stripe';
const stripe = new Stripe('sk_test_...');

const response = await stripe.rawRequest(
  'POST',
  '/v1/beta_endpoint',
  { param: 123 },
  { apiVersion: '2022-11-15; feature_beta=v3' }
);

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,40 @@ const stripe = new Stripe('sk_test_...', {
 });
 ```
 
+### Custom requests
+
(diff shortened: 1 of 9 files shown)


prathmesh-stripe:
Not sure why tests aren't working. They do pass locally.

helenye-stripe (approved):

PR #2190, opened 2024-09-27
Add related_object field to event types, remove fetchRelatedObject() …

ramya-stripe:
…that didnt work

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -8,6 +8,7 @@
 ///<reference path='./EventTypes.d.ts' />
 ///<reference path='./UpcomingInvoices.d.ts' />
 ///<reference path='./Deprecations.d.ts' />
+///<reference path='./V2/Events.d.ts' />
 // Imports: The beginning of the section generated from our OpenAPI spec
 ///<reference path='./AccountLinksResource.d.ts' />
 ///<reference path='./AccountSessionsResource.d.ts' />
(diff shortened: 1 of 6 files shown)


helenye-stripe (dismissed):

helenye-stripe (dismissed):
Can we add a unit test here to show intended use instead? That would've caught the previous failure.

types/index.d.ts
```diff
@@ -491,6 +493,90 @@ declare module 'stripe' {
...
     ): Promise<Stripe.Response<unknown>>;
+
+    /**
+     * Parses webhook event payload into a ThinEvent and verifies webhook signature.
+     * To get more information on the event, pass the id from the returned object to
+     * `stripe.v2.core.events.retrieve()`
+     *
+     * @throws Stripe.errors.StripeSignatureVerificationError
+     */
+    parseThinEvent: (
```
xavdid-stripe, line 504: I'm having trouble understanding why this file is different than the `.d.ts` here: [stripe/stripe-node-next | Types.d.ts#L151-L166](https://github.com/stripe/stripe-node-next/blob/db30eea2a17487811daf1b5cc380bea69e1a7ce8/src/Types.d.ts#L151-L166)

It seems like we only need one or the other

xavdid-stripe (commented):
Also, we should probably document the difference in usage of `relatedObject` in the readme and/or the docs snippets - it'll be the only one that doesn't follow the wider pattern, which could be confusing

ramya-stripe:
@helenye-stripe I wanted to add tests that test the type for column C has the `data` and `related_object` fields. But there is a broader problem with types that I have described in https://jira.corp.stripe.com/browse/DEVSDK-2192 which @xavdid-stripe is helping to figure out.

Added https://github.com/stripe/stripe-node/pull/2190/commits/6f72bb2afffe29387783a683d5e6df646523e4f0 - but like I said earlier this test will pass even without the changes in this PR

Can we unblock this PR?

types/index.d.ts
ramya-stripe: Fixed in https://github.com/stripe/stripe-node/pull/2190/commits/628b5d37d1501b0ea93eafbf546ce55420e23fc9

xavdid-stripe (approved):

PR #2201, opened 2024-10-04
Add fetchRelatedObject to V2 Events if needed

jar-stripe:
### Why?
Some V2 Event types have a stripe object or resource that is related to the event (e.g a billing meter for `v1.billing.meter.error_report_triggered`). Most SDKs add a fetchRelatedObject method to the event objects that support it as a means of easily retrieving this object.  The Node SDK launched V2 events without this method.  This PR adds this method back and updates the examples to match.

### What?
- converted V2/Core/Events.ts to be manually maintained
- added `addFetchRelatedObjectIfNeeded` to inspect the pulled event object and attach a fetchRelatedObject method to query the object specified in `related_object` if it is present; this reuses the Events resource to bind `this` to stripeMethod
- implemented `transformResponseData` for both list and retrieve, to process each event returned from the server
- updated EventTypes generated types to include fetchRelatedObject for the event that has one
- added tests, updated example


src/resources/V2/Core/Events.ts
```diff
@@ -1,12 +1,59 @@
-// File generated from our OpenAPI spec
-
+// This file is manually maintained
 import {StripeResource} from '../../../StripeResource.js';
+
 const stripeMethod = StripeResource.method;
+const RelatedObjects = StripeResource.extend({
+  fetchRelatedObject(pulledEvent: any) {
```
ramya-stripe, line 6: Why not add `fetchRelatedObject()` directly on `Events` which also extends from StripeResource instead of creating a whole new StripeResource?

jar-stripe: I wasn't able to 100% convince myself that there wasn't some state held in StripeResource that pointed to a "main" resource or base path (i.e. Events, /v2/core/events) so this felt safer.  I'm not married to this approach though, if we wanted to attach fetchRelatedObject directly to the Events resource.

ramya-stripe: I'd say lets test it out - it is one less hop to have

jar-stripe: Tested this morning, this works!  added a comment to explain us using the Events resource for fetchRelatedObject but I think this is resolved.

src/resources/V2/Core/Events.ts
```diff
@@ -1,12 +1,58 @@
...
 export const Events = StripeResource.extend({
-  retrieve: stripeMethod({method: 'GET', fullPath: '/v2/core/events/{id}'}),
-  list: stripeMethod({
-    method: 'GET',
-    fullPath: '/v2/core/events',
-    methodType: 'list',
-  }),
+  constructor: function(...args: any) {
+    StripeResource.apply(this, args);
+  },
```
ramya-stripe, line 9: I believe we needed to override the constructor coz we were setting the other resource on the Events resource. Since we dont do that anymore, we don't need to override the constructor right?

src/resources/V2/Core/Events.ts
```diff
@@ -1,12 +1,58 @@
...
+    };
+    return stripeMethod({
+      method: 'GET',
+      fullPath: '/v2/core/events',
+      methodType: 'list',
+      transformResponseData,
+    }).apply(this, args);
+  },
+
+  addFetchRelatedObjectIfNeeded(pulledEvent: any) {
```
ramya-stripe, line 36: Should we prefix the name of this function with `_` given that it will be visible to users when they inspect `stripe.v2.core.events`?

src/resources/V2/Core/Events.ts
jar-stripe: oh that may be true.  The very important call here is that StripeResource.apply.  Let me make sure that gets called in the case where it isnt overridden.

src/resources/V2/Core/Events.ts
jar-stripe: v2 is a `V2Resource`, in which core is a `CoreResource`, in which events is an `EventsResource` which doesn't have this method.  just checked in vscode and all I see are list and retrieve.  but I could put a comment over it that says this is a stripe internal function and should not be called directly or something like that?

src/resources/V2/Core/Events.ts
jar-stripe: deleted the constructor, the generic constructor already calls apply the way we want

src/resources/V2/Core/Events.ts
jar-stripe: I added a comment with a @private tag (I mimic'd another internal only function comment)

ramya-stripe: Yes, Typescript helps us here coz we won't be including this method in the types. What I meant was a JS user can still inspect the object in hand (list all properties or view in debug mode) and find it. Rare, but doable

ramya-stripe (approved):

PR #2202, opened 2024-10-04
Add build scripts to generate source maps

jar-stripe:
### Why?
When debugging into the Stripe SDK from a project or script, it can be useful to have the JS source maps so that you can step into and inspect the original Typescript source.  This PR adds build scripts into package.json to build the cjs and esm packages with source maps.

### What?
- added build-dev, build-esm-dev, and build-cjs-dev yarn build scripts; these are mirrors of build, build-esm, and build-cjs that also pass `--sourceMap true` to tsc


helenye-stripe (approved):

ramya-stripe (approved):

xavdid-stripe:
@jar-stripe I'm not sure if this PR is still something you want to move forward with, but it'll need to be recreated to add the flag to the justfile instead

jar-stripe:
> @jar-stripe I'm not sure if this PR is still something you want to move forward with, but it'll need to be recreated to add the flag to the justfile instead

done!  i think this is is still useful for at least us if/when we go to update the snippets or work on any test applications.  can you give it a quick look at thumbs up the justfile changes?

PR #2215, opened 2024-10-23
Remove empty resources created for service groupings

jar-stripe:
### Why?
The code generator had previously generated resources for V2 services that only contain other services, like V2.Billing, V2.Core, etc.  This is undesirable because it produces strangely named files and confusing empty types.  This PR updates contains generated code after modifying the code generator to treat these services like namespaces instead of resource-ful services.

This could be a breaking change as the CoreResource.d.ts and BillingResource.d.ts type definitions are being removed.  The following therefore would be a type error after this change:
```typescript
  const billing: Stripe.V2.BillingResource = client.v2.billing;
  const meters = await billing.meterEvents.create({
```
It does not change the access path through StripeClient or the interface of the services, though.

### What?
- removes src/resources/V2.ts, src/resources/Billing.ts, and src/resources/V2/Core.ts
- removes types/V2/BillingResource.d.ts, types/V2/CoreResource.d.ts, and types/V2Resource.d.ts
- adds services that used to be in these files to objects within src/resources.ts

### See also
http://go/j/DEVSDK-2196


helenye-stripe:
Can we rebase this off of master to remove the non-related changes? Other than that 👍

helenye-stripe (approved):

jar-stripe:
> Can we rebase this off of master to remove the non-related changes? Other than that 👍

Done!  But this includes a file that shouldn't be there (Margins.d.ts) and will continue to generate until we update the openapi version, so I think this probably needs to wait until that is resolved before we merge.

src/resources/V2/Billing.ts
```diff
@@ -1,16 +0,0 @@
...
-
-import {StripeResource} from '../../StripeResource.js';
-import {MeterEventSession} from './Billing/MeterEventSession.js';
-import {MeterEventAdjustments} from './Billing/MeterEventAdjustments.js';
-import {MeterEventStream} from './Billing/MeterEventStream.js';
-import {MeterEvents} from './Billing/MeterEvents.js';
-export const Billing = StripeResource.extend({
-  constructor: function(...args: any) {
-    StripeResource.apply(this, args);
-    this.meterEventSession = new MeterEventSession(...args);
```
xavdid-stripe, line 11: In this version, `Billing.meterEventSession` is an instance of `MeterEventSession`, but in your updated version it's the `MeterEventSession` itself. Is that intentional? does it change the way users can use the `v2` object?

jar-stripe: good question! are you referring to this code from src/resources.ts ?
```typescript
export const V2 = resourceNamespace('v2', {
  Billing: resourceNamespace('billing', {
    MeterEventAdjustments: V2BillingMeterEventAdjustments,
    MeterEventSession: V2BillingMeterEventSession,
    MeterEventStream: V2BillingMeterEventStream,
    MeterEvents: V2BillingMeterEvents,
  }),
  Core: resourceNamespace('core', {
    EventDestinations: V2CoreEventDestinations,
    Events: V2CoreEvents,
  }),
});
```

The `resourceNamespace` function accepts resources (aka sub services in this context) as an object and automatically camel cases the keys.  Apologies, I thought I had a test script in a comment here but I think I only included it in the codegen PR.  Here's the script I tested with:
```typescript
const apiKey = ...
const customerId = ...
const meterObjectId = ...

(async function() {
  const client = new Stripe(apiKey);
  const meters = await client.v2.billing.meterEvents.create({
    event_name: 'api_requests',
    payload: {
      stripe_customer_id: customerId,
      value: '27',
    },
  });
  console.log(meters);
  const events = await client.v2.core.events.list({
    object_id: meterObjectId,
  });
  console.log(events);
  const dests = await client.v2.core.eventDestinations.list();
  console.log(dests);
})();
```

jar-stripe: Although, as noted above though, it does technically break if a user uses that Stripe.V2.BillingResource type to hold the empty resource we're now deleting.

xavdid-stripe: Ok, I think that makes sense. This seems like a good improvement!

jar-stripe: Thanks!  Resolving this conversation and merging.

PR #2217, opened 2024-10-30
Update changelog with a note on 2024-10-28.acacia API version

ramya-stripe:
We should always talk about which API version is pinned in which SDK version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 ## 17.3.0 - 2024-10-29
-* [#2204](https://github.com/stripe/stripe-node/pull/2204) Update generated code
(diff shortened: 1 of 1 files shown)


prathmesh-stripe (approved):

helenye-stripe (approved):

joergbaier:
@ramya-stripe looks like `LatestApiVersion` was not updated to "2024-10-28.acacia"

xavdid-stripe:
@joergbaier this was fixed in https://github.com/stripe/stripe-node/pull/2218. Given that the runtime value is correct, we'll include this in the next release rather than cutting one now, out of band.

Thanks for flagging!

ramya-stripe:
Thanks for reporting this @joergbaier 
We just shipped a patch update (17.3.1) containing the fix

PR #2218, opened 2024-10-30
Updated API version in types/lib.d.ts

prathmesh-stripe:
1. Updated the API version regex in the `updateApiVersion,.js` script to handle new format of API versions(eg. `2024-09-30.acacia`). This script is run every time we run `codegen-format`.
2. Cleanup of lines that tried to update API version in files(Readme and package.json) that no longer reference an API Version.

### DIFF
--- a/scripts/updateAPIVersion.js
+++ b/scripts/updateAPIVersion.js
@@ -8,7 +8,7 @@ const read = (file) => fs.readFileSync(path.resolve(file)).toString();
 const write = (file, str) => fs.writeFileSync(path.resolve(file), str);
 const edit = (file, cb) => write(file, cb(read(file)));
 
-const API_VERSION = '2[0-9][2-9][0-9]-[0-9]{2}-[0-9]{2}';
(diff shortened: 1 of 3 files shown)


xavdid-stripe (approved):

tinotaylor:
I am 17.3.1 but still getting this error. I just upgrade my dashboard to the latest version:

`Type '"2024-10-28.acacia"' is not assignable to type '"2022-08-01"'.ts(2322)
lib.d.ts(72, 7): The expected type comes from property 'apiVersion' which is declared here on type 'StripeConfig'
(property) Stripe.StripeConfig.apiVersion?: "2022-08-01"
This library's types only reflect the latest API version.

We recommend upgrading your account's API Version to the latest version if you wish to use TypeScript with this library.

If you wish to remain on your account's default API version, you may pass null or another version instead of the latest version, and add a @ts-ignore comment here and anywhere the types differ between API versions.

@docs — https://stripe.com/docs/api/versioning

@docs — https://stripe.com/docs/api/versioning`

xavdid-stripe:
@tinotaylor you probably need to `npm install` again.

Either way, please open a new issue rather than commenting on a PR and we'll be happy to take a look!

PR #2220, opened 2024-10-31
Missing Required Parameters:

omkarkulkarni2704:
## Problem: The required parameter currency is missing.
* In the exampleFunction, you are missing the required currency parameter. This will result in an InvalidRequestError.
* Line No: 104 - You added Currency: 'usd', but in Line No: 123 - You didn't add Currency: 'usd', the currency parameter.

### DIFF
--- a/testProjects/mjs/index.js
+++ b/testProjects/mjs/index.js
@@ -120,7 +120,7 @@ async function exampleMonkeyPatchFunction() {
 
 exampleFunction({
   // The required parameter currency is missing
-  currency: 'usd',
+  currency: 'usd', 
   amount: 2000,
   confirm: true,
   payment_method: 'pm_card_visa',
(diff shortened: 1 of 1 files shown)


omkarkulkarni2704:
* Code:-

try {
    await stripe.paymentIntents.create({
      currency: 'usd',
      amount: 2000,
      confirm: true,
      payment_method: 'pm_card_visa',
    });

exampleFunction({
  // The required parameter currency is missing
  amount: 2000,
  confirm: true,
  payment_method: 'pm_card_visa',

ramya-stripe:
Hey @omkarkulkarni2704 

It is intentional that we are not passing the currency there. We want to test that when we do that `StripeInvalidRequestError` gets throwm

omkarkulkarni2704:
Ok

Issue #2207, opened 2024-10-09
Neither apiKey nor config.authenticator provided


chaosLegacy:
### Describe the bug

When installing either the version 17.0.0 or 17.1.0
in a brand new next.js 14 from scratch after initializing stripe

```
import Stripe from 'stripe';

export const stripe = new Stripe(process.env.STRIPE_SECRET_KEY as string, {
    apiVersion: '2024-06-20',
    typescript: true
})
```

I get the following error: Neither apiKey nor config.authenticator provided
This issue is not seen after downgrading to 16.12.0

### To Reproduce

Create a next.js project
adding stripe version 17.1.0
add a lib folder with a new file stripe.ts
with the following:
```
import Stripe from 'stripe';

export const stripe = new Stripe(process.env.STRIPE_SECRET_KEY as string, {
    apiVersion: '2024-09-30.acacia',
    typescript: true
})
```
run the project
getting this error: Neither apiKey nor config.authenticator provided

### Expected behavior

No error should be seen in either version 17.1.0 and 17.0.0
process.env.STRIPE_SECRET_KEY is well-defined, and I can console.log it value.

### OS

macos

### Node version

v20.11.0

### Library version

v17.1.0

### API version

2024-09-30.acacia

prathmesh-stripe:
I could not reproduce this in stripe-node 17.1.0. Stripe-node seems to be working fine with Nextjs. Here is what my code and  result looks like. 

### Console
Logged API Key and stripe object.
<img width="291" alt="Screenshot 2024-10-10 at 9 50 22 AM" src="https://github.com/user-attachments/assets/eddf56f9-323a-4a56-9561-2929869bde64">

### Code and file organization
<img width="645" alt="Screenshot 2024-10-10 at 10 01 14 AM" src="https://github.com/user-attachments/assets/58c53c84-5caf-4329-b972-95d86b88c2c2">
<img width="681" alt="Screenshot 2024-10-10 at 10 01 02 AM" src="https://github.com/user-attachments/assets/f11fdd3f-96c1-4fbe-b57d-bcf07e2b8791">

We would need more information to reproduce your error.

yafkari:
I have the same error when switching from stripe 16.12.0 to 17.2.0. (I'm also using NextJS 14)

The issue occurs in my case when calling "loadStripe" from a client component with [stripe-js](https://github.com/stripe/stripe-js) (4.8.0)

[...]

PR #2238, opened 2024-12-12
add missing key warning to README

xavdid-stripe:
fixes https://github.com/stripe/stripe-node/issues/2207 by adding more information to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,39 @@ const customer = await stripe.customers.create({
 console.log(customer.id);
 ```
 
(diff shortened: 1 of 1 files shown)


jar-stripe:
not feedback on the PR itself: since the "fixes" comment will auto close that issue, it might be a good idea to follow up on the issue with a link to the updated readme.  there are enough folks on that thread to where some explicit signposts would be well received.

jar-stripe (approved):
excellent additions to the readme!

PR #2249, opened 2025-01-11
add justfile, remove coveralls, tweak ESLint

xavdid-stripe:
### Why?

In an effort to modernize and simplify our local tooling, we're moving our dev commands from makefiles to [justfiles](https://github.com/casey/just). This is intended to be mostly a drop-in replacement, but some command names may change per standardization efforts.

To be in line with our other SDKs, I removed most of the scripts in `package.json`. But, if we find that potential contributors are having trouble with muscle memory, we can re-add common ones (`test`, `lint`, etc) that call into `just`.

In order to untangle formatting and linting (which [Prettier recommends](https://prettier.io/docs/en/integrating-with-linters.html#notes) keeping separate, I removed `plugin:prettier` from the eslint config. This mean the formatting-related lint rules are disabled, deferring to `prettier` instead. This has the added bonus of formatting violations not showing up as linting errors in-editor!

Lastly,  I've moved the integration test suite out of the main test suite. It starts a bunch of processes that don't get canceled (eating up CPU locally). So now they're run _only_ in CI by default (and via the `just integrations-test`, if you really want).

### What?

- add `justfile`
- remove coveralls
- tweak CI to use just
- move prettier out of eslint
- add a prettierignore (since prettier no longer gets to piggyback on eslint's file selection and we're on a super old version that doesn't respect `.gitignore`)
- update README

### See Also

[DEVSDK-2325](https://go/j/DEVSDK-2325)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -242,17 +242,19 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2018,
   },
-  plugins: ['prettier', 'import'],
-  extends: ['plugin:prettier/recommended'],
+  plugins: ['import'],
+  // disable formatting rules - prettier will handle this itself
+  extends: ['prettier'],
   overrides: [
     {
       files: ['**/*.ts'],
-      plugins: ['@typescript-eslint', 'prettier'],
+      plugins: ['@typescript-eslint'],
       extends: [
         'eslint:recommended',
         'plugin:@typescript-eslint/eslint-recommended',
         'plugin:@typescript-eslint/recommended',
-        'plugin:prettier/recommended',
+        // disable formatting rules - prettier will handle this itself
+        'prettier',
       ],
       rules: {
         '@typescript-eslint/no-use-before-define': 0,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   build:
-    name: Build
+    name: 'Static Checks'
     runs-on: 'ubuntu-24.04'
 
     steps:
@@ -40,9 +40,6 @@ jobs:
           key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-yarn-
-
-      - name: Compile JS -> TS
-        run: just install build
 
       - name: Verify Linting
         run: just lint-check
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,6 +54,7 @@ jobs:
     name: Test (${{ matrix.node }})
     needs: [build]
     strategy:
+      fail-fast: false
       matrix:
         os:
           - 'ubuntu-latest'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,13 +17,13 @@ on:
       - sdk-release/**
       - feature/**
 
-
 jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
 
     steps:
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@v2
 
       - name: Setup node
@@ -41,11 +41,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      - name: Build Typescript
-        run: yarn && yarn build
+      - name: Compile JS -> TS
+        run: just install build
 
-      - name: Lint
-        run: yarn lint
+      - name: Verify Linting & Formatting
+        run: just lint-check format-check
 
   test:
     name: Test (${{ matrix.node }})
@@ -53,18 +53,19 @@ jobs:
     strategy:
       matrix:
         os:
-          - "ubuntu-latest"
+          - 'ubuntu-latest'
         node:
           # should include even numbers >= 12
           # see: https://nodejs.org/en/about/previous-releases
-          - "22"
-          - "20"
-          - "18"
-          - "16"
-          - "14"
-          - "12"
+          - '22'
+          - '20'
+          - '18'
+          - '16'
+          - '14'
+          - '12'
     runs-on: ${{ matrix.os }}
     steps:
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@v2
 
       - name: Setup node
@@ -75,6 +76,7 @@ jobs:
       - name: Print Node.js version
         run: node -v
 
+      # used for one of the integration tests
       - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
@@ -98,15 +100,7 @@ jobs:
       - uses: stripe/openapi/actions/stripe-mock@master
 
       - name: Test
-        run: make ci-test
-
-      - name: Coveralls
-        run: yarn report && yarn coveralls
-        if: env.COVERALLS_REPO_TOKEN && matrix.node == '18'
-        env:
-          GITHUB_TOKEN: ${{ secrets.github_token }}
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
-          COVERALLS_FLAG_NAME: node-${{ matrix.node }}
+        run: just ci-test
 
   publish:
     name: Publish
@@ -117,6 +111,8 @@ jobs:
     needs: [build, test]
     runs-on: ubuntu-latest
     steps:
+      # just is called in `yarn prepack`, which is called during the `publish` operation
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@v2
       - run: sudo apt-get install -y oathtool
       - name: Publish to NPM
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: 'ubuntu-24.04'
 
     steps:
       - uses: extractions/setup-just@v2
@@ -57,7 +57,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - 'ubuntu-latest'
+          - 'ubuntu-24.04'
         node:
           # should include even numbers >= 12
           # see: https://nodejs.org/en/about/previous-releases
@@ -114,7 +114,7 @@ jobs:
       startsWith(github.ref, 'refs/tags/v') &&
       endsWith(github.actor, '-stripe')
     needs: [build, test]
-    runs-on: ubuntu-latest
+    runs-on: 'ubuntu-24.04'
     steps:
       # just is called in `yarn prepack`, which is called during the `publish` operation
       - uses: extractions/setup-just@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,7 +97,8 @@ jobs:
         id: yarn-cache
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          # searching very deep deps can time out, so only cache on the root yarn.lock
+          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
           restore-keys: |
             ${{ runner.os }}-yarn-
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,8 +44,11 @@ jobs:
       - name: Compile JS -> TS
         run: just install build
 
-      - name: Verify Linting & Formatting
-        run: just lint-check format-check
+      - name: Verify Linting
+        run: just lint-check
+
+      - name: Verify Formatting
+        run: just format-check
 
   test:
     name: Test (${{ matrix.node }})
--- a/README.md
+++ b/README.md
@@ -595,16 +595,7 @@ New features and bug fixes are released on the latest major version of the `stri
 
 ## Development
 
-Run all tests (installing the dependencies first, if needed)
-
-```bash
-$ just install
-$ just test
-```
-
-If you do not have `yarn` installed, consult its [installation instructions](https://classic.yarnpkg.com/lang/en/docs/install/).
-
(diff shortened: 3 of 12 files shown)


.eslintrc.js
```diff
@@ -242,17 +242,19 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2018,
   },
-  plugins: ['prettier', 'import'],
-  extends: ['plugin:prettier/recommended'],
```
xavdid-stripe, line 246: the prettier plugin _enables_ formatter rules, while the config disables it. We want only the latter.

.github/workflows/main.yml
```diff
@@ -41,30 +41,31 @@ jobs:
...
     needs: [build]
     strategy:
       matrix:
         os:
-          - "ubuntu-latest"
+          - 'ubuntu-latest'
         node:
           # should include even numbers >= 12
           # see: https://nodejs.org/en/about/previous-releases
-          - "22"
```
xavdid-stripe, line 60: prettier auto-formatting

README.md
```diff
@@ -596,14 +595,14 @@ New features and bug fixes are released on the latest major version of the `stri
...
 
 ```bash
-$ yarn install
-$ yarn test
+$ just install
+$ just test
 ```
 
-If you do not have `yarn` installed, you can get it with `npm install --global yarn`.
+If you do not have `yarn` installed, consult its [installation instructions](https://classic.yarnpkg.com/lang/en/docs/install/).
```
xavdid-stripe, line 605: this ensures people can follow whatever the latest guidance is (like installing from a non-js package manager)

justfile
```diff
@@ -0,0 +1,80 @@
...
+_default:
+    just --list --unsorted
+
+# this uses positional-args so that mixed quoted and unquoted arguments
+# (like filtering for a certain test) work the way we expect
+
+# ⭐ run unit tests
+[positional-arguments]
+test *args: build
+    mocha "$@"
```
xavdid-stripe, line 17: This was surprisingly tricky get right! By default, just groups all the args you pass to `*args`, so it wasn't correctly preserving `-g "some thing"`. This works as expected for mixed content

justfile
```diff
@@ -0,0 +1,80 @@
...
+
+# try to compile the example TS file to make sure exports work
+types-test: build
+    tsc --build types/test
+
+# run full integration tests by installing a bunch of packages and starting servers (slow)
+integrations-test: build
+    RUN_INTEGRATION_TESTS=1 mocha test/Integration.spec.ts
```
xavdid-stripe, line 25: I'd rather call `just test` or the `test` recipe as a pre-req, but it doesn't work well with env vars. Mocha doesn't support pytest-style marks to conditionally enable certain suites, so I settled for an env var. We'll have to update this line if/when we move from mocha, but I'm fine with that.

justfile
```diff
@@ -0,0 +1,80 @@
...
+
+# reinstall dependencies, if needed
+install:
+    yarn {{ if is_dependency() == "true" { "--silent" } else { "" } }}
+
+[no-exit-message]
+[private]
+prettier *args:
+    # all the project-relevant JS code
+    prettier "{src,examples,scripts,test,types}/**/*.{ts,js}" {{ args }}
```
xavdid-stripe, line 59: This is the folders eslint cared about, plus some extras (that are already formatted)

justfile
```diff
@@ -0,0 +1,80 @@
...
+
+# `format` needs to install since other scripts run it cold
+
+# ⭐ format all files
+format: install (prettier "--write --loglevel silent")
+    # ensure other files reflect the version in package.json
+    ./scripts/updateAPIVersion.js
```
xavdid-stripe, line 66: This ran as part of the `fix` command before, so I guess run it during the format that codegen runs 🤷

scripts/updateAPIVersion.js
```diff
@@ -12,7 +17,7 @@ const API_VERSION = '2[0-9][2-9][0-9]-[0-9]{2}-[0-9]{2}.[a-z]+';
 
 const main = () => {
   const matches = [
-    ...read('src/apiVersion.ts').matchAll(/ApiVersion . '([^']*)'/g),
```
xavdid-stripe, line 15: I was looking at this file anyway and it didn't make sense that this was _any_ character - we're only looking for `=`.

test/Integration.spec.ts
```diff
@@ -4,6 +4,15 @@ import {FAKE_API_KEY} from './testUtils.js';
 const nodeVersion = parseInt(process.versions.node.split('.')[0], 10);
 
 describe('Integration test', function() {
+  // these tests are expensive and start processes they don't clean up
```
xavdid-stripe, line 7: This is why your CPU goes to 100% when running the test suite locally!

package.json
```diff
@@ -60,19 +59,7 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "yarn build-esm && yarn build-cjs",
```
jar-stripe, line 63: my two concerns here: 1) i do think contributors will expect to be able to run yarn ___ for at least the most common scripts, and 2) we do lose a bit of our own DX in that scripts automatically scope in commands installed into node_modules (vs having to resolve them in an external script)

xavdid-stripe: Makes sense! Like I said in the PR description, I'm happy to include them. I _sort_ of prefer to not include them for now and wait until we get feedback saying "hey, where are the package.json scripts?!" to better gague interest, but it's a small change to add them now and not a hill I'd die on.

To your second point, this line takes care of the same in the justfile: https://github.com/stripe/stripe-node/pull/2249/files/17e620dea6a451d1ece631930533c1ce46184821#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1R6

So recipes can use `mocha` and `eslint` just like we can inside the `package.json` script. But we don't have to `&&` chain everything, which is nice.

.github/workflows/main.yml
```diff
@@ -17,13 +17,13 @@ on:
       - sdk-release/**
       - feature/**
 
-
 jobs:
   build:
     name: Build
```
xavdid-stripe, line 22: I'm not quite sure why we need this step as a pre-req for the others? State isn't carried between jobs, so the `build` we do here is only used in checking linting and formatting. If those pass, then each node version does its _own_ build and then runs the tests.

My _guess_ is that that we use to fail lint/formatting a lot? so we made sure those were good before kicking off the test jobs? We can keep that pattern, but we should maybe rename this step for clarity. I didn't include a built step in `test` before because I thought "well it has a built step, so surely it's all set?".

.github/workflows/main.yml
```diff
@@ -91,22 +97,15 @@ jobs:
         id: yarn-cache
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          # searching very deep deps can time out, so only cache on the root yarn.lock
```
xavdid-stripe, line 100: Was getting weird (flaky?) test failures (e.g. https://github.com/stripe/stripe-node/actions/runs/12720008648/job/35461087788), so I'm tweaking this per suggestions in https://github.com/actions/runner/issues/449

.github/workflows/main.yml
```diff
@@ -41,30 +41,35 @@ jobs:
...
-      - name: Lint
-        run: yarn lint
+      - name: Verify Formatting
+        run: just format-check
 
   test:
     name: Test (${{ matrix.node }})
     needs: [build]
     strategy:
+      fail-fast: false
```
xavdid-stripe, line 57: Seeing if exactly one version fails (like [this run](https://github.com/stripe/stripe-node/actions/runs/12719974592), where _somehow_ only node 14 couldn't run that script for deno??) is useful, especially as our matrix gets large.

package.json
xavdid-stripe: Ok, i've added the `test` command and other non-just instructions to the readme! That should get non-just folks started and we'll reassess if needed.

.github/workflows/main.yml
jar-stripe: yea, I can see this being a "make sure it builds at all" step before we build across the different versions.  Keeping it but renaming it makes sense to me, maybe something like "Verify code will build" or something?

.github/workflows/main.yml
jar-stripe: imho as long as the steps run in parallel, the only reason to not do this is CI cost; if that's not a driver, lets get as much info out as we can.

.github/workflows/main.yml
```diff
@@ -17,13 +17,13 @@ on:
       - sdk-release/**
       - feature/**
 
-
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: 'ubuntu-24.04'
```
jar-stripe, line 23: might be worth a comment at the top explaining the explicit version, especially because this is one LTS later than e.g. ruby or dotnet.  maybe we have a consistent build environment description across the top of each sdk's ci file?

.github/workflows/main.yml
jar-stripe: is this because its finding yarn.lock files inside node_modules?  i agree, that seems superfluous because the root lock file should contain the whole dependency graph

.github/workflows/main.yml
xavdid-stripe: CI for open source, so no concerns there!!

.github/workflows/main.yml
xavdid-stripe: I like the second one more - or an internal doc or something explaining the version choices. 

Let's call it out of scope for this PR but I'll write the doc & associated jira tickets

test/Integration.spec.ts
jar-stripe: is there a fix we can make here?  or is it more complicated than can be fixed as part of this PR?

scripts/updateAPIVersion.js
jar-stripe: is the apiVersion.ts is always automatically updated (never manually changed)?  if its ever manually touched, consider adding \s* on either side of the equals here just in case someone forgets to use the correct amount of white space.

.github/workflows/main.yml
xavdid-stripe: done!

justfile
jar-stripe: what happens to the *args, or are the args represented in both *args and $@?  and if thats true, what happens if test is called from another just rule?

justfile
jar-stripe: if you `export RUN_INTEGRATION_TESTS=1` it will be available to child processes and shells, which may let you reuse the test recipe here

justfile
jar-stripe: this is a little weird that it is part of format, and that comment is confusing.  the api version isnt in package.json, right?

justfile
xavdid-stripe: Because of the way `just` expands input args. It treats them as one big group for quoting purposes, unless told otherwise. It becomes an issue when some of your args are quoted with spaces and some aren't. You can't quote the whole thing, but you also can't leave them unquoted. This is important when wanting to call `mocha someFile.js -g "match this name"`. If the args aren't grouped correctly when passing through `just`, you call `-g match` and then `this` and `name` are extra args.

I did some testing with the following:

```
[positional-arguments]
arg-test *args:
    bash -c 'while (( "$#" )); do echo - $1; shift; done' -- "{{ args }}"
```

Which _should_ print each arg (respecting spaces) on their own line.

`"{{ args }}"`:

```
% .j arg-test a b "c d" e
- a b c d e
```

`{{ args }}`:

```
- a
- b
- c
- d
- e
```

`"$@"`:

```
% .j arg-test a b "c d" e
- a
- b
- c d
- e
```

I mostly pulled from the docs: https://github.com/casey/just?tab=readme-ov-file#positional-arguments

justfile
xavdid-stripe: Unfortunately, lines of `just` recipes can't share state by default:

```
env-test:
    export MY_VAR=asdf
    echo "after: $MY_VAR"
```

gives:

```
sh: MY_VAR: unbound variable
error: Recipe `env-test` failed on line 26 with exit code 127
```

We could write it as a shebang recipe:

```
env-test:
    #!/usr/bin/env bash
    export MY_VAR=asdf
    echo "after: $MY_VAR"
```

but that wouldn't extend to the other recipe. 

The real solution here is probably to not use an env var or split the integration tests into a totally separate suite, but i'm fine with this for now

scripts/updateAPIVersion.js
xavdid-stripe: It's always automatic! it's handled by the openapi bot: https://github.com/stripe/stripe-node/commit/d727b3e1af9ac45cc66d919f6d55b53ead143a3b#diff-4d0377c86ba281e8c4d873dbd1f843f67b6b459307ca1b52ada7b050da8094f5R3

justfile
xavdid-stripe: Weirdly it was always like this, just more abstracted: https://github.com/stripe/stripe-node/blob/2fda90ebc32034cfb56cb355331584faa4038732/package.json#L73

My understanding is that when openapi has a new API version, it has to propgate that string (e.g. `'2024-12-18.acacia'`) through the TS files. Purity wise, this should be run as cleanup after generation. But since the only command `generate` runs is `format`, having it be a cleanup in format made sense to me. You're right though, that this isn't _really_ part of formatting. It could be it's own recipe and we could change `format` to be `format (prettier ...) update-api-version` if you prefer. I don't have strong feelings.

jar-stripe: I think in this case, if its not too much extra work, I would prefer to have update-api-version handle this and change format to be the way you describe.  ROI seems high on that, considering we'll likely forget this nuance 6 months from now.

test/Integration.spec.ts
xavdid-stripe: It's fixable, but IMO out of scope of this PR. I didn't want to get _too_ distracted. Plus we don't care if we're leaking on CI, since it all gets shut down when the job completes

justfile
xavdid-stripe: no problem; done! Added some extra comments, too.

jar-stripe (approved):
Looks good, and I love the paper trail we've left here !

PR #2277, opened 2025-03-21
[BREAKING] swap Stripe.webhooks from a factory function to a static property

xavdid-stripe:
### Why?

In https://github.com/stripe/stripe-node/issues/1956, a user flagged that they wanted to be able to use webhook utilities before creating a `Stripe` instance. Since the webhook utilities were standalone (didn't depend on anything inside `Stripe`) we made them static. But because they were factory functions before, we didn't want to break that mid-cycle. Now that it's time for a major version, we can complete this change.

### What?

- set `Stripe.webhooks` to the result of `createWebhooks()` directly
- set `this.webhooks` inside the `Stripe` constructor to the value at `Stripe.webhooks`

### See Also

- [DEVSDK-1702](https://go/j/DEVSDK-1702)
- https://github.com/stripe/stripe-node/issues/1956

## Changelog

- ❗ `Stripe.webhooks` and `Stripe().webhooks` are no longer functions (just plain objects)
    - if you were calling `Stripe.webhooks().someMethod()`, you should use `Stripe.webhooks.someMethod()` instead
    - there shouldn't be user-facing functionality changing, just a slight breaking change in our internal organization

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -9,7 +9,7 @@ import {
   RequestData,
   RequestOptions,
 } from './Types.js';
-import {WebhookObject, WebhookEvent, createWebhooks} from './Webhooks.js';
+import {WebhookEvent, createWebhooks} from './Webhooks.js';
 import {ApiVersion} from './apiVersion.js';
 import {CryptoProvider} from './crypto/CryptoProvider.js';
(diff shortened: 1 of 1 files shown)


jar-stripe:
This feels like a really big breaking change; I think we should discuss before merging.

mbroshi-stripe (approved):

PR #2284, opened 2025-03-28
remove type checks for removed endpoints

xavdid-stripe:
### Why?

We had some manually maintained types that reference types that have now been removed, failing the build.

### What?

- removed unavailable types
- update test to use a still-present deprecated type


jar-stripe (approved):
LGTM

PR #2289, opened 2025-03-28
add Amount interface

xavdid-stripe:
### Why?

Some v2 APIs have the v2 `Amount` shape now, but that interface only existed in the private beta repos. This copies that interface over to clear CI failures.

### What?

- copy `Amount` interface

### See Also

- [DEVSDK-1265](https://go/j/DEVSDK-1265)


jar-stripe (approved):
LGTM

PR #2293, opened 2025-04-01
Generate beta SDKs for Open API version 1648

ramya-stripe:
### Why?
The generation of automated PRs for version 1648 of the OpenAPI spec is failing. Creating this PR manually to unblock the upcoming release which needs the version 1648

### What?
- Update OPENAPI_VERSION file to have `v1648`
- Run the codegen

No changelog entry as there are no net API changes after version 1646


xavdid-stripe (approved):

types/Subscriptions.d.ts
```diff
@@ -52,7 +52,7 @@ declare module 'stripe' {
       /**
        * Whether this subscription will (if `status=active`) or did (if `status=canceled`) cancel at the end of the current billing period.
        */
-      cancel_at_period_end: boolean;
+      cancel_at_period_end: boolean | null;
```
ramya-stripe, line 55: We have an override to prevent this change - not sure why this is coming in

types/Invoices.d.ts
```diff
@@ -409,6 +409,8 @@ declare module 'stripe' {
 
       status_transitions: Invoice.StatusTransitions;
 
+      subscription?: string | Stripe.Subscription | null;
```
ramya-stripe, line 412: We have an override to remove this field - not sure why this is coming in

PR #2300, opened 2025-04-08
Handle external_resource field

mbroshi-stripe:
### Why?

The `external_account` field should be a union type, not a `string`.

### What?

- Changes `external_account` field in `ExternalAccountsCreateParams` from a `string` to a union type.

## Changelog
- Changes `external_account` field in `ExternalAccountsCreateParams` from a `string` to a union type.

### DIFF
--- a/types/ExternalAccountsResource.d.ts
+++ b/types/ExternalAccountsResource.d.ts
@@ -6,7 +6,11 @@ declare module 'stripe' {
       /**
        * Either a token, like the ones returned by [Stripe.js](https://stripe.com/docs/js), or a dictionary containing a user's external account details (with the options shown below).
        */
-      external_account: string | ExternalAccountCreateParams.ExternalAccount;
+      external_account:
+        | string
+        | ExternalAccountCreateParams.Card
+        | ExternalAccountCreateParams.BankAccount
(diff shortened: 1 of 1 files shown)


types/ExternalAccountsResource.d.ts
```diff
@@ -24,6 +24,45 @@ declare module 'stripe' {
       metadata?: Stripe.MetadataParam;
     }
 
+    namespace ExternalAccountCreateParams {
+      interface ExternalAccount {
+        /**
+         * The type of external account. Should be bank_account.
```
xavdid-stripe, line 30: It seems like this PR didn't get the card create params like python did?

xavdid-stripe (commented):
This looks reasonable, but the changes are smaller than Python; wanted to double check that's intentional. It's 👍 otherwise!

types/ExternalAccountsResource.d.ts
mbroshi-stripe: Amazing catch! I'm not sure why the cards params didn't show up (I must have had a dirty local state when I ran `just gen`), but I just re-generated, and they're showing up now.

xavdid-stripe (approved):

PR #2304, opened 2025-04-15
revert lockfile changes

xavdid-stripe:


xavdid-stripe:
The breakage in master is unrelated to the dependency bumps that this PR reverted, so i'll close.

PR #2308, opened 2025-04-17
RawRequest can use custom host

prathmesh-stripe:
### Why?
Host and streaming option when passed through raw request are not honored while making the actual request.

### What?
- Host and streaming are now settable when passed as request options when making a raw request.

### See Also
https://jira.corp.stripe.com/browse/RUN_DEVSDK-1525

## Changelog
* RawRequest now allows you set `host` and `streaming` in request options. Eg.
```typescript
const file = await stripe.rawRequest(
  'GET',
  '/v1/files/file_123/contents',
  {},
  {host: 'files.stripe.com', streaming: true}
);
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -22,6 +22,7 @@ const OPTIONS_KEYS = [
   'authenticator',
   'stripeContext',
   'additionalHeaders',
+  'streaming',
 ];
 
 type Settings = {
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -217,6 +219,9 @@ export function getOptionsFromArgs(args: RequestArgs): Options {
           [headerName: string]: string;
         };
       }
+      if (params.streaming) {
+        opts.streaming = true;
+      }
```
mbroshi-stripe, line 224: Does it make more sense to do `opts.streaming = params.streaming`? I'm thinking of the case where someone sets `params.streaming = false`

mbroshi-stripe (approved):
I had one question, but otherwise looks good

PR #2354, opened 2025-06-24
Add stripeContext to RequestOptions interface

mbroshi-stripe:
### Why?

The `Stripe-Context` header can be used to set the account on behalf of whom an API call is made. See https://docs.corp.stripe.com/context for more information. We already support its use in StripeClient, but it is not exposed in the `RequestOptions` interface.

### What?

- Add `stripeContext` to the`RequestOptions` interface.

```javascript
const customer = await stripeClient.customers.create(
  {
    email: "jenny.rosen@example.com",
    name: "Jenny Rosen",
  },
  {
    stripeContext: "acct_xyz",
  }
);
```

## Changelog
- Add `stripeContext` to the `RequestOptions` interface.

### DIFF
--- a/types/lib.d.ts
+++ b/types/lib.d.ts
@@ -132,6 +132,11 @@ declare module 'stripe' {
       stripeAccount?: string;
 
       /**
+       * An account on whose behalf you wish to make a request. See https://docs.corp.stripe.com/context for more information.
+       */
+      stripeContext?: string;
+
+      /**
        * The [API Version](https://stripe.com/docs/upgrades) to use for a given request (e.g., '2020-03-02').
        */
       apiVersion?: string;


mbroshi-stripe:
> LGTM. Do we also need to add this in the [StripeConfig](https://github.com/stripe/stripe-node/blob/72392880cd226c07a40e2c9ba8a8f0d23025bf01/types/lib.d.ts#L114)?

I'm actually not sure! What is suppose to go in `StripeConfig`?

> And adding a test for this would be great.

There's already a test! It's just not typed, so there was no type error: https://github.com/stripe/stripe-node/blob/72392880cd226c07a40e2c9ba8a8f0d23025bf01/test/stripe.spec.ts#L655-L666

mbroshi-stripe:
@prathmesh-stripe You're right--there's [a test](https://github.com/stripe/stripe-node/blob/72392880cd226c07a40e2c9ba8a8f0d23025bf01/test/stripe.spec.ts#L607) for passing it into the config object as well, so I'll add to `StripeConfig` as well

PR #2370, opened 2025-07-11
Add strongly typed Thin Event push payloads

xavdid-stripe:
### Why?

As part of the Next Gen Event Handling project, we're adding fully typed push payloads to thin events. These describe the full union of thin events a user could receive. 

> Note: because one of our earlier prototypes was in node, most of this PR is adding that already-reviewed code to the mainline . See node-next `708` & `725` for historical discussion.

### What?

- generate push payload interfaces & union of those
- add `pull` and `fetchRelatedObject` methods to result of `parseThinEvent`
- change return type of `parseThinEvent` to use new union
- add tests
- add address to `getTestServerStripe` callback
- make headers inputs nullalbe
- add usage param to internal request sender
- update example

### TODO

- [ ] add `UknownThinEvent` and make sure type narrowing works nicely

### See Also

- [DEVSDK-2662](https://go/j/browse/DEVSDK-2662)

### DIFF
--- a/src/Types.d.ts
+++ b/src/Types.d.ts
@@ -69,9 +69,14 @@ export type RequestEvent = {
 };
 export type RequestHeaders = Record<string, string | number | string[]>;
 export type RequestOptions = {
-  settings: RequestSettings;
-  streaming: boolean;
-  headers: RequestHeaders;
+  settings?: RequestSettings;
+  streaming?: boolean;
+  headers?: RequestHeaders;
+  stripeContext?: string;
+  /**
+   * NOTE: prefer sending `stripeContext` instead of `stripeAccount` for new code. They're currently identical, but we will eventually discourage and (later) drop support for `stripeAccount`.
+   */
+  stripeAccount?: string;
 };
 export type RequestOpts = {
   authenticator: RequestAuthenticator | null;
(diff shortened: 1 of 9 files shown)


src/Types.d.ts
```diff
@@ -69,9 +69,14 @@ export type RequestEvent = {
 };
 export type RequestHeaders = Record<string, string | number | string[]>;
 export type RequestOptions = {
-  settings: RequestSettings;
-  streaming: boolean;
-  headers: RequestHeaders;
+  settings?: RequestSettings;
+  streaming?: boolean;
+  headers?: RequestHeaders;
+  stripeContext?: string;
```
helenye-stripe, line 75: ooc should we call this out in the changelog?

types/index.d.ts
```diff
@@ -558,7 +557,7 @@ declare module 'stripe' {
        * Optional: timestamp to use when checking signature validity. Defaults to Date.now().
        */
       receivedAt?: number
-    ) => Stripe.ThinEvent;
+    ) => Stripe.V2.EventNotification;
```
helenye-stripe, line 560: i don't remember this discussion, but we've decided these are "v2 event notifications"? Instead of generic no-namespace

helenye-stripe (approved):

src/Types.d.ts
xavdid-stripe: sure! this was just a UX improvement - these all shouldn't have to be specified.

PR #2375, opened 2025-07-18
Update merge script rule to run on every PR

helenye-stripe:
### Why?

We want to make the merge script rule that checks that a merge script PR uses `merge` instead of `squash` required. However, in branches where this doesn't apply it runs forever. Add this to make this run on every PR, but pass on branches that don't satisfy the `/merge-` format.

### See Also

https://jira.corp.stripe.com/browse/RUN_DEVSDK-1798

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -10,12 +10,15 @@ on:
 jobs:
   require_merge_commit_on_merge_script_pr:
     name: Merge script PRs must create merge commits
-    if: ${{ contains(github.head_ref, '/merge-') }}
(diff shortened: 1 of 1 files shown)


jar-stripe:
This is weird; I actually changed the rule in [sdk-codegen/templates/rules.yml.template](https://github.com/stripe/sdk-codegen/blob/master/templates/rules.yml.template) to do this and also to enforce squashed commits for other PRs; i thought I had rolled that out to all the repos but maybe not?

jar-stripe (approved):
👍

PR #2398, opened 2025-09-05
⚠️ Build SDK w/ V2 OpenAPI spec

xavdid-stripe:
### Why?

While SDK support for our v1 APIs comes from our [public openapi spec](https://github.com/stripe/openapi), our v2 endpoints have been sources from an internal dump of the protobuf definitions of those APIs. 

This PR marks our cutover point for our generation source: now we're using the (private, for now) V2 OpenAPI spec! There are a few changes/bugfixes as a result (documented below).

> note: entire PR is codegen

### What?

- generated SDK using v2 openapi spec.

### See Also

[DEVSDK-2750](https://go/j/DEVSDK-2750)

## Changelog
- ⚠️ The delete methods for v2 APIs (the ones in the `StripeClient.v2` namespace) now return a `V2DeletedObject` which has the id of the object that has been deleted and a string representing the type of the object that has been deleted.
- ⚠️ Change the types of nullable properties in objects returned by v2 APIs (the ones in the `StripeClient.v2` namespace) from a null union to optional (e.g. `prop: string | null` -> `prop?: string`).


test/resources/generated_examples_test.spec.js
```diff
@@ -3975,7 +3975,7 @@ describe('Generated tests', function() {
         method: 'GET',
         path: '/v2/core/events?object_id=object_id',
         response:
-          '{"data":[{"context":null,"created":"1970-01-12T21:42:34.472Z","id":"obj_123","object":"v2.core.event","reason":null,"type":"type","livemode":true}],"next_page_url":null,"previous_page_url":null}',
```
jar-stripe, line 3978: is this a proto <> openapi change?  Or is this just because the actual API is different?

test/resources/generated_examples_test.spec.js
```diff
@@ -4033,14 +4033,11 @@ describe('Generated tests', function() {
       {
         method: 'DELETE',
         path: '/v2/core/event_destinations/id_123',
-        response:
-          '{"created":"1970-01-12T21:42:34.472Z","description":"description","enabled_events":["enabled_events"],"event_payload":"thin","events_from":null,"id":"obj_123","metadata":null,"name":"name","object":"v2.core.event_destination","snapshot_api_version":null,"status":"disabled","status_details":null,"type":"amazon_eventbridge","updated":"1970-01-03T17:07:10.277Z","livemode":true,"amazon_eventbridge":null,"webhook_endpoint":null}',
+        response: '{"id":"abc_123","object":"some.object.tag","deleted":true}',
```
jar-stripe: maybe this is a small nit, but v2 deleted objects don't have a deleted flag, right?

types/index.d.ts
```diff
@@ -278,6 +278,7 @@
 ///<reference path='./V2/Billing/MeterEventAdjustments.d.ts' />
 ///<reference path='./V2/Billing/MeterEventSessions.d.ts' />
 ///<reference path='./V2/Billing/MeterEvents.d.ts' />
+///<reference path='./V2/DeletedObjects.d.ts' />
```
jar-stripe: very small nit but this is a weird pluralization

test/resources/generated_examples_test.spec.js
jar-stripe: oh, or is this because of the nullable -> optional change?

jar-stripe (approved):
left a couple of non-blocking comments

test/resources/generated_examples_test.spec.js
xavdid-stripe: that's correct (not yet anyway). I can fix this in codegen

test/resources/generated_examples_test.spec.js
xavdid-stripe: It's a format change thing- the openapi spec is more accurate. Specifically, the if the key is present, it should have a value (won't be present and null). I think `previous_page_url` may be treated differently, or our example spec generation has a minor bug.

types/index.d.ts
xavdid-stripe: good catch! that does seem weird. Probably an overzealous pluralization thing. I can probably fix.

test/resources/generated_examples_test.spec.js
xavdid-stripe: oh I remember this - I put that when I wrote the pattern review for "always include `deleted` for v2 response deletes". It's been stuck in review for ages. I'll comment out for now

PR #2404, opened 2025-09-15
generate private-preview SDK w/ mid Sept changes

xavdid-stripe:
### Why?

We're cutting a mid-month release of the private-preview SDKs that includes changes from September. This PR represents the latest private-preview + those changes, generated with the proto source.

This PR won't merge. It's just here as a record of what we released in this private-preview release.

### What?

- generated SDK using the second iteration of the `2025-08-27.preview` version. It also used protos as the source.


xavdid-stripe:
only codegen'd changes, so I just merged

Issue #2267, opened 2025-03-09
Stripe.SubscriptionSchedule.Phase type is missing trial property


alexiz10:
### Describe the bug

The `Stripe.SubscriptionSchedule.Phase` type in TypeScript is missing the `trial` property. I am currently using SubscriptionSchedules to sign up my users for a free 30-day trial of my product, then, I move them to the free tier. If a user wants to make changes to their subscription before the trial runs out, I have to know this since they can only make certain changes to the trial. To create the first phase, I am using `trial: true` and `end_date: _in 30 days_`, since this phase should only be a trial. When checking `currentPhase.trial`, TypeScript complains that `trial` isn't part of `Phase`. But when I log out my phases to the console, the first phase has `trial: true` and the rest of the phases just don't have `trial`, so this should be an optional property. Currently, I have to extend the `Phase` type into my own custom type and add `trial`.

### To Reproduce

1. Create a Subscription Schedule where the first phase is a trial phase

```
await stripe.subscriptionSchedules.create({
  customer: CUSTOMER_ID,
  start_date: Math.floor(Date.now() / 1000),
  phases: [
    {
      items: [{ price: PRO_MONTHLY_PRICE, quantity: 1 }],
      trial: true,
      end_date: Math.floor(Date.now() / 1000) + 30 * 24 * 60 * 60,
    },
    {
      items: [{ price: FREE_TIER_PRICE, quantity: 1 }],
    },
  ],
});
```

2. Retrieve your Subscription Schedule

```
const schedule = await stripe.subscriptionSchedules.retrieve(NEW_SCHEDULE_ID);
```

3. Log the phases out to the console and verify that the first phase indeed does have a `trial` property set to `true`.

```
console.log(schedule.phases);
```

4. Make an attempt to access the `trial` property in the first phase

```
const isTrial = schedule.phases[0].trial;
```

### Expected behavior

When attempting to access the `trial` property in a Phase, I should not see an error saying that the `trial` property does not exist in my phase.

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

Node v20.12.1

### Library version

stripe v17.7.0

### API version

2025-02-24.acacia

seanzhang-stripe:
Hi @alexiz10 Although you can set [trial](https://docs.stripe.com/api/subscription_schedules/create?lang=curl#create_subscription_schedule-phases-trial) to a phase when creating a subscription schedule, the resulting subscription schedule object doesn't have a `trial` property in their phases. 

[...]

PR #2406, opened 2025-09-21
fix(types): add missing 'trial' property to SubscriptionSchedule.Phase

Ritiky23:
…e (#2267)

### Why?
The `trial` property is missing in the TypeScript type for `Stripe.SubscriptionSchedule.Phase`. This prevents developers from safely accessing `phase.trial` without manually extending the type. Adding it ensures TypeScript reflects the actual API response.

### What?
- Added `trial?: boolean` to the `Phase` interface in the type definitions.
- Verified that all existing tests pass with this update.

### See Also
- GitHub issue: [#2267](https://github.com/stripe/stripe-node/issues/2267)

### DIFF
--- a/types/SubscriptionSchedules.d.ts
+++ b/types/SubscriptionSchedules.d.ts
@@ -378,6 +378,11 @@ declare module 'stripe' {
          * When the trial ends within the phase.
          */
         trial_end: number | null;
+
+        /**
+         * Whether this phase is a trial phase.
+         */
+        trial?: boolean;
       }
 
       namespace Phase {


xavdid-stripe:
Hi @Ritiky23! Thank you for sending this PR. 

Unfortunately, the `trial` property (or lack thereof) is sourced from our OpenAPI spec and needs to be fixed upstream. Even if we merged this, the next autogenerated update would blow it away.

I'll try to track down the request mentioned in the issue and see if anyone has been working on it.

PR #2426, opened 2025-09-26
Add `StripeContext` object

xavdid-stripe:
### Why?

As the `stripe-context` header is evolving, we want to give better tools for creating and managing contexts. This PR adds a new class, `StripeContext`. It can be used anywhere the context option is supplied and gets serialized to a string when making requests. It's also found on the `EventNotification.context` property.

### What?

- add `StripeContext` class
- allow it to be supplied for RequestOptions.context; serialize it into strings
- add tests
- remove code that ignores stripe-context in v1: that's incorrect.

## Changelog

- Add the `StripeContext` class
- ⚠️ Change `EventNotification` (formerly known as `ThinEvent`)'s `context` property from `string` to `StripeContext`
- support `stripe-context` in V1 requests


src/RequestSender.ts
```diff
@@ -718,9 +730,11 @@ export class RequestSender {
           userSuppliedHeaders: options.headers ?? null,
           userSuppliedSettings: options.settings ?? {},
           stripeAccount:
-            apiMode == 'v2' ? null : this._stripe.getApiField('stripeAccount'),
-          stripeContext:
-            apiMode == 'v2' ? this._stripe.getApiField('stripeContext') : null,
+            options.stripeAccount ?? this._stripe.getApiField('stripeAccount'),
```
xavdid-stripe, line 733: we should be sending stripe-context and stripe-account in both v1 and v2

testProjects/mjs/index.js
```diff
@@ -35,6 +35,7 @@ assert(Stripe.errors.StripeInvalidGrantError);
 assert(Stripe.errors.StripeUnknownError);
 
 assert(Stripe.StripeResource);
+assert(Stripe.StripeContext);
```
xavdid-stripe, line 38: this is the only change in this file - the rest is prettier

src/RequestSender.ts
helenye-stripe, line 733: a question that just popped into my mind: are these mutually exclusive or can you set both?

xavdid-stripe: you can set both but the server either ignores one or rejects it. But either way, IMO we shouldn't validate for that on our end

PR #2453, opened 2025-10-02
add missing fetchEvent type for UnknownEventNotification

xavdid-stripe:
### Why?

When writing the improved events example, I found that I missed adding a type to a manual interface. I also realized that we need to always define `fetchRelatedObject` because on unknown events it says the function is always defined. This now matches how the function works in every other SDK.

### What?

- always define `fetchRelatedObject` so that it won't error out if called on an `UnknownEventNotification` with no `related_object`
- add `fetchEvent` to `UnknownEventNotification` interface
- add tests

## Changelog

- Add missing `fetchEvent()` declaration to the `Stripe.Events.UnknownEventNotification` interface
- Tweak `Stripe.Events.fetchRelatedObject` so that it's always defined and returns `null` if there's no `related_object`. This fixes the situation where the `UnknownEventNotification` says that `fetchRelatedObject()` is defined, but calling it throws an error.

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -525,19 +525,21 @@ export function createStripe(
         );
       };
 
-      if (eventNotification.related_object) {
-        eventNotification.fetchRelatedObject = (): Promise<unknown> => {
-          return this._requestSender._rawRequest(
-            'GET',
-            eventNotification.related_object.url,
-            undefined,
-            {
-              stripeContext: eventNotification.context,
-            },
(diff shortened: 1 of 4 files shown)


jar-stripe:
will review in a bit, but the changelog/description should reference UnknownEventNotification (not UnknownThinEvent), right?

xavdid-stripe:
yep, great call. I should go outside

jar-stripe (approved):

PR #2485, opened 2025-10-31
correctly calculate content-length for JSON bodies that contain unicode

xavdid-stripe:
### Why?

We got reports of 400 errors when users called v2 endpoints w/ unicode strings in the body.

The root cause was the API ingress proxy counting bytes more accurately than the Node SDK. We were underreporting and it was giving a confusing error message back as a result. 

### What?

- add test
- calculate the content-length in a unicode-aware way

### See Also

- [RUN_DEVSDK-2018](https://go/j/RUN_DEVSDK-2018)


src/RequestSender.ts
```diff
@@ -722,7 +722,7 @@ export class RequestSender {
             apiMode == 'v2'
               ? 'application/json'
               : 'application/x-www-form-urlencoded',
-          contentLength: requestData.length,
+          contentLength: Buffer.byteLength(requestData, 'utf8'), // if we calculate this wrong, the server treats it as invalid json
```
mbroshi-stripe, line 725: Looks good to me. For my education: Do we need to support non-Node runtimes, and is `Buffer.byteLength` supported in those?

mbroshi-stripe (approved):
Looks good. Had a non-blocking question

src/RequestSender.ts
xavdid-stripe: > Do we need to support non-Node runtimes

great question- yes!

> is `Buffer.byteLength` supported in those?

also yes! It's a standard function that's existed for ages.

image

PR #2498, opened 2025-11-11
Add EventNotificationHandler

xavdid-stripe:
### Why?

We've been designing a streamlined approach to handling incoming events that is easy to get right and hard to get wrong. This PR has the initial implementation of this new system.

> NOTE: final names for classes / methods are still being gaveled, so don't worry about those for now. Weigh in on [this doc](https://docs.google.com/document/d/1hRpQLMLGfOHhZwXrYt5tvsKT-oIya11M4atunvTOih0/edit?tab=t.2m2wcmw7q88e) if you have strong feelings there!

The only other pending item is to add a method to allow handling a webhook without verifying the signature. This is good for testing and for Event Bridge, which doesn't use the signature-based verification. Otherwise, this is ready for review.

### What?

- add `EventHandler` class
- add event handler constructor on `StripeClient`
- add tests
- add types in like 3 placese
- tweak eslint config

### Example usage

```ts
const app = express();
const client = new Stripe(apiKey);
const handler = client.router(
  webhookSecret,
  async (unhandledEvent, client, details) => {
    // something...
  }
);

handler.on('v1.billing.meter.error_report_triggered', async (event) => {
  const meter = await event.fetchRelatedObject();
  console.log(`Billing Meter ${meter.display_name} had a problem`);
});

app.post(
  '/webhook',
  express.raw({type: 'application/json'}),
  async (req, res) => {
    const sig = req.headers['stripe-signature'] ?? '';
    handler.handle(req.body(), sig);
  }
);

app.listen(4242, () => console.log('Running on port 4242'));
```

### See Also

- [DEVSDK-2822](https://go/j/DEVSDK-2822)
- https://github.com/stripe/stripe-node/pull/2370

### DIFF
--- a/src/StripeEventNotificationHandler.ts
+++ b/src/StripeEventNotificationHandler.ts
@@ -67,7 +67,7 @@ const KNOWN_EVENT_TYPES = new Set([
   // event-types: The end of the section generated from our OpenAPI spec
 ]);
 
-export class StripeEventHandler {
+export class StripeEventNotificationHandler {
   private registeredHandlers: Record<string, HandlerCallback> = {};
   private hasHandledEvent = false;
 
@@ -83,7 +83,7 @@ export class StripeEventHandler {
   ) {}
 
   // these types are duplicated in the manual types
-  public on(type: string, handler: HandlerCallback): this {
+  public on(type: string, callback: HandlerCallback): this {
     if (this.hasHandledEvent) {
       throw new Error(
         'Cannot register new handlers after an event has been handled. This is indicative of a bug.'
@@ -94,7 +94,7 @@ export class StripeEventHandler {
       throw new Error(`Handler already registered for event type: ${type}`);
     }
 
-    this.registeredHandlers[type] = handler;
+    this.registeredHandlers[type] = callback;
     return this;
   }
 
@@ -109,6 +109,7 @@ export class StripeEventHandler {
     rawBody: string | Buffer,
     signature: string | Buffer
   ): Promise<void> {
+    // we're not worried about thread safety here because we expect callbacks will be registered synchronously on app startup
     this.hasHandledEvent = true;
     const event = this.client.parseEventNotification(
       rawBody,
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -22,7 +22,7 @@ import {
   pascalToCamelCase,
   validateInteger,
 } from './utils.js';
-import {StripeEventHandler} from './StripeEventNotificationHandler.js';
+import {StripeEventNotificationHandler} from './StripeEventNotificationHandler.js';
 
 const DEFAULT_HOST = 'api.stripe.com';
 const DEFAULT_PORT = '443';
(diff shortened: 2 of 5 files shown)


.eslintrc.js
```diff
@@ -136,7 +136,6 @@ module.exports = {
     'no-restricted-properties': 'error',
     'no-restricted-syntax': 'error',
     'no-return-assign': 'error',
-    'no-return-await': 'error',
```
xavdid-stripe, line 139: this rule is no longer recommended: https://eslint.org/docs/latest/rules/no-return-await

src/Types.d.ts
```diff
@@ -196,6 +197,15 @@ export type StripeObject = {
     data: RequestData,
     options: RequestOptions
   ) => Promise<any>;
+  parseEventNotification: (
```
xavdid-stripe, line 200: this is internal, it's just used for places in the codebase we call this method

types/V2/EventMisc.d.ts
```diff
@@ -26,6 +26,40 @@ declare module 'stripe' {
...
+      isKnownEventType: boolean;
+    }
+
+    type UnhandledEventHandler = (
+      event: V2.Core.EventNotificationBase,
+      client: Stripe,
+      details: UnhandledNotificationDetails
+    ) => Promise<void>;
+
+    class EventRouter {
```
xavdid-stripe: this type is user-facing

types/index.d.ts
```diff
@@ -718,6 +718,18 @@ declare module 'stripe' {
        */
       receivedAt?: number
     ) => Stripe.V2.Core.EventNotification;
+
+    /**
+     * TODO: docs
+     */
+    router: (
```
xavdid-stripe: this is user facing on the client

src/stripe.core.ts
```diff
@@ -543,6 +544,9 @@ export function createStripe(
 
       return eventNotification;
     },
+    router(webhookSecret: string, onUnhandledHandler: any) {
```
xavdid-stripe: these types are internal, only used for our runtime code

src/stripe.core.ts
```diff
@@ -543,6 +544,9 @@ export function createStripe(
...
+      return new StripeEventHandler(this, webhookSecret, onUnhandledHandler);
```
prathmesh-stripe: Same as PHP, should we be returning multiple handlers or just returning the same object so that users can set all events from multiple places on a single event handler.

types/index.d.ts
```diff
@@ -718,6 +718,18 @@ declare module 'stripe' {
...
+      webhookSecret: string,
+      onUnhandledHandler: (
+        event: Stripe.UnhandledEventHandler,
```
prathmesh-stripe: I like Michael's idea about calling this `FallbackEventHandler`. Or `DefaultEventHandler`. I like the former more.

xavdid-stripe:
@prathmesh-stripe this is ready for a final review!

prathmesh-stripe (approved):

src/StripeEventNotificationHandler.ts
```diff
@@ -0,0 +1,140 @@
...
+    private client: StripeObject,
+    private webhookSecret: string,
+    private fallbackCallback: (
+      event: any,
+      client: any,
+      details: any
+    ) => Promise<void>
+  ) {}
+
+  // these types are duplicated in the manual types
```
prathmesh-stripe, line 85: What was the reason for the duplicating the manual types for this function?

xavdid-stripe: iirc because these types were ignored and everything in `types` is what users actually saw. So I wanted to remind myself that any updates had to be made in two places

Issue #2499, opened 2025-11-12
Buffer is undefined


Pita:
### Describe the bug

Hello, 

we are using the stripe npm package within convex, which has a limited node.js enviroment, very similar to Cloudflare workers. After a recent upgrade, we started getting the Error, Buffer is undefined. Searching through the code, I think I managed to narrow it down to this commit: https://github.com/stripe/stripe-node/commit/8f765dbd899ad92f5be37b66d6525f8d025fcd1b

Rather than Buffer it should use a library that is available in all javascript enviroments, not just node.js.

Thank you very much

### To Reproduce

Do any sort of network request with stripe

### Expected behavior

It works

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

22

### Library version

19.3.0

### API version

2025-10-29.clover

xavdid-stripe:
This is released in [v19.3.1](https://github.com/stripe/stripe-node/releases/tag/v19.3.1).

Keep in mind that we only officially support Node and all other environments are best-effort, so use elsewhere with caution!

PR #2500, opened 2025-11-12
Use `TextEncoder` instead of `Buffer` to ensure support in other JS environments

Pita:
### Why?
See #2499 bug description. Buffer is only available in node.js enviroments

### What?
Changed it to TextEncoder which is widely available in all Javascript enviroments and does the same thing


xavdid-stripe (approved):
Awesome, thanks for fixing! And sorry about the hassle. 

Once you sign the CLA, we can merge!

PR #2508, opened 2025-11-18
Update CHANGELOG.md to remove changes on Events

ramya-stripe:
### Why?
The adding of `events` to the Events API was reverted.

### What?
Update changelog.md


artistfashow-crypto (commented):
Markdown all subs and repos as correct I finite nodes response for these yet hazardous and heavens now.

Issue #2538, opened 2026-01-06
File upload via files.create API always timing out with NodeJS


thiagoarruda:
### Describe the bug

Hey folks, after upgrading `stripe-node` to `20.1.0` (from `14.5.0`), our file uploads stopped working, always returning a timeout error:

```shell
StripeConnectionError: Request aborted due to timeout being reached (30000ms)
    at ./node_modules/stripe/cjs/RequestSender.js:405:41
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  type: 'StripeConnectionError',
  raw: {
    message: 'Request aborted due to timeout being reached (30000ms)',
    detail: TypeError: ETIMEDOUT
        at HttpClient.makeTimeoutError (./node_modules/stripe/cjs/net/HttpClient.js:23:28)
        at ClientRequest.<anonymous> (./node_modules/stripe/cjs/net/NodeHttpClient.js:47:56)
        at Object.onceWrapper (node:events:621:28)
        at ClientRequest.emit (node:events:507:28)
        at TLSSocket.emitRequestTimeout (node:_http_client:863:9)
        at Object.onceWrapper (node:events:621:28)
        at TLSSocket.emit (node:events:519:35)
        at Socket._onTimeout (node:net:609:8)
        at listOnTimeout (node:internal/timers:614:17)
        at process.processTimers (node:internal/timers:549:7) {
      code: 'ETIMEDOUT'
    }
  },
```

I've tested a few different versions of the client, and the latest version where the file upload worked as expected is `18.5.0`, and the issue starts happening on `19.0.0`. Also, the issue only happens when using NodeJS. If I run the same snippet with `stripe-node` version `20.1.0` with Bun, it works as expected.

The request does not appear on the logs on Stripe's dashboard.

### To Reproduce

1. Call the `files.create` API using `stripe-node` version `20.1.0` with NodeJS 20, 22 or 24.

### Expected behavior

The file is uploaded successfully, as in version `18.5.0`:

```js
{
  id: 'file_1Smg38EyZ8sZIiBpclanZxFu',
  object: 'file',
  created: 1767728174,
  expires_at: null,
  filename: 'document.jpg',
  links: {
    object: 'list',
    data: [],
    has_more: false,
    url: '/v1/file_links?file=file_1Smg38EyZ8sZIiBpclanZxFu'
  },
  purpose: 'identity_document',
  size: 240,
  title: null,
  type: 'png',
  url: null
}
```

### Code snippets

[...]

PR #2542, opened 2026-01-07
Fix content length calculations for multipart file uploads

jar-stripe:
### Why?
https://github.com/stripe/stripe-node/pull/2500 changed `prepareAndMakeRequest` to use TextEncoder for broader compatibility.  Unfortunately, /v1/files APIs use a multipart file encoder that passes a `Uint8Array` into `prepareAndMakeRequest`; TextEncoder.encode does not support `Uint8Array` and so the change caused contentLength to be computed incorrect, and file uploads to hang/time out (see https://github.com/stripe/stripe-node/issues/2538).

### What?
- refactors out content length calculation to _getContentLength to encapsulate critical logic and for easy testing
- adds type check so we only use `TextEncoder` for string data
- updates `requestData` type and other related types in `_request` to be `string | Uint8Array`
- adds comprehensive tests for `_getContentLength`

### DIFF
--- a/test/RequestSender.spec.ts
+++ b/test/RequestSender.spec.ts
@@ -340,7 +340,9 @@ describe('RequestSender', () => {
         const encoder = new TextEncoder();
         const bytes = encoder.encode(str);
 
-        expect(sender._getContentLength(str)).to.equal(sender._getContentLength(bytes));
+        expect(sender._getContentLength(str)).to.equal(
+          sender._getContentLength(bytes)
+        );
         expect(sender._getContentLength(str)).to.equal(11);
       });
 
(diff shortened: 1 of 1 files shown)


test/RequestSender.spec.ts
```diff
@@ -162,6 +162,224 @@ describe('RequestSender', () => {
...
+
+      it('should handle multiple emoji characters', () => {
+        // Each emoji is typically 4 bytes
+        const result = sender._getContentLength('🎉🎊🎈');
+        expect(result).to.equal(12); // 3 emojis * 4 bytes each
+      });
+
+      it('should calculate correct byte length for Chinese characters', () => {
+        // Chinese characters are typically 3 bytes in UTF-8
+        const result = sender._getContentLength('你好');
```
jar-stripe, line 214: google translate says this says "hello"

test/RequestSender.spec.ts
```diff
@@ -162,6 +162,224 @@ describe('RequestSender', () => {
...
+        expect(result).to.equal(6); // 2 characters * 3 bytes each
+      });
+
+      it('should calculate correct byte length for Arabic characters', () => {
+        // Arabic characters are typically 2 bytes in UTF-8
+        const result = sender._getContentLength('مرحبا');
```
jar-stripe, line 220: google translate says this says "welcome"

PR #2546, opened 2026-01-12
Setup OIDC workflow for setting up trusted publishing on npm registry

prathmesh-stripe:
### Why?
On Dec 9, NPM invalidated all "Classic tokens" in favor of "Granular tokens". These expire every few months and in order to establish a more secure workflow for publishing, we are moving to a method called [trusted publishing](https://docs.npmjs.com/trusted-publishers).

### What?
* Removed NPM Auth Token required for publish
* Removed OTP Auth scaffolding used as 2FA required for publish

### See Also
https://jira.corp.stripe.com/browse/RUN_DEVSDK-2104

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,8 @@ jobs:
   build:
     name: 'Static Checks'
     runs-on: 'ubuntu-24.04'
+    permissions:
+      contents: read
 
     steps:
       - uses: extractions/setup-just@v2
@@ -70,6 +72,8 @@ jobs:
           - '18'
           - '16'
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     steps:
       - uses: extractions/setup-just@v2
       - uses: actions/checkout@v2
@@ -117,6 +121,8 @@ jobs:
       endsWith(github.actor, '-stripe')
     needs: [build, test]
     runs-on: 'ubuntu-24.04'
+    permissions:
+      contents: read
     steps:
       # just is called in `yarn prepack`, which is called during the `publish` operation
       - uses: extractions/setup-just@v2
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -10,6 +10,8 @@ on:
       - opened
       - reopened
 
+permissions: {}
+
 jobs:
   check_auto_merge_method_if_enabled:
     name: Check auto-merge method if enabled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@
   * Add support for event notifications `V2PaymentsSettlementAllocationIntentSplitCanceledEvent`, `V2PaymentsSettlementAllocationIntentSplitCreatedEvent`, and `V2PaymentsSettlementAllocationIntentSplitSettledEvent` with related object `V2.Payments.SettlementAllocationIntentSplit`
   * Remove support for error code `account_rate_limit_exceeded` on `RateLimitError`
 
+## 20.1.2 - 2026-01-07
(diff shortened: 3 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -30,6 +33,9 @@ jobs:
 
       - name: Setup node
         uses: actions/setup-node@v2
+        with:
+          node-version: '22'
+          registry-url: 'https://registry.npmjs.org'
```
jar-stripe: why is the registry url needed?  do we default to a different url?

.github/workflows/main.yml
```diff
@@ -72,6 +78,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
+          registry-url: 'https://registry.npmjs.org'
```
jar-stripe: same as above

.github/workflows/main.yml
```diff
@@ -116,11 +123,13 @@ jobs:
       # just is called in `yarn prepack`, which is called during the `publish` operation
       - uses: extractions/setup-just@v2
       - uses: actions/checkout@v2
-      - run: sudo apt-get install -y oathtool
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '22'
+          registry-url: 'https://registry.npmjs.org'
```
jar-stripe, line 133: same as above

.github/workflows/main.yml
```diff
@@ -129,10 +138,7 @@ jobs:
           echo "Publishing $VERSION with $NPM_TAG tag."
           # stop debug logging now, so we don't log the OTP
           set +x
-          npm publish --otp="$(oathtool -b --totp $NPM_OTP)" --tag $NPM_TAG
-        env:
-          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
-          NPM_OTP: ${{ secrets.NPM_OTP }}
+          npm publish --tag $NPM_TAG
```
jar-stripe, line 145: is this really all we need?  how does the OIDC token get injected in here?

.github/workflows/main.yml
```diff
@@ -129,10 +138,7 @@ jobs:
           echo "Publishing $VERSION with $NPM_TAG tag."
           # stop debug logging now, so we don't log the OTP
           set +x
```
jar-stripe, line 144: is this still needed now that we're not using the OTP?

.github/workflows/main.yml
prathmesh-stripe: I was following the instructions from [npm docs](https://docs.npmjs.com/trusted-publishers#github-actions-configuration).  Although this is still the same default registry.

.github/workflows/main.yml
prathmesh-stripe: The output of npm publish is a LOT of verbose text relating to the composition of the package. This command should be fine although I'll tweak the comment.

.github/workflows/main.yml
prathmesh-stripe: Yes, there was some setup that was required on npmjs.org and we needed `id-token: write` permission which is set at the top.

.github/workflows/main.yml
prathmesh-stripe: Although this would be default when fetched from npm. Are there any issues with having this?

.github/workflows/main.yml
jar-stripe: that makes sense; and definitely cool to have in if its part of their official docs.  i just was curious why/if it was needed

.github/workflows/main.yml
jar-stripe: nah non particularly, esp if its in their docs

jar-stripe (approved):

PR #2553, opened 2026-01-14
Use Node v24 and latest NPM version

prathmesh-stripe:
### Why?
Trusted publishing requires Node 24 and NPM (11.5.1+).

### What?
- Update Node to 24.
- Install the latest version of NPM on major version 11.

### See Also
https://github.com/npm/cli/issues/8730#issuecomment-3657780285

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -141,7 +141,7 @@ jobs:
           echo "Publishing $VERSION with $NPM_TAG tag."
           # stop debug logging to avoid dumping the output of `npm publish` command.
           set +x
-          npm publish --tag $NPM_TAG 
+          npm publish --tag $NPM_TAG
       - uses: stripe/openapi/actions/notify-release@master
         if: always()
         with:


.github/workflows/main.yml
```diff
@@ -127,10 +127,11 @@ jobs:
       # just is called in `yarn prepack`, which is called during the `publish` operation
       - uses: extractions/setup-just@v2
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v4
         with:
-          node-version: '22'
+          node-version: '24'
           registry-url: 'https://registry.npmjs.org'
+      - run: npm install -g npm@11
```
jar-stripe, line 134: not gonna block, but shouldn't this come for free with a modern node version?  or do we need a version of npm even beyond what's bundled in the latest node LTS for this all to work?

jar-stripe (approved):

.github/workflows/main.yml
prathmesh-stripe: We need 11.5.1+. That is explicitly mentioned in the docs and a lot of people stumbled on this.

jar-stripe: 🙃

PR #2554, opened 2026-01-14
Updated permissions and removed debug logging blocker

prathmesh-stripe:
### Why?
Removed debug logger to understand why npm publish isn't working.

### What?
- Removed debug logging.
- Moved the `id-token` permission to the job
- Updated package.json to ensure the repository url is the same in npmjs and github. (Guessing the git:// isn't what npmjs expects).

### See Also
https://github.com/npm/cli/issues/8730

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git://github.com/stripe/stripe-node.git"
+    "url": "https://github.com/stripe/stripe-node"
   },
   "bugs": "https://github.com/stripe/stripe-node/issues",
   "engines": {


package.json
```diff
@@ -18,7 +18,7 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git://github.com/stripe/stripe-node.git"
+    "url": "https://github.com/stripe/stripe-node"
```
jar-stripe, line 21: was this an intended change?

prathmesh-stripe: Yes

jar-stripe: not urgent, but im curious; why http and not git?  is that part of switching to this new publishing model, or a best practice?

prathmesh-stripe: https://github.com/npm/cli/issues/8730#issuecomment-3708284079 -> Followed this user's PR and saw the package.json of that repository.

PR #2151, opened 2024-08-04
removed qs package dependency

yashraj02:
1. Removed qs dependency.
2. Request reference #2117

Issue #2117, opened 2024-06-24
consider replacing `qs` dependency


EddyVinck:
### Is your feature request related to a problem? Please describe.

The `qs` library includes a lot of polyfills via subdependencies. Polyfills that are required for very old versions of Node.js below the Node.js version that Stripe intends to support (v12 according to the README). These polyfills are likely slowing down the Stripe library unnecessarily.

### Describe the solution you'd like

The package is replaced.

### Describe alternatives you've considered

-

### Additional context

The maintainer of `qs` is also acting without the interest of his users in mind for a package he just became a maintainer of: https://x.com/Jordy_vD_/status/1804906230940569952

EddyVinck:
I see there even is a PR for it already https://github.com/stripe/stripe-node/pull/2116

artemtam:
Hi @helenye-stripe! 
I see there are 2 PRs open, any blockers from landing either of them? 

`qs` brings ~20 dependencies, including some completely useless polyfills. It can be replaced by native URLSearchParams or at least with [neoqs](https://github.com/PuruVJ/neoqs) (zero-dependency drop-in replacement). Lmk if you need any help!

<img width="1370" height="550" alt="Image" src="https://github.com/user-attachments/assets/6e469afa-9887-47ca-8b65-c4e3f0460f6c" />

KhafraDev:
Not only is it bringing in tons of dependencies, it brings in around 94 KB of garbage when bundling with esbuild[.](https://esbuild.github.io/analyze/) If qs was removed, a bundle for one of my apps on Cloudflare Workers would decrease by 12.6%.

<img width="540" height="744" alt="Image" src="https://github.com/user-attachments/assets/cb856593-22dc-42e0-a2bd-7f4c89d90119" />

pozylon:
I back this, especially in a world of supply-chain-attacks every package one can easily get rid of should be removed immediately. this is one of those dependencies. also stripe should not support node.js versions older than v20 for newer releases, why not let users just install older versions if they need v12 support?

mbroshi-stripe:
Thanks all for your patience on this one! We just merged https://github.com/stripe/stripe-node/pull/2558, which removes the dependency on `qs`. It will be released with our next scheduled monthly release at the end of this month.

ramya-stripe:
Yesterday's release (v20.3.0) no longer has the `qs` dependency. Thanks for your patience everyone!

PR #2558, opened 2026-01-15
Remove `qs` dependency

mbroshi-stripe:
### Why?

This addresses https://github.com/stripe/stripe-node/issues/2117. There are actually 4 (!) open PRs for this (https://github.com/stripe/stripe-node/pull/2168, https://github.com/stripe/stripe-node/pull/2116, https://github.com/stripe/stripe-node/pull/2146, and https://github.com/stripe/stripe-node/pull/2151). We decided it would be most efficient at this point to just implement our own PR to address the issue. 

### What?

- Removes the `qs` dependency, instead writing our own `stringify` function
- Also deprecates the unneeded `apiMode` parameters since we now stringify the same in both V1 and V2

All tests pass on both `master` and this feature branch, which ensures we have no regressions. I also tested some API calls manually in both Node and Bun runtimes.

## Changelog
- Remove the `qs` dependency, addressing [Issue #2117](https://github.com/stripe/stripe-node/issues/2117).


src/utils.ts
```diff
@@ -52,23 +51,91 @@ export function isOptionsHash(o: unknown): boolean | unknown {
...
-      // Don't use strict form encoding by changing the square bracket control
-      // characters back to their literals. This is fine by the server, and
-      // makes these parameter strings easier to read.
+    encodeURIComponent(value)
+      // Encode characters not encoded by encodeURIComponent but encoded by qs
+      .replace(/!/g, '%21')
+      .replace(/\*/g, '%2A')
+      .replace(/\(/g, '%28')
+      .replace(/\)/g, '%29')
+      .replace(/'/g, '%27')
```
mbroshi-stripe, line 72: Not 100% sure if we want to keep backwards compatibility here with `qs`, but that seems safest.

yarn.lock
```diff
@@ -461,11 +461,6 @@
   dependencies:
     undici-types "~6.21.0"
 
-"@types/qs@^6.9.7":
```
prathmesh-stripe, line 464: ❤️ for removing so many dependencies!

src/utils.ts
prathmesh-stripe: I think it is fine to keep this!

test/utils.spec.ts
```diff
@@ -123,6 +123,561 @@ describe('utils', () => {
...
+        utils.queryStringifyRequestData({
+          value: NaN,
+        })
+      ).to.equal('value=NaN');
+    });
+
+    it('Handles Infinity by converting to string', () => {
+      expect(
+        utils.queryStringifyRequestData({
+          value: Infinity,
```
prathmesh-stripe, line 704: Is this supported by Stripe API?

test/utils.spec.ts
```diff
@@ -123,6 +123,561 @@ describe('utils', () => {
...
+        utils.queryStringifyRequestData({
+          date: new Date(1234567890123),
+        })
+      ).to.equal('date=1234567890');
+    });
+
+    it('Handles NaN by converting to string', () => {
+      expect(
+        utils.queryStringifyRequestData({
+          value: NaN,
```
prathmesh-stripe, line 684: Is this a supported use case?

test/utils.spec.ts
```diff
@@ -123,6 +123,561 @@ describe('utils', () => {
...
+
+    it('Handles nested null values', () => {
+      expect(
+        utils.queryStringifyRequestData({
+          a: {
+            b: null,
+            c: 'value',
+          },
+        })
+      ).to.equal('a[b]=&a[c]=value');
```
prathmesh-stripe, line 348: I'm questioning if `a[b]` is captured as empty string or null value on the API side.

mbroshi-stripe: It's handled as an empty string. This is important functionality for emptyable fields, so we want to keep this behavior.

test/utils.spec.ts
mbroshi-stripe: If a user inputs
```js
stripeClient.paymentIntents.create({
  amount: NaN,
  currency: 'usd',
});
```
We need to do _something_. I don't think the SDK should be opinionated about whether it's a permissible value. I think we should serialize it and let the Stripe API decide. So, serializing to `NaN` seems reasonable to me, and is the current behavior with `qs`

test/utils.spec.ts
mbroshi-stripe: Ditto as above

PR #2562, opened 2026-01-19
Document usage of undocumented API parameters

ramya-stripe:
Added section on using undocumented parameters and properties in API requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,7 +418,7 @@ expect(event.id).to.equal(payload.id);
 
 In some cases, you might encounter parameters on an API request or fields on an API response that aren’t available in the SDKs.
 This might happen when they’re undocumented or when they’re in preview and you aren’t using a preview SDK. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -414,6 +414,11 @@ const event = stripe.webhooks.constructEvent(payloadString, header, secret);
 // Do something with mocked signed event
 expect(event.id).to.equal(payload.id);
 ```
+### How to use undocumented parameters and properties
+
+In some cases, you might encounter parameters on an API request or fields on an API response that aren’t available in the SDKs.
+This might happen when they’re undocumented or when they’re in preview and you aren’t using a preview SDK. 
+See [undocumented params and properties](https://docs.stripe.com/sdks/server-side#undocumented-params-and-fields) to send those parameters or access those fields.
```
ramya-stripe:
```suggestion
See [undocumented params and properties](https://docs.stripe.com/sdks/server-side?lang=node#undocumented-params-and-fields) to send those parameters or access those fields.
```

PR #2569, opened 2026-01-29
Inital PoC

mbroshi-stripe:

### DIFF
--- a/test/resources/SerializeBatch.spec.js
+++ b/test/resources/SerializeBatch.spec.js
@@ -113,7 +113,7 @@ describe('serializeBatch methods', () => {
   });
 
   describe('subscriptionSchedules.serializeBatchCreate', () => {
-    it('produces correct structure with null path_params', () => {
+    it('produces correct structure without path_params', () => {
       const result = JSON.parse(
         stripe.subscriptionSchedules.serializeBatchCreate({
           customer: 'cus_123',
@@ -121,7 +121,7 @@ describe('serializeBatch methods', () => {
       );
 
       expect(result.id).to.match(UUID_REGEX);
-      expect(result.path_params).to.be.null;
+      expect(result).to.not.have.property('path_params');
       expect(result.params).to.deep.equal({customer: 'cus_123'});
       expect(result.stripe_version).to.be.a('string').and.not.be.empty;
     });


src/resources/Accounts.ts
```diff
@@ -84,6 +85,26 @@ export const Accounts = StripeResource.extend({
     method: 'GET',
     fullPath: '/v1/accounts/{account}/persons/{person}',
   }),
+  serializeBatchUpdate(
+    account: string,
+    params: Record<string, unknown> = {},
```
xavdid-stripe, line 90: `Record<string, unknown>` is fine, but we _do_ know this type right? it's a `AccountUpdateParams`.

You should probably touch base with @prathmesh-stripe since where/how you define types like this is about to change significantly (in the next few hours)

prathmesh-stripe: Is there a codegen PR for this?

mbroshi-stripe: Thanks for the head up. Yeah, I asked Claude about this one during implementation, and it assured me that params don't have types. I didn't investigate too fully, though.

prathmesh-stripe: https://github.com/stripe/stripe-node/blob/daedb44bad4fe528e6fc1dd6f8482099124f3b27/types/AccountsResource.d.ts#L1940

This will be dead code in a few hours but it does already exist

mbroshi-stripe: Should you merge your change in first then? Or I'll rebase onto your branch?

prathmesh-stripe: I think that will be ideal. You can look at these PRs till then:
https://github.com/stripe/stripe-node/pull/2619
https://github.com/stripe/sdk-codegen/pull/3174

mbroshi-stripe: This is what Claude says:

---

Here's the situation:

**The `.d.ts` type definition already uses the correct type** — `params?: AccountUpdateParams`. This is generated by `generateSerializeMethodSignaturesSyntax` which calls `getSerializeTargetParamsTypeName` to derive the proper type name.

**The `.ts` implementation file uses `Record<string, unknown>`** — This is generated by `generateSerializeMethodSyntax` which hardcodes `'params: Record<string, unknown> = {}'`.

The fix is straightforward: in `generateSerializeMethodSyntax` (`src/node/generator.tsx`), the `params` parameter should use the concrete params type name instead of `Record<string, unknown>`. Since this is a `.ts` file (not `.js`), it can reference the types.

However, there's a subtlety — the implementation file doesn't import the params types. The `.d.ts` declares them, but the `.ts` resource file uses `StripeResource.extend({...})` which doesn't have type imports for params. Looking at the existing pattern, the other methods in the `.ts` file use `...args: any[]` and don't reference params types at all.

---

It sounds like in general in the `.ts` files, we don't have types. I was thinking of just leaving this be, but I am no TS expert, so happy to do whatever you guys say.

prathmesh-stripe: If you merge this, I'll fix it when I ship the actual full TS support.

Issue #2327, opened 2025-05-03
TypeScript types are not accurate when expanding data and it's very annoying.


chevcast:
### Describe the bug

I'm creating a new issue under the bug category because I don't think #1556 really captured how irritating this is. It makes using this otherwise fantastic package very tedious.

When you pass the `expand` property to expand API response data, TypeScript typings are unnaffected, leaving it up to the user to augment the stripe types manually.

### To Reproduce

Let's say I want to retrieve the stripe checkout session to simply display a checkout success page. In plain JavaScript it's as simple as this:

```js
const session = (await stripe.checkout.sessions.retrieve(checkoutSessionId, {
    expand: ["subscription", "subscription.items.data.product"]
}));
```

Now our session will have the full subscription object and our subscription items will include the associated price AND the product associated with that price.

However, in TypeScript the type of `session.subscription` is `Stripe.Checkout.Session.subscription: string | Stripe.Subscription | null`. Because `string` is still a possibility I cannot use `session.subscription` directly. Instead I have to assert the type at every point.

```ts
const subscription = session.subscription as Stripe.Subscription;
```

Doesn't seem too bad for a top-level property, but it quickly gets out of hand:

```ts
const productName = ((session.subscription as Stripe.Subscription).items.data[0].price.product as Stripe.Product).name;
```

Capturing and asserting the type at every level gets extremely irritating and unreadable so you then end up having to dig deep into the Stripe type tree and create your own expanded types:

```ts
type ExpandedSession = Stripe.Checkout.Session & {
    subscription: Stripe.Subscription & {
        items: {
            data: Stripe.SubscriptionItem & {
                price: Stripe.Price & {
                    product: Stripe.Product;
                };
            };
        };
    };
};
const session = (await stripe.checkout.sessions.retrieve(checkoutSessionId, {
    expand: ["subscription", "subscription.items.data.product"]
})) as Stripe.Response<ExpandedSession>;
```

I've never used a package that made managing types this tedious. It's so annoying and so unlike any other well-typed packages I've ever used that I think it deserves to be documented as a bug. Feel free to disagree but I have a hard time believing I'm the only one that finds this such a chore.

### Expected behavior

[...]

PR #2577, opened 2026-02-18
feat: Add getId helper method to extract item IDs from Stripe objects

### LINKED ISSUE #1793: Helper method to get item ids

I'm often writing the same TypeScript code to narrow the types returned by the API to extract item IDs. Example:

```ts
const price = await stripe.prices.retrieve(priceId);
const { product } = price;

// this part
const productId = (typeof product === 'string' ? product : product.id);
```

It would be nice if this library included a helper method for this, e.g.

```ts
function getId(stripeObject: { id: string } | string) {
  return typeof stripeObject === 'string' ? stripeObject : stripeObject.id;
}
```

Then I could write

```ts
const price = await stripe.prices.retrieve(priceId);
const { product } = price;

const productId = getId(product);
```

I could just write this function for use in my personal projects, but if this is a common problem/use-case I think including it in the library would be helpful to others too

### Additional context

_No response_

imanali2237:
**# Add getId() Helper Method
Fixes #1793**

Implements a helper method to simplify extracting IDs from Stripe objects that can be either:
- Expanded objects with an 'id' property
- Simple string IDs

This eliminates the need for the repetitive pattern:

```
const id = (typeof obj === 'string' ? obj : obj.id)
Now users can simply write:
const id = stripe.getId(obj)
hy?
```
Developers frequently need to extract IDs from Stripe API responses. The Stripe API returns expanded objects (e.g., [{ id: 'prod_123', name: 'Product' }] In some cases and simple string IDs in others depending on whether the object was expanded. This forces users to write repetitive defensive code to handle both cases.

By providing a built-in helper method, we:

- Reduce boilerplate code for users
- Improve type safety with proper TypeScript support
- Provide a standard, discoverable API for this common operation
- Improve code readability and maintainability

**What?**

- ✨ Added [getId(stripeObject: {id: string} | string): string]helper function in [utils.ts]
- ✅ Added comprehensive unit tests covering:
- String ID inputs
- Object ID inputs
- Various object types with id properties
- Edge cases (empty strings, etc)
- 📦 Exported the helper through the Stripe instance (available as [stripe.getId()
- 📚 Added complete TypeScript type definitions and JSDoc documentation
- 🔗 Integrated with existing utility exports

**Testing**
All tests pass successfully:
npm test -- --grep "getId"
✓ should return the string when given a string
✓ should return the id property when given an object with an id property
✓ should work with various object types that have an id property
✓ should handle empty strings
✓ should handle objects with empty string id

5 passing

**Before** 
```
const price = await stripe.prices.retrieve(priceId);
const { product } = price;

// Had to write this pattern repeatedly
const productId = (typeof product === 'string' ? product : product.id);
```
**After**:

```
const price = await stripe.prices.retrieve(priceId);
const { product } = price;

// Simply use the helper
const productId = stripe.getId(product);
```

### DIFF
--- a/src/Types.d.ts
+++ b/src/Types.d.ts
@@ -143,6 +143,7 @@ export type StripeObject = {
   StripeResource: StripeResourceConstructor;
   errors: any;
   webhooks: any;
+  getId: (stripeObject: {id: string} | string) => string;
   _prepResources: () => void;
   _setAppInfo: (appInfo: AppInfo) => void;
   _prevRequestMetrics: Array<{
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -19,6 +19,7 @@ import * as resources from './resources.js';
 import {
   createApiKeyAuthenticator,
   determineProcessUserAgentProperties,
+  getId,
   pascalToCamelCase,
   validateInteger,
 } from './utils.js';
@@ -542,6 +543,10 @@ export function createStripe(
       };
 
       return eventNotification;
+    },
+
+    getId(stripeObject: {id: string} | string): string {
+      return getId(stripeObject);
     },
   } as StripeObject;
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -557,3 +557,13 @@ export function parseHeadersForFetch(
     return [key, parseHttpHeaderAsString(value)];
   });
 }
+
+/**
+ * Extracts the ID from a Stripe object or string.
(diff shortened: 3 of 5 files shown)


prathmesh-stripe:
Thanks for the PR. We updated our readme to include this helper method for our users in https://github.com/stripe/stripe-node/pull/2579. We have layers of nested expandable classes and we would be looking to solve that with a more generic solution. You can read more about this on https://github.com/stripe/stripe-node/issues/2327

PR #2590, opened 2026-03-05
Add Stripe-Request-Trigger header

xavdid-stripe:
### Why?

We're working on surfacing more information to users around what their integration does when handling events. The first step of this is to tie subsequent API requests to the act of event processing. This PR adds a header to do just that.

There's some messiness here between `headers` and `additionalHeaders`, but it's not a thing we need to fix this second. I also fixed a bug with `additionalHeaders` that would have overridden auth headers (which makes me think no one has been using it)

### What?

- add `prepare` recipe to justfile to aid automation
- add `Stripe-Request-Trigger` header in event helpers
- fix up a bug with rawRequest's `additionalHeaders`
- update node version (so it matches codegen; we didn't do it last oct)

### See Also

- [DEVSDK-3012](https://go/j/DEVSDK-3012)

### DIFF
--- a/src/resources/V2/Core/Events.ts
+++ b/src/resources/V2/Core/Events.ts
@@ -57,6 +57,9 @@ export const Events = StripeResource.extend({
         }).apply(this, [
           {
             stripeContext: pulledEvent.context,
+            headers: {
+              'Stripe-Request-Trigger': `event=${pulledEvent.id}`,
+            },
           },
         ]),
     };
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -65,6 +65,7 @@ jobs:
         node:
           # should include even numbers >= 16
           # see: https://docs.stripe.com/sdks/versioning?lang=node#stripe-sdk-language-version-support-policy
+          - '24'
```
jar-stripe, line 68: dont forget to merge into the other branches after this lands 😄

jar-stripe (approved):

PR #2617, opened 2026-03-20
Add decimal_string support with vendored Decimal type

jar-stripe:
### Why?
The Stripe API uses `decimal_string` fields — high-precision decimal values that travel as JSON strings on the wire to avoid floating-point loss. This PR adds a `Stripe.Decimal` type and bidirectional coercion so these fields are transparent to SDK users: pass a `Stripe.Decimal` in, get a `Stripe.Decimal` back.

Coercion fires for both v1 and v2 methods. The codegen-emitted `requestSchema`/`responseSchema` metadata drives which fields get coerced; methods without decimal fields are unaffected. The `apiMode === "v2"` guard in `StripeResource.ts` has been removed — coercion now fires whenever a schema is present, matching Tier 1 SDK behavior (Java, Go, .NET).

### What?
**Runtime (hand-written):**
- **`src/Decimal.ts`**: Vendored from the Stripe-authored `apps-extensibility-sdk`. Uses `BigInt` coefficient × 10^exponent for arbitrary precision with zero external dependencies. Branded TypeScript type (`__decimalBrand`). Frozen instances. Full arithmetic (+, -, ×, ÷), comparison, and formatting methods.
- **`src/Types.d.ts`**: Adds `{kind: 'decimal_string'}` to `V2RuntimeSchema` discriminated union.
- **`src/V2Coercion.ts`**: Adds `decimal_string` coercion in `coerceV2ResponseData` (`Decimal.from(value)`) and `coerceV2RequestData` (duck-type detect Decimal → `.toString()`). Decimal plugs into the existing int64 coercion pipeline.
- **`src/StripeResource.ts`**: Removes `apiMode === "v2"` guard — coercion fires whenever `spec.requestSchema`/`spec.responseSchema` is present.
- **`src/stripe.cjs.node.ts`** / **`src/stripe.cjs.worker.ts`** / **`src/stripe.esm.node.ts`** / **`src/stripe.esm.worker.ts`**: Export `Decimal` for both CJS and ESM consumers.
- **`types/lib.d.ts`**: Manual `Stripe.Decimal` interface declaration (instance methods + static `from()`/`zero`), `RoundDirection`, `DecimalRoundingOptions`, `DecimalRoundingPresets`, `isDecimal()`, `DEFAULT_DIV_PRECISION`.
- **`tsconfig.cjs.json`**: Bumped `target` from `es2017` to `es2020` for BigInt literal support.

**Generated:**
- All methods with `decimal_string` fields in request params or response (v1 and v2) now emit `requestSchema`/`responseSchema` entries with `{kind: 'decimal_string'}`.
- `primitiveDecimalType` changed from `string` to `Decimal`, so TypeScript types use `Stripe.Decimal` instead of `string` for these fields.

### See Also
- int64_string pipeline this builds on: https://github.com/stripe/stripe-node/pull/2600

## Changelog
- :warning: **Breaking change:** All `decimal_string` fields changed type from `string` to `Stripe.Decimal` in both request params and response objects. Code that reads or writes these fields as `string` will need to use `Stripe.Decimal` instead (construct via `Decimal.from("1.23")`, serialize via `.toString()`). Affected fields across v1 and v2 APIs:
  - **Checkout.Session**: `currency_conversion.fx_rate`
  - **Climate.Order**: `metric_tons`; **Climate.Product**: `metric_tons_available`
  - **CreditNoteLineItem**: `unit_amount_decimal`
  - **InvoiceItem**: `quantity_decimal`, `unit_amount_decimal`
  - **InvoiceLineItem**: `quantity_decimal`, `unit_amount_decimal`
  - **Issuing.Authorization** / **Issuing.Transaction** (and TestHelpers): `quantity_decimal`, `unit_cost_decimal`, `gross_amount_decimal`, `local_amount_decimal`, `national_amount_decimal`
  - **Plan**: `amount_decimal`, `flat_amount_decimal`, `unit_amount_decimal`
  - **Price**: `unit_amount_decimal`, `flat_amount_decimal` (including `currency_options` and `tiers`)
  - **V2.Core.Account** / **V2.Core.AccountPerson**: `percent_ownership`
  - Request params on **Invoice**, **Product**, **Quote**, **Subscription**, **SubscriptionItem**, **SubscriptionSchedule**, **PaymentLink**: `unit_amount_decimal`, `flat_amount_decimal`, `quantity_decimal` (where applicable)

### DIFF
--- a/test/Decimal.spec.ts
+++ b/test/Decimal.spec.ts
@@ -330,21 +330,21 @@ describe('Decimal', () => {
     });
 
     it('throws on division by zero', () => {
-      expect(() => Decimal.from('42').div(Decimal.zero, 34, 'half-up')).to.throw(
-        'Division by zero'
-      );
+      expect(() =>
+        Decimal.from('42').div(Decimal.zero, 34, 'half-up')
+      ).to.throw('Division by zero');
     });
 
     it('throws on negative precision', () => {
-      expect(() => Decimal.from('10').div(Decimal.from('3'), -5, 'half-up')).to.throw(
-        'precision must be a non-negative integer'
-      );
+      expect(() =>
+        Decimal.from('10').div(Decimal.from('3'), -5, 'half-up')
+      ).to.throw('precision must be a non-negative integer');
     });
 
     it('throws on non-integer precision', () => {
-      expect(() => Decimal.from('10').div(Decimal.from('3'), 2.5, 'half-up')).to.throw(
-        'precision must be a non-negative integer'
-      );
+      expect(() =>
+        Decimal.from('10').div(Decimal.from('3'), 2.5, 'half-up')
+      ).to.throw('precision must be a non-negative integer');
     });
 
     it('half-up rounding', () => {
@@ -649,12 +649,8 @@ describe('Decimal', () => {
     });
 
     it('rounds with half-even', () => {
-      expect(Decimal.from('1.235').toFixed(2, 'half-even')).to.equal(
-        '1.24'
-      );
-      expect(Decimal.from('1.225').toFixed(2, 'half-even')).to.equal(
-        '1.22'
-      );
+      expect(Decimal.from('1.235').toFixed(2, 'half-even')).to.equal('1.24');
+      expect(Decimal.from('1.225').toFixed(2, 'half-even')).to.equal('1.22');
     });
 
     it('throws on negative decimal places', () => {


types/apiVersion.d.ts
```diff
@@ -1,4 +0,0 @@
-// File generated from our OpenAPI spec
```
xavdid-stripe, line 1: I think it'll get re-added so it maybe doesn't matter, but should this have been removed?

tsconfig.cjs.json
```diff
@@ -3,7 +3,7 @@
     "outDir": "./cjs",
     "module": "commonjs",
     "moduleResolution": "node",
-    "target": "es2017",
+    "target": "es2020",
```
xavdid-stripe, line 6: i'm going to bump this again soon, but it's fine as is

src/V2Int64.ts
```diff
@@ -78,6 +86,12 @@ export const coerceV2ResponseData = (
       }
       return data;
 
+    case 'decimal_string':
+      if (typeof data === 'string') {
+        return Decimal.from(data);
```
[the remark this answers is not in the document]
xavdid-stripe: not a bad suggestion

src/V2Coercion.ts
```diff
@@ -1,3 +1,4 @@
+import {Decimal} from './Decimal.js';
```
xavdid-stripe, line 1: it's a nit, but this file is for Int64- either we should rename the file or split out the functionality (seems like the former is better)

types/apiVersion.d.ts
jar-stripe: this shouldnt have been deleted; its back now!

src/V2Coercion.ts
jar-stripe: renaming to V2Coercion.ts

Issue #2625, opened 2026-03-26
Stripe.Decimal is undefined at runtime even though TypeScript doesn't complain


kenmueller:
### Describe the bug

`Stripe.Decimal` is `undefined` even though typescript doesn't complain. My temporary patch:

```ts
// @ts-expect-error Stripe.Decimal is undefined at runtime
import { type Stripe, Decimal as _StripeDecimal } from 'stripe';

// eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
const StripeDecimal = _StripeDecimal as typeof Stripe.Decimal;
```

### To Reproduce

```ts
import { Stripe } from 'stripe';
Stripe.Decimal.from('1.0');
```

### Expected behavior

Works at runtime

### Code snippets

```JavaScript

```

### OS

macOS

### Node version

24.14.1

### Library version

21.0.0

### API version

2026-03-25.dahlia

prathmesh-stripe:
Thanks for reporting this. I was able to reproduce this error and we will ship a fix for it today.

We are also working on a TypeScript version of the Node SDK which we will release early next week and that should resolve your concern around type errors.

prathmesh-stripe:
The latest release([v21.0.1](https://github.com/stripe/stripe-node/releases/tag/v21.0.1)) should fix your issue.

kenmueller:
Thanks, that fixed it.

PR #2626, opened 2026-03-26
Fix export for Stripe.Decimal in CJS and ESM

prathmesh-stripe:
### Why?
User reported an issue where they were unable to access `Stripe.Decimal` in the Node SDK while using TypeScript and ESM import. 

Decimal class wasn't exported correctly and this is a problem with both CJS and ESM style of imports.

### What?
- Added a Decimal property on the exported Stripe class. 
- Added integration tests 

### See Also
https://github.com/stripe/stripe-node/issues/2625

## Changelog
* Fixes runtime error when using Stripe.Decimal. Resolves [#2625](https://github.com/stripe/stripe-node/issues/2625)

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -13,6 +13,7 @@ import {
 import {createWebhooks} from './Webhooks.js';
 import {ApiVersion} from './apiVersion.js';
 import {CryptoProvider} from './crypto/CryptoProvider.js';
+import {Decimal} from './Decimal.js';
 import {HttpClient, HttpClientResponse} from './net/HttpClient.js';
 import {PlatformFunctions} from './platform/PlatformFunctions.js';
 import * as resources from './resources.js';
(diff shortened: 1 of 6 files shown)


src/stripe.core.ts
```diff
@@ -82,6 +83,7 @@ export function createStripe(
   Stripe.HttpClientResponse = HttpClientResponse;
   Stripe.CryptoProvider = CryptoProvider;
   Stripe.webhooks = createWebhooks(platformFunctions);
+  Stripe.Decimal = Decimal;
```
mbroshi-stripe, line 86: Can you add a test for this?

jar-stripe: Hmm...we have several tests for this.  Let me research why they didn't catch it.

jar-stripe: Leaving this for posterity.  There was a bug in the test bed I was using to test the Decimal types.  This has been fixed, verified the original breakage and then also verified the fix within this PR works.

PR #2631, opened 2026-03-26
Prathmesh/merge node beta

prathmesh-stripe:
### Why?

Ran merge script and resolved conflicts. 
Includes Node SDK's Typescript changes.

### DIFF
--- a/src/Webhooks.ts
+++ b/src/Webhooks.ts
@@ -53,7 +53,7 @@ export type WebhookObject = {
     payload: WebhookPayload,
     header: WebhookHeader,
     secret: string,
-    tolerance?: null,
+    tolerance?: number,
     cryptoProvider?: CryptoProvider,
     receivedAt?: number
   ) => Event;
@@ -81,7 +81,7 @@ export function createWebhooks(
       payload: WebhookPayload,
       header: WebhookHeader,
       secret: string,
-      tolerance?: null,
+      tolerance?: number,
       cryptoProvider?: CryptoProvider,
       receivedAt?: number
     ): Event {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -306,6 +306,11 @@ export function protoExtend<T extends Record<string, any>>(
   // eslint-disable-next-line @typescript-eslint/no-this-alias
   const Super = this;
 
+  // This initialization logic is somewhat sensitive to be compatible with
+  // divergent JS implementations like the one found in Qt. See here for more
+  // context:
+  //
+  // https://github.com/stripe/stripe-node/pull/334
   // Create a subclass that properly extends the parent class (works with ES6 classes)
   const Constructor = class extends Super {
     constructor(...args: any[]) {
--- a/test/stripe.spec.ts
+++ b/test/stripe.spec.ts
@@ -7,16 +7,15 @@ import {expect} from 'chai';
 import {StripeSignatureVerificationError} from '../src/Error.js';
 import {ApiVersion} from '../src/apiVersion.js';
 import {createApiKeyAuthenticator, detectAIAgent} from '../src/utils.js';
-import {Stripe} from '../src/stripe.core.js';
+import Stripe = require('../src/stripe.cjs.node.js');
 import {NodePlatformFunctions} from '../src/platform/NodePlatformFunctions.js';
-
+import {Stripe as StripeCore} from '../src/stripe.core.js';
 import {
   FAKE_API_KEY,
   getRandomString,
   getStripeMockClient,
   getTestServerStripe,
 } from './testUtils.js';
-import StripeModule = require('../src/stripe.cjs.node.js');
 import crypto = require('crypto');
 import {StripeContext} from '../src/StripeContext.js';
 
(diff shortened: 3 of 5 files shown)


src/StripeEventNotificationHandler.ts
```diff
@@ -1,8 +1,93 @@
...
+  event: Events.UnknownEventNotification,
+  client: Stripe,
+  details: UnhandledNotificationDetails
+) => Promise<void>;
 
 // this is an internal-only type; we write a user-facing one separately
 type HandlerCallback = (event: any, client: any) => Promise<void>;
 
+// event-notification-map: The beginning of the section generated from our OpenAPI spec
+interface EventNotificationMap {
```
xavdid-stripe: Can you talk about why you added this? We should be able to use `EventNotification['type']` like before

src/StripeEventNotificationHandler.ts
```diff
@@ -85,16 +170,23 @@ export class StripeEventNotificationHandler {
...
-    ) => Promise<void>
+    private fallbackCallback: FallbackCallback
   ) {}
 
   // these types are duplicated in the manual types
+  public on<T extends keyof EventNotificationMap>(
+    type: T,
+    callback: (event: EventNotificationMap[T], client: Stripe) => Promise<void>
+  ): this;
+  public on(
```
xavdid-stripe: we don't want this second constructor- if someone is using an unknown type, they should have to explicitly suppress the TS error

src/StripeEventNotificationHandler.ts
prathmesh-stripe: To support types on 
```
public on<T extends keyof EventNotificationMap>(
    type: T,
    callback: (event: EventNotificationMap[T], client: Stripe) => Promise<void>
  ): this;
```

xavdid-stripe: Right, but the types worked before when they used the existing interfaces:

```ts
      on<T extends Stripe.V2.Core.EventNotification['type']>(
        eventType: T,
        callback: (
          eventNotification: Extract<
            Stripe.V2.Core.EventNotification,
            {type: T}
          >,
          client: Stripe
        ) => void
      ): this;
```

src/utils.ts
```diff
@@ -299,28 +299,35 @@ export function getOptionsFromArgs(args: RequestArgs): Options {
...
+): {new (...args: any[]): StripeResourceObject & T} {
   // eslint-disable-next-line @typescript-eslint/no-this-alias
   const Super = this;
-  const Constructor = Object.prototype.hasOwnProperty.call(sub, 'constructor')
-    ? sub.constructor
-    : function(this: StripeResourceObject, ...args: any[]): void {
-        Super.apply(this, args);
-      };
 
-  // This initialization logic is somewhat sensitive to be compatible with
```
xavdid-stripe, line 314: we probably don't want to pull this comment- seems useful still

src/Webhooks.ts
```diff
@@ -52,18 +53,18 @@ export type WebhookObject = {
     payload: WebhookPayload,
     header: WebhookHeader,
     secret: string,
-    tolerance: null,
-    cryptoProvider: CryptoProvider,
-    receivedAt: number
-  ) => WebhookEvent;
+    tolerance?: null,
```
xavdid-stripe: was this change intentional?

src/StripeEventNotificationHandler.ts
prathmesh-stripe: Fixed

test/resources/V2/Core/Events.spec.js
```diff
@@ -31,7 +31,7 @@ const v2EventPayloadWithRelatedObject = `
     "created": "1970-01-12T21:42:34.472Z",
     "id": "obj_123",
     "livemode": true,
-    "object":"v2.core.event",
+    "object":"v2.core.event_destination.ping",
```
xavdid-stripe: this shouldn't change - the `object` is still `v2.core.event`. `.ping` could go in `type` though

test/stripe.spec.ts
```diff
@@ -26,36 +29,39 @@ const CUSTOMER_DETAILS = {
 
 describe('Stripe Module', function() {
   describe('config object', () => {
+    /* eslint-disable no-warning-comments, no-new */
     it('should only accept either an object or a string', () => {
       expect(() => {
-        Stripe(FAKE_API_KEY, 123);
+        new StripeModule(FAKE_API_KEY, 123);
```
xavdid-stripe: was this change in line with the breaking changes you expected?

testProjects/mjs-ts/package.json
```diff
@@ -3,12 +3,14 @@
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
+  "type": "module",
   "dependencies": {
     "stripe": "file:../../",
-    "ts-node": "^10.9.1"
+    "ts-node": "^10.9.1",
```
xavdid-stripe: if we're not using `ts-node` anymore, it can be removed from these deps

justfile
```diff
@@ -65,16 +66,11 @@ prettier *args: install
...
-format: (prettier "--write --loglevel silent") _update-api-version
+format: (prettier "--write --loglevel silent")
 
 # verify formatting of files (without changes)
 format-check: (prettier "--check")
 
-# propagate automatic changes; should be run after generation
-# in practice, that means it runs after formatting, since that's the only recipe that the generator calls
-_update-api-version:
-    cp src/apiVersion.ts types/apiVersion.d.ts
```
xavdid-stripe, line 76: does API version live somewhere else now?

prathmesh-stripe: https://github.com/stripe/stripe-node/blob/master/src/apiVersion.ts
It is just not coped to the types folder.

src/Webhooks.ts
prathmesh-stripe: https://github.com/stripe/stripe-node/blob/0291e44ee2d91c04bae5b0173a1756922fd12335/types/Webhooks.d.ts#L35-L44

prathmesh-stripe: I'll update this to `tolerance?: number,`

test/resources/V2/Core/Events.spec.js
prathmesh-stripe: Fixed

xavdid-stripe (approved):

PR #2633, opened 2026-03-26
update extensibility SDK name in comments

xavdid-stripe:
### Why?

The Stripe Apps team asked that we update the name of their SDK in our comments

### What?

- update comments

### See Also

[RUN_DEVSDK-2299](https://go/j/RUN_DEVSDK-2299)

### DIFF
--- a/src/Decimal.ts
+++ b/src/Decimal.ts
@@ -73,7 +73,7 @@ export interface DecimalRoundingOptions {
  * {@link DecimalImpl.round}:
  *
  * ```ts
- * declare module '@stripe/extensibility-sdk/stdlib' {
+ * declare module 'stripe' {
  *   interface DecimalRoundingPresets {
  *     'my-custom-preset': DecimalRoundingOptions;
  *   }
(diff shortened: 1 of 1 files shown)


xavdid-stripe:
Fair, i've updated! 

I've updated the docs so that they work, but i'm a little concerned that we're not exporting the things the doc expects. For instance, you can't type something as a `Stripe.Decimal`, since it's only exported as a class, not a type. So a user couldn't write a function that accepted a stripe decimal type, for instance. 

Do we need to update our exports?

src/Decimal.ts
```diff
@@ -1015,9 +1015,9 @@ class DecimalImpl {
  *
  * @example
  * ```ts
- * import { Decimal, RoundDirection } from '@stripe/apps-extensibility-sdk/stdlib';
```
xavdid-stripe, line 1018: This wouldn't have worked - everything is imported under the root `Stripe` namespace

jar-stripe:
> For instance, you can't type something as a `Stripe.Decimal`, since it's only exported as a class, not a type. So a user couldn't write a function that accepted a stripe decimal type, for instance.

My understanding was that if exported but not as a type, you can do anything with it (i.e. its only restricted if its a type export).  I'll confirm.

jar-stripe:
> > For instance, you can't type something as a `Stripe.Decimal`, since it's only exported as a class, not a type. So a user couldn't write a function that accepted a stripe decimal type, for instance.
> 
> My understanding was that if exported but not as a type, you can do anything with it (i.e. its only restricted if its a type export). I'll confirm.

You are in fact correct.  I'll put up a PR.

xavdid-stripe:
You're welcome to take this one over! It came in as a run ticket so I was grabbing, but you seem to have a lot more context

jar-stripe:
Folded into #2634 — the stale `@stripe/apps-extensibility-sdk/stdlib` comments were deleted entirely rather than renamed.

PR #2638, opened 2026-03-31
Converted V2/Amount.ts to V2/V2Amount.ts

prathmesh-stripe:
### Why?

Converted V2/Amounts.ts to V2Amounts.ts. Conflicts originated in private-preview spec due to usage of Amount.ts class. See https://github.com/stripe/stripe-node/pull/2636 for exact file.

### What?
* Changed filename of V2/Amounts.ts to V2/V2Amounts.ts. This is to avoid conflicts in imported types and generated types in some deeply nested structures. (See CreditGrant.Amount in V2/Billing/ServiceActions.ts).
* Ran generator that caused the files with an extra new line.
* Files to review:
  * Anything with more than 1 line diff 
  * src/stripe.core.ts
  * src/stripe.esm.ts
  * src/resources/V2/V2Amounts.ts
  * src/resources/V2/Core/Events.ts
  * src/resources/V2/Core/AccountTokens.ts

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -49,7 +49,6 @@ import {
   Decimal,
 } from './shared.js';
 import {EventNotification as V2EventNotification} from './resources/V2/Core/Events.js';
-import {V2Amount} from './resources/V2/V2Amounts.js';
 
 // StripeInstanceImports: The beginning of the section generated from our OpenAPI spec
 import {
@@ -2458,5 +2457,4 @@ export declare namespace Stripe {
   export {StripeRawError};
   export import ErrorType = _Error;
   export import Events = V2.Core.Events;
-  export {V2Amount};
 }
--- a/src/stripe.esm.node.ts
+++ b/src/stripe.esm.node.ts
@@ -50,7 +50,6 @@ import {
   Decimal,
 } from './shared.js';
 import {EventNotification as V2EventNotification} from './resources/V2/Core/Events.js';
-import {V2Amount} from './resources/V2/V2Amounts.js';
 
 // StripeInstanceImports: The beginning of the section generated from our OpenAPI spec
 import {
@@ -2458,7 +2457,6 @@ export declare namespace Stripe {
   export {StripeRawError};
   export import ErrorType = _Error;
   export import Events = V2.Core.Events;
-  export {V2Amount};
 }
 
 Stripe.initialize(new NodePlatformFunctions());


src/stripe.core.ts
```diff
@@ -2457,4 +2458,5 @@ export declare namespace Stripe {
   export {StripeRawError};
   export import ErrorType = _Error;
   export import Events = V2.Core.Events;
+  export {V2Amount};
```
prathmesh-stripe, line 2461: I'm removing these exports so that we can change this easily in future if needed.

src/resources/V2/Core/Events.ts
```diff
@@ -104,334 +118,948 @@ export interface EventBase {
...
+ */ export interface UnknownEventNotification extends EventNotificationBase {
   /**
    * Object containing the reference to API resource relevant to the event.
-   */
-  export interface RelatedObject {
-    /**
-     * Unique identifier for the object relevant to the event.
-     */
-    id: string;
+   */ related_object: V2.Core.Events.RelatedObject | null;
```
xavdid-stripe, line 266: I don't think it matters, but it's weird that these moved to share a line with their block comment

src/resources/V2/Core/Events.ts
```diff
@@ -104,334 +118,948 @@ export interface EventBase {
...
+  | V2CoreEventDestinationPingEvent;
+
 export type EventNotification =
-  | Events.V1BillingMeterErrorReportTriggeredEventNotification
-  | Events.V1BillingMeterNoMeterFoundEventNotification
-  | Events.V2CoreEventDestinationPingEventNotification;
+  | V1BillingMeterErrorReportTriggeredEventNotification
+  | V1BillingMeterNoMeterFoundEventNotification
+  | V2CoreAccountClosedEventNotification
+  | V2CoreAccountCreatedEventNotification
```
xavdid-stripe, line 221: I don't believe the rest of these events are GA yet, right? master doesn't have them

prathmesh-stripe: These are GA. Master was hardcoded, now the file is generated.

prathmesh-stripe: Commit earlier has all these events: https://github.com/stripe/stripe-node/blob/25b1b03aa8c58d53d93db32ec71ac419d5b0dc52/types/V2/Core/EventTypes.d.ts#L5-L47

src/resources/V2/Core/Events.ts
prathmesh-stripe: I'll fix this in codegen.

xavdid (approved):

PR #2642, opened 2026-04-01
Update README.md

prathmesh-stripe:
### Why?

- Flipped around the order of ESM and CJS snippets in README. Since ESM is more widely used now, ESM snippets shows up first and CJS snippet follows it.

- Snippets use the term StripeClient now. 
- Also updated the stripe import statements in CJS to be multiple lines now.


README.md
```diff
@@ -41,27 +41,30 @@ value:
...
+  email: 'customer@example.com',
+});
+
+console.log(customer.id);
+```
+
+Or using CJS:
+```js
+const Stripe = require('stripe');
+const stripeClient = S('sk_test_...');
```
xavdid: typo: `S('sk_test_...')`

prathmesh-stripe: Fixed. Thanks

PR #2645, opened 2026-04-02
⚠️ Remove `stripeMethod` and standardize how function args are handled (including removing callback support)

xavdid:
### Why?

The Node SDK's API methods have always been a little unusual among our SDKs. Rather than generate API calls directly, it generated "method specs" which included a path, expected arguments, and response type (among others). Then, because the runtime function couldn't know anything about the incoming arguments, it shipped a bunch of logic to extract path arguments, params objects, options objects, and API keys. This has worked well enough, but was overly generous about the inputs it would handle. This has lead to both bugs and surprising behavior.

So the goal of this PR is simple: **fully type our runtime code to match the generated type signatures**.

As a result, we're removing a lot of ways it's been _possible_ to call these methods. All of these used to be valid, but would show type errors during editing:

```ts
stripe.customers.retrieve('cus_123', 'sk_test_123')
stripe.customers.create({name: 'david', host: 'example.com'}, 'sk_test_123')
stripe.customers.create({apiKey: 'sk_test_123'})
stripe.customers.list(customers => {
  // do something with customers
})
```

Now, callers have to conform to a much more consistent (and well typed) signature:

```ts
  create(
    params?: Checkout.SessionCreateParams,
    options?: RequestOptions
  ): Promise<Response<Session>>;
```

This means we're dropping support for:

- callbacks
- non-path-param strings
- `RequestOptions` as the first argument if the method takes `Params`
- Keys being allowed on either `params` or `options` objects

This gets us a _much_ more consistent HTTP request experience internally and reduces the opportunity for surprising behavior.

### What?

- refactor every generated method to call `this._makeRequest` with the proper inputs instead of going through `StripeMethod`
- remove `StripeMethod` and a bunch of supporting code (like url param replacers)
- updated many tests to not use callbacks anymore (without changing functionality)
- removed unused tests
- removed `.method` and `.extend` functions - these were never meant to be public facing and weren't used outside our tests
- update signature for `StripeResource._makeRequest` to take method, path, params, etc instead of an untyped array. Update body accordingly
- fixed some internal docs links
- disambiguated a `RequestOptions` type since only one was user-facing

### See Also

- [DEVSDK-3060](https://go/j/DEVSDK-3060)

## Changelog

- ⚠️ Refactor how incoming method arguments are parsed. Type signatures for API methods should be _much_ more accurate and reliable now
  - ⚠️ Remove support for providing callbacks to API methods. Use `async / await` instead
  - ⚠️ Remove support for passing a plain API key as a function arg. If supplied on a per-request basis, it should be in the `RequestOptions` under the `apiKey` property
  - ⚠️ Keys from `params` and `options` objects are no longer mixed. If present on a method, `RequestParams` must always come first and `RequestOptions` must always come second. To supply options without params, pass `undefined` as the first argument explicitly
  - ⚠️ Removed methods from `StripeResource`: `createFullPath`, `createResourcePathWithSymbols`, `extend`, `method` and `_joinUrlParts`. These were mostly intended for internal use and we no longer need them

As a result, the following call patterns are no longer supported:

```ts
stripe.customers.retrieve('cus_123', 'sk_test_123')
stripe.customers.create({name: 'david', host: 'example.com'}, 'sk_test_123')
stripe.customers.create({apiKey: 'sk_test_123'})
stripe.customers.list(customers => {
  // do something with customers
})
```

If those look familiar, head over to the [migration guide](https://github.com/stripe/stripe-node/wiki/Migration-guide-for-v22) to update your code.

### DIFF
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -323,6 +323,17 @@ function getDoneCallback(args: Array<any>): IterationDoneCallback | null {
   return onDone;
 }
 
+/**
+ * We allow four forms of the `onItem` callback (the middle two being equivalent),
+ *
+ *   1. `.autoPagingEach((item) => { doSomething(item); return false; });`
+ *   2. `.autoPagingEach(async (item) => { await doSomething(item); return false; });`
+ *   3. `.autoPagingEach((item) => doSomething(item).then(() => false));`
+ *   4. `.autoPagingEach((item, next) => { doSomething(item); next(false); });`
+ *
+ * In addition to standard validation, this helper
+ * coalesces the former forms into the latter form.
+ */
 function getItemCallback<T>(
   args: Array<any>
 ): IterationItemCallback<T> | undefined {
@@ -347,6 +358,10 @@ function getItemCallback<T>(
     );
   }
 
+  // This magically handles all three of these usecases (the latter two being functionally identical):
+  // 1. `.autoPagingEach((item) => { doSomething(item); return false; });`
+  // 2. `.autoPagingEach(async (item) => { await doSomething(item); return false; });`
+  // 3. `.autoPagingEach((item) => doSomething(item).then(() => false));`
   return function _onItem(item, next): void {
     const shouldContinue = onItem(item);
     next(shouldContinue);
@@ -450,6 +465,9 @@ function wrapAsyncIteratorWithCallback<T>(
 
       const item = iterResult.value;
       return new Promise((next) => {
+        // Bit confusing, perhaps; we pass a `resolve` fn
+        // to the user, so they can decide when and if to continue.
+        // They can return false, or a promise which resolves to false, to break.
         onItem(item, next);
       }).then((shouldContinue) => {
         if (shouldContinue === false) {


src/resources/Billing/Alerts.ts
```diff
@@ -14,92 +13,84 @@ export class AlertResource extends StripeResource {
...
     params?: Billing.AlertListParams,
     options?: RequestOptions
-  ): ApiListPromise<Alert>;
-  list(options?: RequestOptions): ApiListPromise<Alert>;
-  list(...args: any[]): Promise<Response<any>> {
-    return stripeMethod({
-      method: 'GET',
-      fullPath: '/v1/billing/alerts',
+  ): ApiListPromise<Alert> {
+    return this._makeRequest('GET', '/v1/billing/alerts', params, options, {
```
xavdid, line 17: Basically all of the generated resources changes are the same pattern: instead of making a `stripeMethod`, just make the HTTP request like we do in all other langs. Also, remove overload signatures and just use the one that specifies both params and options, if applicable

src/resources/Billing/Alerts.ts
```diff
@@ -14,92 +13,84 @@ export class AlertResource extends StripeResource {
...
       methodType: 'list',
-    }).call(this, ...args);
+    }) as any;
```
xavdid, line 19: need this so we can change the return type of `_makeRequest` into the resource we say we're returning.

We could make `_makeRequest` generic, but at some point in the chain TS has to trust that we're doing what we say we are, so there's not much point.

src/lib.ts
```diff
@@ -133,11 +113,13 @@ export interface RequestOptions {
 
   /**
    * An account id on whose behalf you wish to make a request.
+   *
+   * NOTE: prefer sending `stripeContext` instead of `stripeAccount` for new code. They're currently identical, but we will eventually discourage and (later) drop support for `stripeAccount`.
```
xavdid, line 117: This comment was always supposed to be here, but I accidentally put it on an internal-facing type

src/lib.ts
```diff
@@ -156,6 +138,21 @@ export interface RequestOptions {
    * Specify a timeout for this request in milliseconds.
    */
   timeout?: number;
+
+  /**
+   * Provide a custom authenticator function for this request.
+   */
+  authenticator?: import('./Types.js').RequestAuthenticator;
```
xavdid, line 145: We always supported these, we just never had types for them

src/RequestSender.ts
xavdid, line 479: this got the same treatment as _request

src/Types.ts
```diff
@@ -91,14 +91,14 @@ export type RequestEvent = {
   usage?: Array<string>;
 };
 export type RequestHeaders = Record<string, string | number | string[]>;
-export type RequestOptions = {
+/**
+ * this is similar but distinct from the user-facing `RequestOptions`
+ */
+export type InternalRequestOptions = {
```
xavdid, line 97: Disambiguate the two (slightly different) options shapes

src/utils.ts
```diff
@@ -176,43 +175,6 @@ export function extractUrlParams(path: string): Array<string> {
   return params.map((param) => param.replace(/[{}]/g, ''));
 }
 
-/**
- * Return the data argument from a list of arguments
- *
- * @param {object[]} args
- * @returns {object}
- */
-export function getDataFromArgs(args: RequestArgs): RequestData {
```
xavdid, line 185: This helped us turn ambiguous strings and placeholders into urls, but we just generate those now

test/resources/Account.spec.js
```diff
@@ -88,30 +88,6 @@ describe('Account Resource', () => {
         settings: {},
       });
     });
-
-    it('Sends the correct request with secret key as first object', () => {
```
xavdid, line 92: this ended up being the same test as above, since we don't support options objects in the first position anymore

test/resources/Account.spec.js
```diff
@@ -88,30 +88,6 @@ describe('Account Resource', () => {
...
-        auth: params.apiKey,
-        method: 'GET',
-        url: '/v1/account',
-        data: null,
-        headers: {},
-        settings: {},
-      });
-    });
-
-    it('Sends the correct request with a callback', () => {
```
xavdid, line 105: we don't support this anymore!

test/resources/Account.spec.js
```diff
@@ -225,7 +201,8 @@ describe('Account Resource', () => {
         stripe.account.retrieveExternalAccount(
           'accountIdFoo321',
           'externalAccountIdFoo456',
-          TEST_AUTH_KEY
+          undefined,
```
xavdid, line 204: these changes are basically all just using the more correct call shape

test/resources/generated_examples_test.spec.js
```diff
@@ -4361,9 +4361,12 @@ describe('Generated tests', function() {
         },
       });
 
-    await realStripe.v2.core.accounts.list((err) => {
```
xavdid, line 4364: no more callbacks

src/resources/Billing/Alerts.ts
jar-stripe: can we use unknown instead of any?  its a bit safer against accidental assignment

xavdid: unfortunately not. We get `Type 'unknown' is not assignable to type 'Promise<Response<Account>>` since it's not sure we can do the conversion. 

This the same thing we do in Python - we `cast` in those generated calls because we know it's correct and we have to match a shape to a user-facing type. It's simpler in node because they're all just interfaces, no runtime types

PR #2659, opened 2026-04-04
Add Ref<T> type for typed v2 object references

jar-stripe:
### Why?

Custom objects need typed references to other Stripe objects. A reference has the wire shape `{type, id, url}` and in the SDK becomes `Ref<T>` — a container that can fetch the referenced object with TypeScript type safety.

### What?

- `Stripe.V2.Ref<T>` interface with `type`, `id`, `url` fields and `fetch(): Promise<T>`
- `attachRefFetch<T>` attaches `fetch()` using a `Stripe` instance and optional `targetSchema`
- `fetch()` calls `makeRequest` (extracted standalone function from `StripeResource._makeRequest`) so the response goes through the same pipeline as service methods — including v2 type coercion via `responseSchema`
- `makeRequest` extracted from `StripeResource._makeRequest`: accepts a `Stripe` instance + optional `requestDataProcessor`. `_makeRequest` delegates to it. This lets non-resource code (Ref) use the same request pipeline.
- `coerceV2ResponseData` takes a required `Stripe` instance (not optional) and passes it to `attachRefFetch` for recursive fetch support (a.b.fetch().c.fetch() works)
- `V2RuntimeSchema` `refObject` kind carries optional `targetSchema` for the target resource's coercion schema
- `V2Coercion` handles `refObject` in both request (passthrough) and response (attach fetch + coerce) paths

### See Also

Generated with [Claude Code](https://claude.com/claude-code)


src/resources/V2/Ref.ts
```diff
@@ -0,0 +1,49 @@
...
+ * `requestSender`.
+ */
+export const attachRefFetch = <T>(
+  ref: RefWireShape,
+  requestSender: RequestSender
+): Ref<T> => {
+  return {
+    ...ref,
+    fetch(): Promise<T> {
+      return requestSender._rawRequest('GET', ref.url, undefined, {}, [
```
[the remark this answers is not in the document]
jar-stripe: The returned object is frozen (spread copy) and `ref` is the same object — there's no mutation scenario in practice since these come from parsed API responses. But capturing `url` as a const for clarity is a good idea. Will update.

src/resources/V2/Ref.ts
```diff
@@ -0,0 +1,49 @@
...
+        'ref_fetch',
```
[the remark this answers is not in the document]
jar-stripe: Will add a test for `attachRefFetch` — should verify it issues a GET to `ref.url` and records the usage telemetry tag.

src/resources/V2/Ref.ts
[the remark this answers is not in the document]
jar-stripe: Actually, on reflection — the returned object is a spread copy from a parsed API response, not something consumers mutate. And "fetch uses the URL the ref was created with" is the correct semantic even if the copy were mutated. Leaving as-is.

src/resources/V2/index.ts
```diff
@@ -4,6 +4,9 @@ import {Stripe} from '../../stripe.core.js';
 import {DeletedObject} from './DeletedObject.js';
 import {Billing} from './Billing/index.js';
 import {Core} from './Core/index.js';
+import {Ref, RefWireShape} from './Ref.js';
+
+export {attachRefFetch} from './Ref.js';
```
xavdid, line 9: Is this export needed? Most of the other internal callsites import directly from the definition, but one pulls it from here.

src/resources/V2/Ref.ts
```diff
@@ -0,0 +1,62 @@
...
+ * that resolves to the full object.
+ */
+export interface Ref<T> {
+  /** String representing the type of the referenced object. */
+  type: string;
+
+  /** Unique identifier for the referenced object. */
+  id: string;
+
+  /** URL that can be used to retrieve the full object. */
```
xavdid, line 17: nit: it's a url _path_, like `/v2/core/account/acct_123`

src/resources/V2/Ref.ts
```diff
@@ -0,0 +1,62 @@
...
+  url: string;
+
+  /** Fetches the full object this ref points to. */
+  fetch(): Promise<T>;
```
xavdid, line 21: I don't _love_ attaching the client to the `Ref` interface. It tightly couples their retrieval to the shape. I know we do it for event notifications, but that's just JSON we're imparting meaning to. We'll presumably be interacting with `Ref`s we've retrieved from the API, so we've already got an active client.

We definitely want to make retrieval easy. How do you feel about `client.fetch(ref)`? It could pick up the inner type of the `Ref` and be strongly typed as well, but the actual `ref` object is just the `{type, id, url}`, so it's easier to use.

testProjects/types/typescriptTest.ts
```diff
@@ -349,6 +349,8 @@ async (): Promise<void> => {
   let f: Stripe.Events.V1BillingMeterErrorReportTriggeredEventNotification;
   // union of all V2 Events
   let g: Stripe.V2.Core.Event;
+  // V2 Ref type
+  let h: Stripe.V2.Ref<Stripe.V2.Core.Account>;
```
xavdid, line 353: Should add a test for the typing of `fetch` (however we end up implementing it)- fetching `h` here should yield a `V2.Core.Account`.

src/V2Coercion.ts
```diff
@@ -98,14 +104,28 @@ export const coerceV2ResponseData = (
       }
       return data;
 
+    case 'refObject':
+      if (typeof data === 'object' && !Array.isArray(data)) {
+        return attachRefFetch(
+          data as RefWireShape,
```
xavdid, line 110: rather than than the hard type cast here, is it worth writing a type guard to handle the assertion? or this isn't a concern here?

src/V2Coercion.ts
```diff
@@ -98,14 +104,28 @@ export const coerceV2ResponseData = (
...
         return data;
       }
       const obj = data as Record<string, unknown>;
       for (const key of Object.keys(schema.fields)) {
         if (key in obj) {
-          obj[key] = coerceV2ResponseData(obj[key], schema.fields[key]);
+          obj[key] = coerceV2ResponseData(
+            obj[key],
+            schema.fields[key],
+            stripe
```
xavdid, line 127: As an extension to my design comments above, it's surprising as a reader to see that `coerceV2ResponseData` needs a full `Stripe` object. Feels like a code smell

xavdid (commented):
Took a read through and had some design suggestions. I like standardizing around `Ref` as a primitive, but not sure this is the exact right approach.

jar-stripe:
Changing direction for now.

PR #2670, opened 2026-04-08
tolerance, receivedAt, timestamp are now optional in Webhook.ts

### LINKED ISSUE #2666: v22 - change to `generateTestHeaderString`
### Describe the bug

After upgrading to v22, the shape of the input type for `generateTestHeaderString` now requires more properties.

The old manually defined type made all properties except payload and secret optional (https://github.com/stripe/stripe-node/blob/9c0a9cb7482d19c2e6e6b5b5574c1a2f0149a2a9/types/Webhooks.d.ts#L104).

The existing type that is now exported has all properties required (https://github.com/stripe/stripe-node/blob/master/src/Webhooks.ts#L30).

### To Reproduce

1. Look at test under webhook signing examples: https://github.com/stripe/stripe-node/blob/master/examples/webhook-signing/test/main.ts#L63
2. Compare options to type in Webhook.ts
3. Observe the following error:
```
error TS2345: Argument of type '{ payload: string; secret: string; }' is not assignable to parameter of type 'WebhookTestHeaderOptions'.
  Type '{ payload: string; secret: string; }' is missing the following properties from type 'WebhookTestHeaderOptions': timestamp, scheme, signature, cryptoProvider

116     const stripeSignature = stripe.webhooks.generateTestHeaderString({
                                                                         ~
117       payload: payloadString,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
118       secret: testWebhookSecret,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
119     });
```

### Expected behavior

I expect there to be no typescript error when calling `generateTestHeaderString` as the stripe example test does (or I can update my usage to include all other properties, but I would expect the stripe example test to do the same).

### OS

Linux (Arch, btw)

### Node version

Node v24.14.1

### Library version

stripe-node v22.0.0

### API version

latest

### Additional context

_No response_

prathmesh-stripe:
### Why?

In v22, several webhook method parameters that were previously optional in manually maintained became required due to a type consolidation. 
`generateTestHeaderString` only truly requires `payload` and `secret` (all other fields have defaults). `verifyHeader`/`constructEvent` work without explicit `tolerance` and `cryptoProvider` arguments. Users upgrading to v22 now get TypeScript errors when calling these methods with the same arguments that worked in v21.

### What?
- Made `timestamp`, `scheme`, `signature`, and `cryptoProvider` optional in `WebhookTestHeaderOptions`
- Made `tolerance` and `cryptoProvider` optional in `verifyHeader`/`verifyHeaderAsync` signatures
- Made `scheme` optional in `parseHeader`
- Updated `prepareOptions` return type to reflect that defaulted fields are resolved to non-optional
- Added tests covering all optional parameter default behaviors

### See Also

- Fixes #2666

### DIFF
--- a/src/Webhooks.ts
+++ b/src/Webhooks.ts
@@ -233,6 +233,11 @@ export function createWebhooks(
         secret
       );
 
+      /**
+       * TODO(MAJOR): https://go/j/DEVSDK-3087
+       * Passing in 0 by default skips timestamp tolerance verifications. Although it is mostly used in test,
+       * we should change the default behavior to pass DEFAULT_TOLERANCE instead of 0 in the next major.
+       */
       validateComputedSignature(
         payload,
         header,
@@ -274,6 +279,11 @@ export function createWebhooks(
         secret
       );
 
+      /**
+       * TODO(MAJOR): https://go/j/DEVSDK-3087
+       * Passing in 0 by default skips timestamp tolerance verifications. Although it is mostly used in test,
+       * we should change the default behavior to pass DEFAULT_TOLERANCE instead of 0 in the next major.
+       */
       return validateComputedSignature(
         payload,
         header,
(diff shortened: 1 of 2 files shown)


src/Webhooks.ts
```diff
@@ -238,7 +238,7 @@ export function createWebhooks(
         header,
         details,
         expectedSignature,
-        tolerance,
+        tolerance || 0,
```
prathmesh-stripe, line 246: Alternate here is setting this to `DEFAULT_TOLERANCE` would make existing tests break. 
Setting this to 0 is same as passing undefined which skips the tolerance verification in the current flow preserving the nature of this method.

xavdid: I read through stuff and I think this is fine because it's how it's always worked. But, I think it's worth changing this behavior in the next major to default to 300 (and still allow users setting `0` to opt out)

I'd:

- add / comment a next-major ticket to change this
- add a docstring to `validateComputedSignature` that mentions that by default, the method doesn't do timestamp verification. It's mostly meant for tests or offline processing where the delivery time of the event isn't important. Integrations that process webhooks as they come in should use `constructEvent` instead.

xavdid (approved):
Looks good, but I had a comment about the default.

It's how it's always worked so we're not making things _worse_, but we should ship with better defaults in the future.

Issue #2665, opened 2026-04-07
v22 - Ability to easily log all requests without reparsing


TastyPi:
### Is your feature request related to a problem? Please describe.

Add the ability to log all requests/responses to a logger.

### Describe the solution you'd like

A parameter in the `Stripe` constructor which allows setting up a logger.

### Describe alternatives you've considered

- `Stripe.createFetchHttpClient` with a custom `fetch` implementation that logs requests - the reason we didn't use this is it would require reparsing the request body, which is awkward (we store request bodies as structured logs)
- `Stripe.on("request"/"response")` fires events when requests/responses happen, however they do not include the bodies of the request.

### Additional context

Before the latest version, we built a custom implementation of `Stripe.HttpClient` which logs every request/response and delegates to a real `Stripe.HttpClient`. This was easy to since it was just an interface, and had specific types for everything. Now, `Stripe.HttpClient` is exported as a class, which makes it more awkward to wrap.

I've also noticed that the second parameter of `Stripe`'s constructor is just a `Record<string, unknown>`, which means all the type safety has been lost.

mbroshi-stripe:
@TastyPi Thanks for bringing this to our attention! It sounds like there are two unrelated issues:
1. Give affordance for logging request/response bodies
2. Add type-safety to the second parameter in the `Stripe` constructor

We'll look into both issues and circle back if we have any additional questions.

prathmesh-stripe:
@TastyPi Now you can configure and emit event body to help with logging. This should be now part of the [v22.0.2 release](https://github.com/stripe/stripe-node/releases/tag/v22.0.2). I hope it solves your logging issue.

PR #2681, opened 2026-04-09
Add emitEventBodies config to include bodies in request/response events

mbroshi-stripe:
### Why?

Users want to log request/response bodies via the existing `request` and `response` events (see #2665). Currently these events only carry metadata (method, path, status, timing), so users who want structured logging have to build custom `HttpClient` wrappers and re-parse bodies.

### What?

- Added `emitEventBodies` config option (default `false`) to opt in to including bodies in events
- When enabled, the `request` event includes a `body` field with the original request data object (pre-serialization)
- When enabled, the `response` event includes a `body` field with the parsed JSON response object
- Moved response event emission to after JSON parsing (timing fields are still captured before parsing, so `elapsed`/`request_end_time` accuracy is preserved)
- `HttpClient` now uses the `HttpClientInterface` interface type instead of the `HttpClient` class, allowing users to provide custom implementations without extending the class
- Added 8 tests covering: config defaults, request bodies, response bodies, disabled mode, GET requests, error responses, and a structured logging integration test

### See Also

- #2665

## Changelog

- Added `emitEventBodies` config option to include request and response bodies in `request`/`response` events.
- Changed `httpClient` config type from `HttpClient` class to `HttpClientInterface` interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,7 @@ stripeClient.off('request', onRequest);
   idempotency_key: 'abc123',         // Only present if provided
   method: 'POST',
   path: '/v1/customers',
+  body: {name: 'test'},              // Only present if emitEventBodies is true
   request_start_time: 1565125303932  // Unix timestamp in milliseconds
 }
 ```
@@ -377,15 +378,16 @@ stripeClient.off('request', onRequest);
 ```js
 {
   api_version: 'latest',
-  account: 'acct_TEST',              // Only present if provided
-  idempotency_key: 'abc123',         // Only present if provided
+  account: 'acct_TEST',                       // Only present if provided
+  idempotency_key: 'abc123',                  // Only present if provided
   method: 'POST',
   path: '/v1/customers',
   status: 402,
   request_id: 'req_Ghc9r26ts73DRf',
-  elapsed: 445,                      // Elapsed time in milliseconds
-  request_start_time: 1565125303932, // Unix timestamp in milliseconds
-  request_end_time: 1565125304377    // Unix timestamp in milliseconds
+  body: {id: 'cus_123', object: 'customer'},  // Only present if emitEventBodies is true
+  elapsed: 445,                               // Elapsed time in milliseconds
+  request_start_time: 1565125303932,          // Unix timestamp in milliseconds
+  request_end_time: 1565125304377             // Unix timestamp in milliseconds
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ stripeClient.off('request', onRequest);
   idempotency_key: 'abc123',                  // Only present if provided
   method: 'POST',
   path: '/v1/customers',
-  status: 402,
+  status: 200,
   request_id: 'req_Ghc9r26ts73DRf',
   body: {id: 'cus_123', object: 'customer'},  // Only present if emitEventBodies is true
   elapsed: 445,                               // Elapsed time in milliseconds
--- a/src/RequestSender.ts
+++ b/src/RequestSender.ts
@@ -200,6 +200,12 @@ export class RequestSender {
             return jsonResponse;
           },
           (e: Error) => {
+            if (
+              this._stripe.getEmitEventBodiesEnabled() &&
+              (e as any).rawBody
+            ) {
+              responseEvent.body = (e as any).rawBody;
+            }
             throw new StripeAPIError({
               message: 'Invalid JSON received from the Stripe API',
               exception: e,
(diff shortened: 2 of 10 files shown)


src/RequestSender.ts
```diff
@@ -206,6 +205,11 @@ export class RequestSender {
         )
         .then(
           (jsonResponse) => {
+            if (this._stripe.getEmitEventBodiesEnabled()) {
+              responseEvent.body = jsonResponse;
+            }
+            this._stripe._emitter.emit('response', responseEvent);
+
```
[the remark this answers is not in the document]
mbroshi-stripe: Fixed in eb11515d. Moved the body assignment to right after `toJSON()` resolves, before the error check. Both success and error responses now include the body when `emitEventBodies` is enabled.

README.md
```diff
@@ -377,15 +378,16 @@ stripeClient.off('request', onRequest);
 ```js
 {
   api_version: 'latest',
-  account: 'acct_TEST',              // Only present if provided
-  idempotency_key: 'abc123',         // Only present if provided
+  account: 'acct_TEST',                       // Only present if provided
+  idempotency_key: 'abc123',                  // Only present if provided
   method: 'POST',
   path: '/v1/customers',
   status: 402,
```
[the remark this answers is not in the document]
mbroshi-stripe: Fixed in eb11515d. Changed to `status: 200`.

src/Types.ts
```diff
@@ -87,6 +87,7 @@ export type RequestEvent = {
   idempotency_key?: string;
   method?: string;
   path?: string;
+  body?: Record<string, any>;
```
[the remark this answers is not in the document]
mbroshi-stripe: Good call — updated `RequestEvent.body` to use the `RequestData` type alias in eb11515d. Kept `ResponseEvent.body` as `Record<string, any>` since there isn't a corresponding `ResponseData` type.

test/stripe.spec.ts
```diff
@@ -1215,4 +1225,154 @@ describe('Stripe Module', function() {
       );
     });
   });
+
+  describe('emitEventBodies', () => {
+    it('should include request body in request event when enabled', (done) => {
+      getTestServerStripe(
+        {emitEventBodies: true} as any,
```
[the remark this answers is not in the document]
mbroshi-stripe: Fixed in eb11515d. Widened `getTestServerStripe` to accept `StripeConfig` and removed all `as any` casts.

test/stripe.spec.ts
```diff
@@ -1215,4 +1225,154 @@ describe('Stripe Module', function() {
...
+          stripe.on('request', (event) => {
+            try {
+              expect(event.body).to.be.an('object');
+              expect(event.body).to.have.property(
+                'description',
+                'test customer'
+              );
+            } catch (e) {
+              done(e);
+            }
```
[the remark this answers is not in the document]
mbroshi-stripe: Fixed in eb11515d. All event handler tests now track an `eventFired` flag and assert it after the request completes.

src/RequestSender.ts
```diff
@@ -219,7 +224,10 @@ export class RequestSender {
 
             callback(null, jsonResponse);
           },
-          (e) => callback(e, null)
+          (e) => {
+            this._stripe._emitter.emit('response', responseEvent);
```
prathmesh-stripe, line 234: This is great. We emit response even in the case of invalid JSON. Does this also catch when toJSON() fails?

src/stripe.esm.node.ts
prathmesh-stripe, line 958: Shouldn't this also be an interface?

mbroshi-stripe: ~Yep, good catch.~ Actually, I'm not clear on why that should change. Can you expand on what you're proposing?

src/RequestSender.ts
mbroshi-stripe: We could. That would change `body?: Record<string, any>` --> `body?: Record<string, any> | string;`. I'll push up a commit and let me know if you think it's worthwhile (it's slightly invasive). In theory, the response should always be JSON.

mbroshi-stripe: Here's the commit: https://github.com/stripe/stripe-node/pull/2681/changes/f99e0f8f6477ea0c25fcc6014c4bdebf2a576501

prathmesh-stripe (approved):

src/stripe.esm.node.ts
prathmesh-stripe: I'll have to play around with this to see if there is an easier wya. Merging this because we can ship something today to help solve a user request.

PR #2685, opened 2026-04-14
Adds `parseEventNotificationAsync` to match existing sync function

pedroits:
### Why?
In environments like Supabase Edge Functions (Deno), Cloudflare Workers, and Vercel Edge:

- `constructEventAsync` does not support `v2.core.event` webhooks
- It suggests using `stripe.parseEventNotification`
- However, `parseEventNotification` relies on a synchronous crypto provider

This results in the runtime error:

```

SubtleCryptoProvider cannot be used in a synchronous context

````

Since these environments only support asynchronous cryptography, there is currently no way to validate `v2.core.event` webhooks.

### What?

- Added `parseEventNotificationAsync` implementation
- Preserved backward compatibility (no changes to existing sync APIs)
- updated tests

### Usage

```ts
const event = await stripe.parseEventNotificationAsync(
  body,
  signature,
  endpointSecret
);
```

### Notes

* No breaking changes
* Unblocks Stripe webhook validation in modern edge runtimes
  * Supabase Edge Functions
  * Cloudflare Workers
  * Vercel Edge Functions
* Aligns SDK behavior with existing async patterns (`constructEventAsync`)
  * About this, i didn't like how it was done (function/code duplication) but it was done following the same bad patterns as in `constructEvent` and `constructEventAsync`

### See Also

- stripe/stripe-node/issues/2684

### DIFF
--- a/src/Webhooks.ts
+++ b/src/Webhooks.ts
@@ -159,7 +159,7 @@ export function createWebhooks(
           : JSON.parse(payload);
       if (jsonPayload && jsonPayload.object === 'v2.core.event') {
         throw new Error(
-          'You passed an event notification to stripe.webhooks.constructEvent, which expects a webhook payload. Use stripe.parseEventNotification instead.'
+          'You passed an event notification to stripe.webhooks.constructEvent, which expects a webhook payload. Use stripe.parseEventNotificationAsync instead.'
         );
       }
       return jsonPayload;
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -1617,6 +1617,85 @@ export class Stripe {
 
     return eventNotification;
   }
+
+  async parseEventNotificationAsync(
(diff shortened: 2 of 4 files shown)


xavdid:
Thank you for the submission and sorry for the bug. Looks like the formatter needs to run but the PR itself looks good. 

> About this, i didn't like how it was done (function/code duplication) but it was done following the same bad patterns as in constructEvent and constructEventAsync

Yeah, we'll get that cleaned up on our end, no worries.

xavdid (approved):

fredkzk:
Which version includes this? I cant find it in my node_modules yet. I'm with v22.0.1

prathmesh-stripe:
@fredkzk We shipped this in our latest [v22.0.2 release](https://github.com/stripe/stripe-node/releases/tag/v22.0.2).

Issue #1707, opened 2023-03-13
Errors miss useful stack traces


oleg-codaio:
### Describe the bug

When using `stripe-node`, API errors seem to lose the full stack trace, making it hard to debug and find the code that triggered a particular error.

For instance, if I make a bad `customers.retrieve()` call (with an invalid customer ID), I get an error with the following stack trace:

```
No such customer: 'some_fake_id'
Error: No such customer: 'some_fake_id'
    at Function.generate (/somedir/node_modules/.pnpm/stripe@11.7.0/node_modules/stripe/lib/Error.js:38:24)
    at res.toJSON.then.StripeAPIError.message (/somedir/node_modules/.pnpm/stripe@11.7.0/node_modules/stripe/lib/RequestSender.js:104:43)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### To Reproduce

1. Trigger a failing Stripe network request, such as `customers.retrieve`
2. Observe the thrown error's stack trace

### Expected behavior

I'd expect the stack trace to contain the stack up to where the Stripe API request was made. Instead, I only get a partial internal stack.

### OS

macOS

### Node version

Node v16.17.1

### Library version

11.7.0

### API version

2022-11-15

### Additional context

See also: https://github.com/stripe/stripe-node/pull/118

oleg-codaio:
To leave some notes on a proposed fix, I think the issue is that [async stack traces](https://v8.dev/docs/stack-trace-api#async-stack-traces) don't work due to use of `.then()` directly on Promises, instead of `await`:

https://github.com/stripe/stripe-node/blob/3dc8c54cf666bf5e4843b797542cc46252461560/src/RequestSender.ts#L144-L152

One solution is to save the original stack trace when the call is made via `new Error().stack`, then splat it into the stack of any thrown error later. This would have a small penalty of constructing stack traces on every Stripe request (which I'd imagine pales in comparison to the actual network request), so if that's a concern then this could be made optional too.

Issue #1542, opened 2022-09-01
Use async functions within request handling code


rtpg:
### Is your feature request related to a problem? Please describe.

Currently, there is a lot of `setTimeout`s and other non-async-y code that means that we cannot take advantage of V8's async stack trace feature. This means that we lose out on stack traces, the bare minimum that one would want for debugging an issue in production.

 While support for the async stack trace feature came in Node 12, support for the async/await syntax (which would allow for users of newer node to get a stack trace) has been present from Node 7.

### Describe the solution you'd like

At the very least, please use `async`/`await` syntax within the core request calling methods (`StripeResource`, `StripeMethod`), and remove usage of raw `setTimeout`, instead doing things like `await Promise((resolve, reject) => setTimeout(reject, 0))` to make sure that async stack traces are maintained.

 Doing this pervasively should make the code easier to manage, reduce a lot of indentation, and be functionally equivalent given how this all works.

### Describe alternatives you've considered

Monkeypatching is theoretically possible, and there are options to capture error traces myself, but that involves creating spurious error objects all over and increasing risks of updates.

### Additional context

https://github.com/stripe/stripe-node/issues/1066 is an example of this issue

rtpg:
I'm actually a _bit_ convinced  [the stack trace breakage was at one point being done intentionally](https://github.com/stripe/stripe-node/blob/18f88e487723cba994afd38dc645a709211e33b3/lib/utils.js#L317) from this utility function, at least for some methods.

While if you do things in the most straightforward way, retries will lead to deeper callstacks, there is another option that is also available and easy to express with `async`/`await`: an "iterative" for loop. It's understandably harder to express with promises since you can only go deeper in a callstack with promises. 

I might be totally offbase on the purpose of that utility function or even on how much using the `async` keyword will fix my problem, though. But I would appreciate somebody responsible for this library to look at it.

yejia-stripe:
Hi @rtpg! Thank you for the report! We'll look into this and circle back when we have more information. In the meantime, does the workaround from #1066 work for you?

rtpg:
@yejia-stripe sorry for taking a while to get back. Unfortunate the workaround does not work for usages around async iteration:

```
for await (const sub of wrappedStripe.subscriptions.list({limit: 10})) {
    console.log(sub)
```

^ this code snippet doesn't work with the proxying (I imagine because the list implements both async iteration and a standard promise mechanism).

rtpg:
[...]

Issue #1066, opened 2020-11-16
Missing stack trace on errors


AleG94:
Consider the following code snippet:

```js
const customerData = {}; // some customer data

try {
   await stripe.customers.create(customerData);
} catch (err) {
   console.error(err);
} 
```

When the api call fails due to any kind of error, the promise is rejected so an error is thrown and the catch block is triggered.

This is an example of error thrown by the `stripe-node` module: 

```js
Error: No such PaymentMethod: 'pm_card_vaaaaasa'
    at Function.generate (C:\Users\Alessandro\Documents\Projects\clickout\billing\node_modules\stripe\lib\Error.js:39:16)
    at IncomingMessage.<anonymous> (C:\Users\Alessandro\Documents\Projects\clickout\billing\node_modules\stripe\lib\StripeResource.js:190:33)
    at Object.onceWrapper (events.js:299:28)
    at IncomingMessage.emit (events.js:215:7)
    at endReadableNT (_stream_readable.js:1184:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
```

As you can see, the original stack trace is lost and there is no reference to the file and line of code that threw the error, making debug very difficult. 

For reference, this is the stack trace of a standard javascript `Error`, thrown on purpose (for this bug report) from the same file, just a line before the create customer call.

```js
error: Error: this is a standard error stack track
    at CustomerService._createCustomer (C:\Users\Alessandro\Documents\Projects\clickout\billing\app\services\customer.service.js:147:13)
    at C:\Users\Alessandro\Documents\Projects\clickout\billing\app\services\customer.service.js:50:25
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async LockService.withLock (C:\Users\Alessandro\Documents\Projects\clickout\billing\app\services\lock.service.js:16:22)
    at async CustomerService.create (C:\Users\Alessandro\Documents\Projects\clickout\billing\app\services\customer.service.js:21:22)
    at async handler (C:\Users\Alessandro\Documents\Projects\clickout\billing\app\api\v1\controllers\customer.controller.js:35:24)`
```

If you take a look at it, the stack trace contains a clear reference to the function, file and line that caused the error, along with upper level function calls.

_node: 12.13.1_
_stripe-node:  8.121.0_

brendanm-stripe:
Hey @AleG94 -- thanks for the example and sample output. I'm looking into this to see what's possible today and what might be doable in future.

[...]

PR #2688, opened 2026-04-17
Preserve stack traces for async operations

xavdid:
### Why?

Because of the way the internals of the SDK's async HTTP calling code are structured, when we throw an error, it loses its original stack trace. Users have only been able to trace issues back to where our HTTP calls originate from, not where in their code caused the HTTP call that failed.

The best solution is to use promises all the way down our HTTP chain, which will let node resolve stack traces as we'd expect. But that's a pretty big internal change and is part of why the team hadn't prioritized it in the past.

We talked it over this week and decided to take a more tactical approach: by manually instantiating an unthrown error before doing any async operations, we can reference its call stack if the async operation later fails. That gives us pretty usable tracebacks (seen below). I've also manually confirmed that it works for `autoPagingEach` and `autoPagingToArray`.

As far as I can tell, we didn't try this before because of perceived performance implications, but based on local testing with real API calls, the difference is negligible (approx. 30ms / request, which is indistinguishable from network jitter). We'll still probably go through and simplify this code in the future, but this is a reasonable stopgap until we get to that.

before:

```
Error: Invalid API Key provided: sk_test_fake
    at generateV1Error (/Users/xavdid/stripe/hello-sdks/node/node_modules/stripe/src/Error.ts:27:12)
    at <anonymous> (/Users/xavdid/stripe/hello-sdks/node/node_modules/stripe/src/RequestSender.ts:194:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
```

after:

```
Error: Invalid API Key provided: sk_test_fake
    at generateV1Error (/Users/xavdid/stripe/hello-sdks/node/node_modules/stripe/src/Error.ts:27:12)
    at <anonymous> (/Users/xavdid/stripe/hello-sdks/node/node_modules/stripe/src/RequestSender.ts:194:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
Originating from:
    at CustomerResource._makeRequest (/Users/xavdid/stripe/hello-sdks/node/node_modules/stripe/src/StripeResource.ts:97:27)
    at CustomerResource.list (/Users/xavdid/stripe/hello-sdks/node/node_modules/stripe/src/resources/Customers.ts:207:17)
    at listCustomers (/Users/xavdid/stripe/hello-sdks/node/index.ts:9:52)
    at <anonymous> (/Users/xavdid/stripe/hello-sdks/node/index.ts:20:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

### What?
 
- manually capture tracebacks before performing async operations. Then, bind that stack to the error if one is thrown (in 3 places)
- tests

### See Also

- https://github.com/stripe/stripe-node/issues/1542
- https://github.com/stripe/stripe-node/issues/1707
- https://github.com/stripe/stripe-node/issues/1066
- [DEVSDK-1103](https://go/j/DEVSDK-1103)


src/autoPagination.ts
```diff
@@ -416,6 +427,7 @@ function makeAutoPagingToArray<T>(
     opts,
     onDone: IterationDoneCallback
   ): Promise<Array<any>> {
+    const callSiteStack = new Error().stack;
```
jar-stripe, line 428: its a small nit, but consider adding a util for "getCurrentCallStack" to move this and the same above out of the business logic

src/autoPagination.ts
```diff
@@ -398,7 +404,12 @@ function makeAutoPagingEach<T>(
       asyncIteratorNext,
       // @ts-ignore we might need a null check
       onItem
-    );
+    ).catch((err) => {
+      if (callSiteStack) {
```
jar-stripe: is there a chance it can be null or undefined?  this is a small stylistic nit, but attachCallSiteToError could take a null | undefined callSiteStack arg and do this filtering there, to keep this part clean (i.e. without the conditional)

src/utils.ts
```diff
@@ -489,3 +489,23 @@ export function parseHeadersForFetch(
...
 }
+
+const CALL_SITE_MARKER = '\nOriginating from:';
+
+/**
+ * We do some async indirection in our HTTP calling code that means stack traces aren't correct traced back to their actual origin in userland
+ * So, we call this with a manually sourced error right before we do async HTTP operations, capturing the stack trace
+ * NOTE: Modifies the `err` arg.
+ */
+export function attachCallSiteToError(err: Error, callSiteStack: string): void {
```
jar-stripe: is there any way to do this without modifying the Error?  could we create a new Error and just transfer everything over?  My concern is if some Error derived class is incompatible with this logic for some reason, or some execution environment doesn't provide mutable errors for some reason.  Its a small concern but a new object might be safer.

jar-stripe (changes requested):
We should have a utility for getting the stack, given there are a few places use that.  That's the biggest requested change, the others would be nice to consider/test for.

src/autoPagination.ts
xavdid: So funny story: this is what I had originally! But then `getCurrentCallStack` is included in the generated stack trace, which reads a little oddly. We could modify the trace to exclude it, but that felt kludgy. Given that the implementation is just a single line, it felt small enough to just repeat.

src/autoPagination.ts
xavdid: > is there a chance it can be null or undefined?

according to the types, yeah. Fair point, will update!

src/autoPagination.ts
jar-stripe: ah thats fair, and lopping off a piece of the stack is usually how I handle that 😅 .  ok, im ok with this

src/utils.ts
xavdid: We could copy the old stack to a new one, but i'm not sure how that would affect the actual stack trace if we need to track down where in the SDK it _did_ actually come from.

Given that we guard against `err.stack` not being present I think worst case this will do nothing but won't fail any worse than today (I'd be really surprised about an immutable property on a user-defined error). And, if we go through and async everything, then we'll pull this code. So IMO we ship the simple thing and fix it for real when we can.

jar-stripe (approved):

PR #2619, opened 2026-03-23
Improved TypeScript support in the Node SDK

prathmesh-stripe:
We introduced inline TypeScript types and removed the manually maintained types folder. 

Read [sneak peak announcement](https://github.com/stripe/stripe-node/discussions/2575)
[Migration guide](https://github.com/stripe/stripe-node/wiki/Migration-guide-for-v22)

PR #2704, opened 2026-05-05
Fix file upload regression: restore multipart request data processor

mbroshi-stripe:
### Why?

File uploads to `/v1/files` broke in v22 (works in v21). When `FileResource` was
converted from `StripeResource.extend({...})` to a class-based pattern in #2619,
the `requestDataProcessor = multipartRequestDataProcessor` assignment was dropped.

Without it, file upload POST bodies go through `queryStringifyRequestData()` instead
of `multipartDataGenerator()`, producing malformed request bodies that the Stripe API
rejects with 400 "Invalid request (check your POST parameters)".

### What?

- Restored `requestDataProcessor = multipartRequestDataProcessor` on `FileResource`
- Added regression test verifying file uploads produce multipart-encoded bodies
  (a `Uint8Array` containing proper `Content-Disposition: form-data` headers)

### See Also
- Introduced in: #2619 (commit 8d8bd302)

## Changelog

- Fixed file uploads failing with `StripeInvalidRequestError` on the `/v1/files` endpoint since v22.

### DIFF
--- a/test/resources/Files.spec.js
+++ b/test/resources/Files.spec.js
@@ -47,6 +47,29 @@ describe('Files Resource', () => {
   });
 
   describe('create', () => {
+    it('Encodes file upload as multipart/form-data', () => {
+      const testFilename = path.join(__dirname, 'data/minimal.pdf');
+      const f = fs.readFileSync(testFilename);
+
+      return stripe.files
+        .create({
+          purpose: 'dispute_evidence',
+          file: {
+            data: f,
+            name: 'minimal.pdf',
+            type: 'application/octet-stream',
+          },
+        })
+        .then(() => {
+          const lastData = stripe.REQUESTS[stripe.REQUESTS.length - 1];
+          expect(lastData).to.be.an.instanceOf(Uint8Array);
+          const asString = new TextDecoder().decode(lastData);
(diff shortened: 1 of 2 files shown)


test/resources/Files.spec.js
```diff
@@ -47,6 +47,29 @@ describe('Files Resource', () => {
...
+          file: {
+            data: f,
+            name: 'minimal.pdf',
+            type: 'application/octet-stream',
+          },
+        })
+        .then(() => {
+          const lastData = stripe.REQUESTS[stripe.REQUESTS.length - 1];
+          expect(lastData).to.be.an.instanceOf(Uint8Array);
+          const asString = new TextDecoder().decode(lastData);
```
jar-stripe: only thing to check here; is TextDecoder right?  I want to say this was the source of an implementation problem and we had to switch to something else.

jar-stripe (approved):
LGTM; please double check that the test is using the correct class before merging

test/resources/Files.spec.js
mbroshi-stripe: It is right. We use `TextEncoder` in our production code. It doesn't really matter since this test just uses ASCII, but I updated to `TextDecoder('utf8')` to be more explicit on the encoding.

PR #2711, opened 2026-05-12
Export manually maintained types for OAuthToken and OAuth param classes

prathmesh-stripe:
### Why

The OAuth-related types (`OAuthToken`, `OAuthTokenParams`, `OAuthAuthorizeUrlParams`, `OAuthAuthorizeUrlOptions`, `OAuthDeauthorization`, `OAuthDeauthorizeParams`) were defined as module-private interfaces in `OAuth.ts` and not re-exported from the `Stripe` namespace. This meant users who wanted to type-annotate variables or function parameters with these types had no way to reference them — they had to inline the shapes or use `any`.

### What
- Export all OAuth interfaces and the `OAuthAuthorizeUrlParams` namespace from `src/resources/OAuth.ts`
- Re-export them from the `Stripe` namespace in `stripe.core.ts` and `stripe.esm.node.ts`
- Add type-compilation tests in `testProjects/types/typescriptTest.ts` to verify the exports are usable

## Changelog
```ts
// ❌ Before (v22.0.0-v22.1.1)
const oAuthToken: Stripe.OAuthToken = {};

// ✅ After (works on v22.1.2+)
const oAuthToken: Stripe.OAuthToken = {};
```


xavdid (approved):
nice! great tests & example

PR #2712, opened 2026-05-12
Exported types for nested param classes

prathmesh-stripe:
### Why?
When param types are re-exported using `export type X = Namespace.X`, only the type itself is exposed — any nested namespaces within that type (e.g. `AccountCreateParams.Configuration`) are lost. This means users can't reference nested param interfaces like `Stripe.V2.Core.AccountCreateParams.Configuration` even though the top-level `AccountCreateParams` is accessible.

### What?
- Changed all `export type X = Namespace.X` declarations in resource index files to `export import X = Namespace.X`, which re-exports the full namespace (including nested interfaces/namespaces) rather than just the type alias
- Added a type test verifying `Stripe.V2.Core.AccountCreateParams.Configuration` is accessible which wasn't accessible before this.

## Changelog
* Types inside nested param class namespaces(e.g. `Stripe.V2.Core.AccountCreateParams.Configuration`, `Stripe.V2.Core.AccountCreateParams.Dashboard`) are now accessible.


xavdid (approved):
nice!

PR #2715, opened 2026-05-13
Add WinterTC compatibility ESLint rule

jar-stripe:
### Why?

stripe-node must run across Node.js, browsers, Deno, Bun, and Cloudflare Workers. Today there's no automated way to catch Node.js-specific API usage leaking into shared code. This adds a lint rule based on the [WinterTC Minimum Common Web Platform API spec](https://min-common-api.proposal.wintertc.org/) (ECMA-429) to flag these issues at development time.

No existing ESLint plugin covers this spec — we searched npm, GitHub, and the WinterTC55 org. This is a custom rule.

### What?

**New ESLint rule** (`eslint-rules/wintertc-compat.js`):
- Flags imports from Node.js built-in modules (`http`, `crypto`, `events`, etc.)
- Flags Node.js-specific globals (`Buffer`, `process`, `__dirname`, `__filename`, `global`, `require`, `module`, `exports`)
- Flags `NodeJS.*` type references (`NodeJS.Timeout`, `NodeJS.ReadableStream`, etc.)
- Enabled for `src/**/*.ts`, disabled for Node-specific platform files (`Node*.ts`, `stripe.*.node.ts`)
- Added `--rulesdir eslint-rules` to the lint command in the Justfile

**Fixes (non-breaking):**
- Replace `NodeJS.Timeout` with `ReturnType<typeof setTimeout>` (portable across runtimes)
- Remove `Buffer` from `FileData` union type (`Buffer extends Uint8Array`, so callers are unaffected)
- Move `process`-dependent code (`emitWarning`, `detectAIAgent`, `determineProcessUserAgentProperties`) into platform functions via `Stripe.initialize()`. Also fixes an existing bug where `emitWarning` would crash on non-Node runtimes.
- Add `emitWarning()`, `getEnv()`, `getRuntimeVersion()` to `PlatformFunctions` base class with cross-runtime defaults

**Deferred to next major (public API types):**
- `NodeJS.ReadableStream` in `StripeStreamResponse` and `HttpClient` (DEVSDK-3112)
- `EventEmitter` in `StreamingFile` and `PlatformFunctions` (DEVSDK-3113)
- `http.Agent` in `StripeConfig` (DEVSDK-3114)

These are suppressed with `eslint-disable` + TODO comments referencing the tickets.

**Other:**
- Downgrade `no-warning-comments` from `error` to `warn` so TODO comments don't block CI
- DEVSDK-3111: Follow-up to implement `getEnv()`/`getRuntimeVersion()` for Deno, Bun, and Workers

### See Also

- WinterTC Minimum Common Web Platform API spec: https://min-common-api.proposal.wintertc.org/
- ECMA TC55 (WinterTC): https://github.com/WinterTC55

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -1,4 +1,4 @@
-// TODO(DEVSDK-3114): Remove http.Agent from shared types in next major version.
+// DEVSDK-3114: Remove http.Agent from shared types in next major version.
 // eslint-disable-next-line wintertc-compat
 import {Agent} from 'http';
 
@@ -281,7 +281,7 @@ export interface ApiSearchResultPromise<T>
   autoPagingToArray(opts: {limit: number}): Promise<Array<T>>;
 }
 
-// TODO(DEVSDK-3112): Replace with WHATWG ReadableStream in next major version.
+// DEVSDK-3112: Replace with WHATWG ReadableStream in next major version.
 // eslint-disable-next-line wintertc-compat
 export type StripeStreamResponse = NodeJS.ReadableStream;
 
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -1,4 +1,4 @@
-// DEVSDK-3114: Remove http.Agent from shared types in next major version.
+// TODO(DEVSDK-3114): Remove http.Agent from shared types in next major version.
 // eslint-disable-next-line wintertc-compat
 import {Agent} from 'http';
 
@@ -281,7 +281,7 @@ export interface ApiSearchResultPromise<T>
   autoPagingToArray(opts: {limit: number}): Promise<Array<T>>;
 }
 
-// DEVSDK-3112: Replace with WHATWG ReadableStream in next major version.
+// TODO(DEVSDK-3112): Replace with WHATWG ReadableStream in next major version.
 // eslint-disable-next-line wintertc-compat
 export type StripeStreamResponse = NodeJS.ReadableStream;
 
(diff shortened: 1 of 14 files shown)


src/lib.ts
```diff
@@ -315,7 +319,8 @@ export interface AppInfo {
 }
 
 export interface FileData {
-  data: string | Buffer | Uint8Array;
+  // Buffer (Node.js) is accepted here because it extends Uint8Array.
+  data: string | Uint8Array;
```
prathmesh-stripe, line 323: Are you sure that this is not a breaking change in all Node versions we support?

prathmesh-stripe (approved):
Just one comment, rest of this look great! Thanks for adding this.

src/lib.ts
jar-stripe: good callout!  I am confident its good based on https://nodejs.org/docs/latest-v20.x/api/buffer.html#buffer but i'll have claude run a manual test before merging.

jar-stripe: Confirmed this is safe across all supported Node versions (>=18).

`Buffer extends Uint8Array` in `@types/node` (and at runtime) since Node 4.x. Verified by compiling a test under `--strict` that assigns `Buffer.from(...)` to `data: string | Uint8Array` — zero errors. Existing user code that passes a `Buffer` continues to type-check because `Buffer` is a subtype of `Uint8Array`.

PR #2717, opened 2026-05-15
Add V2List and V2ListPromise types for v2 list endpoints

jar-stripe:
### Why?

V2 list API responses have a different shape than v1 lists — they return `{data, next_page_url, previous_page_url}` instead of `{object, data, has_more, url}`. The SDK was using `ApiList<T>` (the v1 type) for both, meaning users couldn't access `next_page_url` or `previous_page_url` without type assertions.

### What?

- Adds `V2List<T>` interface with `data`, `next_page_url`, and `previous_page_url` fields
- Adds `V2ListPromise<T>` promise wrapper with auto-pagination methods, parallel to `ApiListPromise<T>`
- Exports both types from the `Stripe` namespace
- Regenerates V2 resource files to return `V2ListPromise` instead of `ApiListPromise`
- Adds tests verifying the v2 list response shape

#### Note to the reviewer
To avoid introducing a breaking change, `V2List` temporarily includes the deprecated properties from `ApiList` (`object`, `has_more`, `url`).  This means existing code that expects an `ApiList` from v2 list endpoints will continue to compile without errors.  These properties are not populated at runtime for v2 lists (this was already the case before this PR) and should be removed at the next major release.

### See Also

N/A

## Changelog
* Adds `V2List` and `V2ListPromise` types to represent API v2 list responses.  These types include deprecated fields from the previously returned list shapes.  These fields will be removed in the next major release.


mbroshi-stripe (commented):
Looks good to me, but adding @prathmesh-stripe in case there is some export edge case I'm not aware of (I know we've seen issues with, e.g., `Stripe.Decimal` recently).

src/lib.ts
```diff
@@ -235,6 +235,39 @@ export interface ApiListPromise<T>
...
+export interface V2ListPromise<T>
+  extends Promise<Response<V2List<T>>>,
+    AsyncIterableIterator<T> {
+  autoPagingEach(
+    handler: (item: T) => boolean | void | Promise<boolean | void>,
+    onDone?: (err: any) => void
+  ): Promise<void>;
+
+  autoPagingToArray(
+    opts: {limit: number},
```
xavdid, line 295: I believe `opts` should be optional? it'll pick up a default limit if they don't specify. `limit` should be optional too probably. it doesn't make sense to not include in the object now, but we ever add a second option, `limit` would need to be optional

src/stripe.core.ts
```diff
@@ -28,6 +28,8 @@ import {
   RawRequestOptions,
   ApiList,
   ApiListPromise,
+  V2List,
```
xavdid, line 31: I believe we need these in all the `stripe.(esm|cjs).*.ts` files as well

xavdid (commented):
I'll leave to prathmesh for final approval since he has the most context. Approach looks good though.

One thing to add: a block in `typescriptTest` that verifies the _types_ are correct. I believe these runtime tests would pass today, since the API shape hasn't changed.

Lastly, this is definitely a breaking change (and the PR title should reflect that). If we're settled on doing an out-of-band major release for it, we should write the migration docs and things.

src/lib.ts
jar-stripe: The existing `ApiListPromise` doesn't define opts as optional:
```
  autoPagingToArray(
    opts: {limit: number},
    onDone?: (err: any) => void
  ): Promise<Array<T>>;
```

Do we want to change that for one or both in this PR?

src/stripe.core.ts
jar-stripe: Added to esm, but cjs doesn't have the same import/export pattern currently.

prathmesh-stripe: @jar-stripe I am working on a PR for that. I'll add this change to mine and both of them will be released together. Please make sure you merge this PR after the private-preview releases today.

jar-stripe: This PR is required for the private preview releases today, I think?

src/lib.ts
xavdid: might as well? it's backwards compatible, but we'll have to check that we're picking up the default at runtime if it's suddenly undefined.

src/lib.ts
```diff
@@ -235,6 +235,68 @@ export interface ApiListPromise<T>
...
+
+  /**
+   * TODO(DEVSDK-2534): remove these properties in our next major release.
+   * these deprecated properties were copied from ApiList<T> to not break
+   * existing code.  these properties will continue to be not populated at
+   * runtime.
+   */
+
+  /**
+   * @deprecated This property is not populated at runtime for v2 lists
```
xavdid, line 268: nit: mention that these are just here for typescript compatibility (the current comment is accurate, but could use context)

xavdid (approved):
approving to unblock- I think the shim will work.

Still should add the typescript test mentioned above- we rely on those for regression checking!

src/lib.ts
jar-stripe: Yea, lets file an issue for this and pick it up separately.

test/resources/V2/Core/Events.spec.js
```diff
@@ -123,6 +123,49 @@ describe('V2 Core Events Resource', () => {
...
+          method: 'GET',
+          path: '/v2/core/events?object_id=foo',
+          response: `{
+            "data": [${v2EventPayloadWithoutRelatedObject}],
+            "next_page_url": "/v2/core/events?object_id=foo&page=next",
+            "previous_page_url": "/v2/core/events?object_id=foo&page=prev"
+          }`,
+        },
+      ]);
+      const resp = await mockStripe.v2.core.events.list({object_id: 'foo'});
```
prathmesh-stripe, line 138: What is this typed as? You would also have to make changes where the list method is defined.

src/lib.ts
jar-stripe: added in https://github.com/stripe/stripe-node/pull/2717/commits/421570d2ab464f7837064a484dd3e84ac6a253fc

test/resources/V2/Core/Events.spec.js
jar-stripe: That method is generated, so I was planning on either following up this PR or letting our auto generated PRs take care of it.  Want me to push generated code to this PR instead?

jar-stripe: Nvm, I had to in order to get the tests to pass.

prathmesh-stripe (approved):

PR #2723, opened 2026-05-21
Emit warning when `stripe-notify` header is present in response

xavdid:
### Why?

We've noticed a trend where users (and their agents) will sometimes reach for suboptimal API endpoints/methods because of outdated information. For example, using `/v1/accounts` when `/v2/core/accounts` is available and has a superset of the functionality from v1.

In an effort to help steer users towards best practices, we're introducing a new header to provide feedback during the development process. The SDK is responsible for surfacing this header, if present.

Once enabled internally, the header will be returned by the server for:

- all testmode calls
- livemode calls made by an agent

We may tweak that going forwards. But no matter the conditions, the SDKs will act as a thin pipe.

### What?

- Emit warning when we see the `Stripe-Notice` header in a response
- add tests

### See Also

- https://go/agent-steering
- [DEVSDK-2430](https://go/j/browse/DEVSDK-2430)

### DIFF
--- a/src/RequestSender.ts
+++ b/src/RequestSender.ts
@@ -9,6 +9,7 @@ import {
 import {StripeContext} from './StripeContext.js';
 import {
   RequestHeaders,
+  ResponseHeaders,
   RequestEvent,
   ResponseEvent,
   RequestCallback,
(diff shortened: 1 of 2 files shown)


src/RequestSender.ts
```diff
@@ -96,6 +97,15 @@ export class RequestSender {
     return headers['request-id'] as string;
   }
 
+  private _emitStripeNotice(headers: ResponseHeaders): void {
+    const notice = headers['stripe-notice'];
```
jar-stripe, line 101: just a sanity check; all header keys are lower case?

jar-stripe (approved):

src/RequestSender.ts
xavdid: Yep, node lowercases everything interally before we access headers. 

I'll tweak the test to send a capitalized case header (since that's probably how the server sends it) but the implementation should be correct

PR #2733, opened 2026-06-01
fix: add concurrency guard, fix empty-page truncation in V2ListIterator, and harden parseHttpHeaderAsNumber

### LINKED ISSUE #2730: V2ListIterator.next() has no concurrency guard, causing duplicate page fetches on parallel calls
### Describe the bug

items in the result set.

    }
    ...
  }

### To Reproduce

### Expected behavior

exactly as V1Iterator does.

### Code snippets

```JavaScript

```

### OS

### Node version

### Library version

### API version

### Additional context

### LINKED ISSUE #2731: V2ListIterator silently terminates iteration when an intermediate page has an empty data array
### Describe the bug

After V2ListIterator exhausts the current page and calls turnPage(), it 
calls .next() on the returned iterator exactly once. If the API returns 
a page with data: [] but a non-null next_page_url — which can occur during 
server-side cursor races — the single .next() returns {done: true} and the 
method immediately returns {done: true, value: undefined}, silently dropping 
all items on every subsequent page.

Current code:
  const nextPageIterator = await this.turnPage();
  if (!nextPageIterator) return {done: true, value: undefined};
  const result = nextPageIterator.next();
  if (!result.done) return {done: false, value: result.value};
  return {done: true, value: undefined};  // gives up if page was empty

### To Reproduce

Use a V2 list endpoint. Simulate or encounter a server response where an 
intermediate page returns data: [] with next_page_url pointing to a page 
that does contain items. Iteration will stop at the empty page and never 
reach the items.

### Expected behavior

If a page has data: [] but next_page_url is set, the iterator should 
continue fetching pages until it finds an item or next_page_url is null.

### Code snippets

```JavaScript

```

### OS

Not environment-specific — reproduced by code inspection of src/autoPagination.ts. The missing while loop in V2ListIterator.next() will truncate iteration on empty intermediate pages regardless of operating system.

### Node version

Reproducible on any supported Node version. Verified against Node 20 LTS. The bug is structural — after turnPage() returns, .next() is called exactly once on the new iterator. If that page is empty, iteration terminates regardless of Node version or runtime environment.

### Library version

stripe-node master branch — confirmed in src/autoPagination.ts V2ListIterator.next() method (current master HEAD).

### API version

Affects V2 API endpoints only (paths beginning with /v2/). Triggered when the server returns an intermediate page with data: [] and a non-null next_page_url — possible under cursor-race conditions during high-concurrency V2 list requests.  Any API version using V2 list endpoints (2025-03-31.basil or later).

### Additional context

V1Iterator does not have this problem because it recurses through iterate() 
which checks has_more in a loop. V2ListIterator needs the equivalent 
while loop pattern.

Related to the concurrency issue filed separately. Both are in V2ListIterator 
and can be fixed in a single PR.

### LINKED ISSUE #2732: parseHttpHeaderAsNumber returns NaN instead of null for absent or non-numeric headers
### Describe the bug

parseHttpHeaderAsNumber in src/utils.ts returns NaN when the header is 
absent, undefined, or contains a non-numeric string. The return type is 
declared as number, but NaN is not a valid number in any meaningful sense 
and violates the type contract for callers.

Current implementation:

  export function parseHttpHeaderAsNumber(header): number {
    const number = Array.isArray(header) ? header[0] : header;
    return Number(number);
  }

Number(undefined) === NaN
Number(null)      === NaN  
Number('')        === NaN
Number('bad')     === NaN

The primary call site is in RequestSender.ts where retry-after header 
is parsed:

  parseHttpHeaderAsNumber(res.getHeaders()['retry-after'])

When retry-after is absent — which is the common case — this returns NaN. 
The value is passed as the retryAfter argument to _getSleepTimeInMS where 
Number.isInteger(NaN) === false accidentally prevents incorrect behaviour. 
However NaN silently propagates as a number-typed value and any future 
caller that does arithmetic on the return value directly will produce NaN 
throughout their calculation without any indication of what went wrong.

### To Reproduce

import { parseHttpHeaderAsNumber } from './src/utils.ts';

console.log(parseHttpHeaderAsNumber(undefined));  // NaN — should be null
console.log(parseHttpHeaderAsNumber(null));        // NaN — should be null
console.log(parseHttpHeaderAsNumber(''));          // NaN — should be null
console.log(parseHttpHeaderAsNumber('bad'));       // NaN — should be null
console.log(parseHttpHeaderAsNumber('30'));        // 30  — correct

### Expected behavior

parseHttpHeaderAsNumber should return null for absent, undefined, empty, 
or non-numeric header values, and return a number only when the header 
contains a valid numeric string. The return type should be number | null 
to accurately reflect this contract.

Suggested fix:

  export function parseHttpHeaderAsNumber(header): number | null {
    if (header === undefined || header === null) return null;
    const raw = Array.isArray(header) ? header[0] : header;
    const parsed = Number(raw);
    return Number.isFinite(parsed) && String(raw).trim() !== '' 
      ? parsed 
      : null;
  }

### Code snippets

```JavaScript

```

### OS

Not environment-specific — reproduced by code inspection of src/utils.ts.  Number(undefined) === NaN is consistent across all operating systems and  JavaScript runtimes.

### Node version

Reproducible on any supported Node version. Verified against Node 20 LTS.  This is a language-level behaviour of the Number() constructor, not a  Node version-specific issue.

### Library version

stripe-node master branch — confirmed in src/utils.ts  parseHttpHeaderAsNumber function (current master HEAD). Primary call site: src/RequestSender.ts line 687.

### API version

Not API version specific. parseHttpHeaderAsNumber is used to parse the  retry-after response header in RequestSender.ts and applies to all API  versions and all endpoints. The missing header scenario occurs on any  response that does not include a retry-after header, which is the  majority of responses.

### Additional context

The current implementation is saved from user-visible consequences by a 
coincidental guard in _getSleepTimeInMS:

  if (Number.isInteger(retryAfter) && retryAfter! <= MAX_RETRY_AFTER_WAIT)

Number.isInteger(NaN) === false, so the NaN value is ignored and the 
sleep calculation proceeds correctly. This means there is no current 
user-visible bug — however the type contract is broken and the safety 
is accidental rather than intentional.

Any future refactor that reads the return value of parseHttpHeaderAsNumber 
directly without the isInteger guard will produce silent NaN propagation.

This fix is related to two concurrency and pagination issues filed 
separately against V2ListIterator (#2730, #2731). All three will be 
addressed in a single PR since they were discovered together during 
the same code review session.

m-zandi:
## Summary

Fixes three related bugs found during code review of `src/autoPagination.ts`
and `src/utils.ts`. Two bugs are in `V2ListIterator`, one is in a shared
utility function.

---

## Bug 1 — V2ListIterator missing concurrency guard

Closes #2730

`V2ListIterator.next()` had no `promiseCache` deduplication guard, unlike
`V1Iterator` which has had this since the original pagination implementation.

When `.next()` is called in parallel while a page transition is in flight,
both calls independently reach `turnPage()` and issue duplicate HTTP
requests for the same page, producing duplicated or out-of-order items.

**Fix:** Added `private promiseCache: PromiseCache` and the same
deduplication wrapper used by `V1Iterator`. Extracted private `_next()`
method to keep the concurrency wrapper clean.

---

## Bug 2 — V2ListIterator silently truncates on empty intermediate pages

Closes #2731

After `turnPage()`, the old code called `.next()` on the new iterator
exactly once. If the API returned `data: []` with `next_page_url` set,
that single `.next()` returned `{done: true}` and the method immediately
returned `{done: true}`, silently dropping all items on subsequent pages.

**Fix:** The new private `_next()` method uses a `while` loop that keeps
fetching pages until it finds an item or exhausts `next_page_url` —
matching the pattern `V1Iterator` uses through its `iterate()` recursion.

---

## Bug 3 — parseHttpHeaderAsNumber returns NaN for absent headers

Closes #2732

`Number(undefined) === NaN`. When `retry-after` was absent,
`parseHttpHeaderAsNumber` returned `NaN` typed as `number`, violating
the return type contract. `Number.isInteger(NaN) === false` accidentally
prevented incorrect sleep behaviour, but the safety was coincidental
rather than intentional.

**Fix:** Return type changed to `number | null`. Explicit guards added
for `undefined`, `null`, empty strings, and non-numeric values.

---

## Why these are in one PR

Bugs 1 and 2 are fixed in the same method (`V2ListIterator._next()`).
The concurrency guard requires extracting `_next()`, and the while loop
lives inside that same method. Separating them would produce two PRs
modifying identical lines.

Bug 3 is two lines in `utils.ts`, found during the same investigation.
Too small for a separate PR but warranted its own issue for traceability.

---

## Files changed

### DIFF
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -176,6 +176,10 @@ class V2ListIterator<T> implements AsyncIterator<T> {
   private options: RequestOptions | undefined;
   private spec: MakeRequestSpec | undefined;
   private stripeResource: StripeResourceObject;
+  // Concurrency guard: mirrors V1Iterator.promiseCache to prevent duplicate
+  // page-fetch requests when .next() is called in parallel.
+  private promiseCache: PromiseCache;
+
   constructor(
     firstPagePromise: Promise<PageResult<T>>,
     options: RequestOptions | undefined,
@@ -188,7 +192,9 @@ class V2ListIterator<T> implements AsyncIterator<T> {
     this.options = options;
     this.spec = spec;
     this.stripeResource = stripeResource;
+    this.promiseCache = {currentPromise: null};
   }
+
   private async initFirstPage(): Promise<void> {
     if (this.firstPagePromise) {
       const page = await this.firstPagePromise;
@@ -197,6 +203,7 @@ class V2ListIterator<T> implements AsyncIterator<T> {
       this.nextPageUrl = page.next_page_url || null;
     }
   }
+
   private async turnPage(): Promise<Iterator<T> | null> {
     if (!this.nextPageUrl) return null;
     const page = await this.stripeResource._makeRequest(
@@ -210,19 +217,43 @@ class V2ListIterator<T> implements AsyncIterator<T> {
     this.currentPageIterator = page.data[Symbol.iterator]();
     return this.currentPageIterator;
   }
-  async next(): Promise<IteratorResult<T>> {
+
+  private async _next(): Promise<IteratorResult<T>> {
     await this.initFirstPage();
-    if (this.currentPageIterator) {
-      const result = this.currentPageIterator.next();
-      if (!result.done) return {done: false, value: result.value};
+
+    // Walk through pages until we find an item or exhaust all pages.
+    // Previously this only called turnPage() once, silently terminating
+    // iteration if an intermediate page happened to be empty.
+    while (true) {
+      if (this.currentPageIterator) {
+        const result = this.currentPageIterator.next();
+        if (!result.done) return {done: false, value: result.value};
+      }
+      const nextPageIterator = await this.turnPage();
+      if (!nextPageIterator) {
+        return {done: true, value: undefined};
+      }
     }
-    const nextPageIterator = await this.turnPage();
-    if (!nextPageIterator) {
-      return {done: true, value: undefined};
+  }
+
+  next(): Promise<IteratorResult<T>> {
+    /**
+     * If a user calls .next() multiple times in parallel,
+     * return the same result until something has resolved
+     * to prevent page-turning race conditions.
+     */
+    if (this.promiseCache.currentPromise) {
+      return this.promiseCache.currentPromise;
     }
-    const result = nextPageIterator.next();
-    if (!result.done) return {done: false, value: result.value};
-    return {done: true, value: undefined};
+
+    const nextPromise = (async (): Promise<IteratorResult<T>> => {
+      const ret = await this._next();
+      this.promiseCache.currentPromise = null;
+      return ret;
+    })();
+
+    this.promiseCache.currentPromise = nextPromise;
+    return nextPromise;
   }
 }
 
--- a/test/bugfixes.spec.ts
+++ b/test/bugfixes.spec.ts
@@ -1,0 +1,174 @@
+// @ts-nocheck
+/**
+ * Tests for three bugs fixed in this PR:
+ *
+ *  Bug 1 – V2ListIterator: no concurrency guard; parallel .next() calls
+ *           could issue duplicate page-fetch requests.
+ *           Closes #2730
+ *
+ *  Bug 2 – V2ListIterator: silent truncation when an intermediate page
+ *           has an empty data array.
+ *           Closes #2731
+ *
+ *  Bug 3 – parseHttpHeaderAsNumber: returned NaN for absent or
+ *           non-numeric headers instead of null.
+ *           Closes #2732
+ */
+
+import {expect} from 'chai';
+import {makeAutoPaginationMethods} from '../src/autoPagination.js';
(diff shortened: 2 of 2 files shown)


jar-stripe:
Hi @m-zandi thank you for the issues and for the PRs!  Unfortunately, we cannot merge this one as is; it has lint errors and the bugfixes test file does not slot nicely into the repo structure; we also use a PR template to keep our pull request descriptions consistent.

I went ahead and fixed Bug 3 in a separate PR https://github.com/stripe/stripe-node/pull/2734; I invite you to refactor your other two fixes on this PR into PRs that match our file structure and PR description template.  I will look to close these out this week though, so if you would like to do this please do so in the next few days.

Thanks!

jar-stripe:
I have not heard back so I am closing this PR.  We invite you to submit a new PR once you familiarize yourself with the contributing guidelines of this repo and sign the CLA.  Thank you!

PR #2735, opened 2026-06-02
Add promiseCache concurrency guard to V2ListIterator.next()

### LINKED ISSUE #2730: V2ListIterator.next() has no concurrency guard, causing duplicate page fetches on parallel calls

theherrovn-sys:
Port the `promiseCache` deduplication pattern from `V1Iterator` to `V2ListIterator` to prevent duplicate HTTP requests when `.next()` is called in parallel while a page transition is in flight.

## Problem

`V2ListIterator.next()` had no concurrency guard. If two pieces of code called `.next()` simultaneously while a page was in flight — which can happen inside `autoPagingEach` under certain async patterns — both calls independently reached `turnPage()`, issuing two HTTP requests for the same next page.

## Fix

This ports the exact same `promiseCache` pattern used in `V1Iterator` (lines ~111–130 of `src/autoPagination.ts`):

- Added a `promiseCache: PromiseCache` property to `V2ListIterator`
- Renamed the core logic to a private `_next()` method
- Wrapped `_next()` in a new `next()` that checks `promiseCache.currentPromise` before proceeding, coalescing parallel calls into a single in-flight promise

## Changes

- `src/autoPagination.ts`: Added `promiseCache` property, constructor init, and guarded `next()` + private `_next()` methods in `V2ListIterator`

Fixes #2730

### DIFF
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -176,6 +176,7 @@ class V2ListIterator<T> implements AsyncIterator<T> {
   private options: RequestOptions | undefined;
   private spec: MakeRequestSpec | undefined;
   private stripeResource: StripeResourceObject;
+  private promiseCache: PromiseCache;
   constructor(
     firstPagePromise: Promise<PageResult<T>>,
     options: RequestOptions | undefined,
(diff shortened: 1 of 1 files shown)


theherrovn-sys:
@cla-assistant check

PR #2749, opened 2026-06-09
fix: add concurrency guard and fix empty-page truncation in V2ListIterator

### LINKED ISSUE #2730: V2ListIterator.next() has no concurrency guard, causing duplicate page fetches on parallel calls

### LINKED ISSUE #2731: V2ListIterator silently terminates iteration when an intermediate page has an empty data array

m-zandi:
### Why?

`V2ListIterator` in `src/autoPagination.ts` was added without two 
robustness patterns that `V1Iterator` has carried since the original 
pagination implementation.

First, without a `promiseCache` concurrency guard, parallel `.next()` 
calls race to call `turnPage()` independently, issuing duplicate HTTP 
requests for the same page and producing duplicated or out-of-order 
items (#2730).

Second, after `turnPage()` returns, `.next()` was called exactly once 
on the new iterator. If the API returned `data: []` with `next_page_url` 
set, that single call returned `{done: true}` and iteration terminated 
silently, dropping all items on subsequent pages (#2731).

### What?

- Added `private promiseCache: PromiseCache` to `V2ListIterator` matching 
  the deduplication pattern in `V1Iterator`
- Extracted `private _next()` method containing a `while` loop that 
  continues fetching pages until an item is found or `next_page_url` 
  is exhausted
- Added `next()` concurrency wrapper that coalesces parallel calls into 
  a single in-flight promise
- Added tests for both fixes in `test/autoPagination.spec.ts`

### See Also

Closes #2730
Closes #2731

### DIFF
--- a/src/autoPagination.ts
+++ b/src/autoPagination.ts
@@ -173,10 +173,11 @@ class V2ListIterator<T> implements AsyncIterator<T> {
   private firstPagePromise: Promise<PageResult<T>> | null;
   private currentPageIterator: Iterator<T> | null;
   private nextPageUrl: string | null;
-  private promiseCache: PromiseCache;
   private options: RequestOptions | undefined;
   private spec: MakeRequestSpec | undefined;
   private stripeResource: StripeResourceObject;
+  private promiseCache: PromiseCache;
+
   constructor(
     firstPagePromise: Promise<PageResult<T>>,
     options: RequestOptions | undefined,
@@ -186,11 +187,12 @@ class V2ListIterator<T> implements AsyncIterator<T> {
     this.firstPagePromise = firstPagePromise;
     this.currentPageIterator = null;
     this.nextPageUrl = null;
-    this.promiseCache = {currentPromise: null};
     this.options = options;
     this.spec = spec;
     this.stripeResource = stripeResource;
+    this.promiseCache = {currentPromise: null};
   }
+
   private async initFirstPage(): Promise<void> {
     if (this.firstPagePromise) {
       const page = await this.firstPagePromise;
@@ -199,6 +201,7 @@ class V2ListIterator<T> implements AsyncIterator<T> {
       this.nextPageUrl = page.next_page_url || null;
     }
   }
+
   private async turnPage(): Promise<Iterator<T> | null> {
     if (!this.nextPageUrl) return null;
     const page = await this.stripeResource._makeRequest(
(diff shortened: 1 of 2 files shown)


src/autoPagination.ts
```diff
@@ -173,10 +173,11 @@ class V2ListIterator<T> implements AsyncIterator<T> {
   private firstPagePromise: Promise<PageResult<T>> | null;
   private currentPageIterator: Iterator<T> | null;
   private nextPageUrl: string | null;
-  private promiseCache: PromiseCache;
   private options: RequestOptions | undefined;
   private spec: MakeRequestSpec | undefined;
   private stripeResource: StripeResourceObject;
+  private promiseCache: PromiseCache;
```
jar-stripe, line 179: This feels a bit unnecessary; it was previously defined at line 176, and your PR moves it to 179.  Please lint your submission for changes like this as they can increase the time it takes to review the PR.

src/autoPagination.ts
```diff
@@ -212,48 +215,40 @@ class V2ListIterator<T> implements AsyncIterator<T> {
...
-  private async nextFromNewPage(): Promise<IteratorResult<T>> {
-    const nextPageIterator = await this.turnPage();
-    if (!nextPageIterator) {
-      return {done: true, value: undefined};
+    while (true) {
+      if (this.currentPageIterator) {
+        const result = this.currentPageIterator.next();
+        if (!result.done) return {done: false, value: result.value};
+      }
+      const nextPageIterator = await this.turnPage();
```
jar-stripe, line 226: This will fail our lint rules; we discourage awaits inside while loops.  You can run `just prepare` to lint and format before committing to your branch.

src/autoPagination.ts
```diff
@@ -212,48 +215,40 @@ class V2ListIterator<T> implements AsyncIterator<T> {
...
       return this.promiseCache.currentPromise;
     }
-
     const nextPromise = (async (): Promise<IteratorResult<T>> => {
-      try {
-        return await this._next();
-      } finally {
-        this.promiseCache.currentPromise = null;
-      }
+      const ret = await this._next();
```
jar-stripe, line 238: if `_next` raises an exception, currentPromise will be stuck as a rejected promise.  You should catch the error and clear the promise variable.

jar-stripe:
Hi @m-zandi, thank you for resubmitting this PR but these issues were already addressed in https://github.com/stripe/stripe-node/pull/2747.  I went ahead and gave your PR a review either way; please consider this feedback for your next submission.  Thank you!

PR #2754, opened 2026-06-16
Add PR disclaimer to README

xavdid:
### Why?

We've noticed an uptick of spam and low-effort PRs to our repos recently. The team is currently below expected headcount, so the time we have to take to sort through these (just in case they're opened in good faith) takes away from our ability to perform core job functions.

As a result, we're disabling PRs from new contributors. If there's a change you'd like to see in the repo, please open an issue and we'll discuss! You may still have a chance to work on it, we're just disabling the ability to send us unsolicited changes.

### What?

- add PR disclaimer to readme
- disable PRs from new contributors (done via a GH setting)

### See Also

[RUN_DEVSDK-2488](https://go/j/RUN_DEVSDK-2488)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,19 +51,18 @@ console.log(customer.id);
 ```
 
 Or using CJS:
+
 ```js
 const Stripe = require('stripe');
 const stripeClient = Stripe('sk_test_...');
 
-stripeClient.customers.create({
-  email: 'customer@example.com',
-})
-  .then(customer => console.log(customer.id))
-  .catch(error => console.error(error));
+stripeClient.customers
+  .create({
+    email: 'customer@example.com',
+  })
+  .then((customer) => console.log(customer.id))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,20 +51,19 @@ console.log(customer.id);
...
 ```js
 const Stripe = require('stripe');
 const stripeClient = Stripe('sk_test_...');
 
-stripeClient.customers.create({
-  email: 'customer@example.com',
-})
-  .then(customer => console.log(customer.id))
-  .catch(error => console.error(error));
+stripeClient.customers
```
jar-stripe, line 59: I think the other changes in this file are all format fixes, but it would be worth double checking

jar-stripe (approved):

README.md
xavdid: I confirmed!

Issue #2661, opened 2026-04-06
`instanceof Stripe.ErrorType.StripeError` passes type check but throws at runtime


egorFiNE:
### Describe the bug

Unsure how to catch Stripe errors in `catch(err: any) {... }`. 

### To Reproduce

Before (v21):

```typescript
} catch (e: any) {
  if (e instanceof Stripe.ErrorType && e.statusCode === 404) {
    // ... 
  }
}
```

Now this code triggers a ts error because obviously. I'm using TS 6.0.

### Expected behavior

Not sure what would the correct way of separating Stripe Error instances from others in the update type system.

### Code snippets

```JavaScript

```

### OS

doesn't matter

### Node version

doesn't matter

### Library version

v22.0.0

### API version

2026-03-25.dahlia

### Additional context

With a slim hope that management sees this feedback: 

Stop it, just stop. I feel like every new Stripe SDK release lately brings unwelcome and breaking changes and more useless work that brings zero value to anyone involved except Stripe's developers.

Stripe started as a product for developers to eliminate frustration and reduce friction.

Now Stripe is itself a source of frustration since each new SDK version (since about v17) adds friction.

prathmesh-stripe:
`Stripe.error.ErrorType` is now the class constructor itself, can you try the following: 

```ts
try {
} catch (e: any) {
  if (e instanceof Stripe.error.ErrorType && e.statusCode === 404) {
    // ... 
  }
}
```

egorFiNE:
@prathmesh-stripe 
```
Property 'ErrorType' does not exist on type 'typeof import("...stripe/esm/Error", { with: { "resolution-mode": "import" } })
```

xavdid:
Hey @egorFiNE, sorry for the hassle here. I've been playing around with error handling and casting and wasn't able to reproduce your issue. This is what i've got (using TS 6.0 and `v20.4.1` of the node SDK):

```ts
import { Stripe } from "stripe";

const stripeClient = new Stripe("sk_test_123");

try {
  await stripeClient.customers.list();
} catch (error) {
  // auth error, since it's not a real API key
  console.log((error as Stripe.errors.StripeError).statusCode); // 401
  console.log(error instanceof Stripe.errors.StripeError); // true
}
```

When migrating to v22 (with all the type changes), I needed to swap the way I referred to the error _type_ (not runtime class):

```diff
import { Stripe } from "stripe";

const stripeClient = new Stripe("sk_test_123");

try {
[...]

PR #2758, opened 2026-06-18
Fix `Stripe.ErrorType.StripeError` incorrectly being usable as a runtime class

xavdid:
### Why?

Because of the way our errors were re-exported, Typescript allowed the use of `Stripe.ErrorType` in runtime code:

```ts
// passed `tsc`, but shouldn't:
if (error instanceof Stripe.ErrorType.StripeError) { 
  // ...
}
```

But at runtime, it would blow up:

```
16 |   if (error instanceof Stripe.ErrorType.StripeError) {
                                   ^
TypeError: undefined is not an object (evaluating 'Stripe.ErrorType.StripeError')
```

This was caused by the way we exported `ErrorType`. It was being described as a class by the type system, but was only usable as a type. To fix, we don't export it as a class in the type system anymore. There's no runtime impact of this, it's purely a types change

> Note: I _believe_ we don't need `ErrorType` any more at all. It was added to work around this bug, but given that `.errors` now works the way we expect (both as a type and as a value) we can probably drop it in the next major (filed as [DEVSDK-3165](https://go/j/DEVSDK-3165))

### What?

- explicitly export a namespace of errors instead of the runtime module `_Error`

### See Also

- https://github.com/stripe/stripe-node/issues/2661
- [DEVSDK-3141](https://go/j/DEVSDK-3141)

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -2593,6 +2593,105 @@ export declare namespace Stripe {
 
   export {StripeContext as StripeContextType};
   export {StripeRawError};
-  export import ErrorType = _Error;
+  // ErrorTypeNamespaces: The beginning of the section generated from our OpenAPI spec
+  export namespace ErrorType {
+    export type StripeError = InstanceType<typeof _Error.StripeError>;
+    export type StripeCardError = InstanceType<typeof _Error.StripeCardError>;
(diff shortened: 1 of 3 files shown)


jar-stripe (approved):
Change looks good.  Small bit: the PR description could probably be edited a bit, but otherwise GTG

PR #2761, opened 2026-06-22
Encode URI path params in `accounts.retrieve`

xavdid:
### Why?

Small follow up to https://github.com/stripe/stripe-node/pull/2750 to cover a method we missed. It was generated separately, so we missed it in the initial fix.

### What?

- add `encodeURIComponent` call to `accounts.retrieve()`
- add test

### DIFF
--- a/src/resources/Accounts.ts
+++ b/src/resources/Accounts.ts
@@ -52,7 +52,7 @@ export class AccountResource extends StripeResource {
     if (typeof id === 'string') {
       return this._makeRequest(
         'GET',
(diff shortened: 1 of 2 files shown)


jar-stripe (approved):

xavdid:
The tests themselves are passing, but the types-test step is failing because of an upstream issue in the `@types/node` dependency. 

Given that it's unrelated to this PR, i'm going to merge through it and we can file a run ticket for the actual issue, if it persists

Issue #2683, opened 2026-04-12
v22.0.1: `import type Stripe from 'stripe'` does not expose `Stripe.Customer`


seigo2016:
### Describe the bug

`stripe@22.0.1` appears to fix the CommonJS constructor regression from #2660.

However, there seems to be a separate TypeScript issue specifically with type-only default imports:

```ts
import type Stripe from 'stripe';

let customer: Stripe.Customer; // TS2694: Namespace 'StripeConstructor' has no exported member 'Customer'
```

In contrast, the standard documented usage continues to work.
```ts
import Stripe from 'stripe';

const stripe = new Stripe('your_stripe_secret_key', {
  apiVersion: '2026-03-25.dahlia',
});

async function createCustomer() {
  return await stripe.customers.create({
    email: '',
  });
}
```

So this does not appear to be a general regression in namespaced Stripe types.
It appears to be specifically about `import type Stripe from 'stripe'`.

### To Reproduce

1. Create a TypeScript project with
  - `module: "commonjs"`, `moduleResolution: "node"`, `esModuleInterop: true`
2. Install stripe@22.0.1
3. Use:
```ts
import type Stripe from 'stripe';
let customer: Stripe.Customer;
```
4. Run tsc

### Expected behavior

Either
- `import type Stripe from 'stripe'` should support references like `Stripe.Customer`
- the migration guide / README should document that this pattern is no longer supported and recommend the replacement pattern.

### Code snippets

```JavaScript

```

### OS

WSL2 (Debian) on Windows 11

### Node version

Node v22.22.2

### Library version

stripe@22.0.1

### API version

2026-03-25.dahlia

### Additional context

The v22 migration guide documents breaking changes to exported TypeScript types, but it does not mention import type Stripe from 'stripe' as a changed or unsupported pattern. It documents other typing changes such as:

- removal of the old types/ directory
- `Stripe.StripeContext` -> `StripeContextType`
- changed type usage for `Stripe.errors.StripeError`

- Related reports:
  - #2660 follow-up comment: `import type Stripe from 'stripe'` / `Stripe.Customer`
  - #2662: missing namespaced exports such as `Stripe.Checkout.SessionCreateParams`
- I am not sure whether these are exactly the same root cause, but they appear closely related.

konradliebig:
It seems like there is currently no way to access any types with `Stripe.X` in version 22.

[...]

PR #2765, opened 2026-06-24
fix: CJS companion namespace access for resource types at all depths

jar-stripe:
### Why?

CJS consumers can't access nested type namespaces like `Stripe.Price.Recurring`, `Stripe.Billing.Alert.Status`, or `Stripe.Terminal.Configuration.Tipping`. The CJS declaration file uses flat type aliases inside `declare namespace StripeConstructor`, which drops companion namespaces that ESM carries automatically via `export { X }`.

Additionally, nested resource types (under product namespaces like Billing, Terminal, Apps, etc.) were broken in **both** ESM and CJS because companion namespace blocks were rendered nested under the product namespace instead of at module scope where they merge with the interface.

After this change, all type paths that work in ESM also work in CJS.

### What?

- Resource files: companion namespaces now render at module scope (`export namespace Alert { ... }`) instead of nested under the product namespace (`export namespace Billing { export namespace Alert { ... } }`)
- CJS declaration file: companion namespace blocks now emitted for resource sub-types (e.g. `Price.Recurring`, `Alert.Status`) at all depths
- CJS declaration file: params companion blocks now emitted at full depth (previously only 1 level)
- Internal references in resource files updated (e.g. `Billing.Alert.Status` → `Alert.Status`)

### See Also

- #2683
## Changelog

- Fixes TypeScript type access for nested namespaces in CJS mode (e.g. `Stripe.Price.Recurring`, `Stripe.Subscription.BillingMode`)
- Fixes TypeScript type access for nested resource types e.g. `Stripe.Billing.Alert.Status`, `Stripe.Terminal.Configuration.Tipping` in both ESM and CJS modes.

### DIFF
--- a/src/resources/Checkout/Sessions.ts
+++ b/src/resources/Checkout/Sessions.ts
@@ -653,12 +653,12 @@ export interface Session {
   /**
    * Settings for price localization with [Adaptive Pricing](https://docs.stripe.com/payments/checkout/adaptive-pricing).
    */
-  adaptive_pricing: Checkout.Session.AdaptivePricing | null;
+  adaptive_pricing: Session.AdaptivePricing | null;
 
   /**
    * When set, provides configuration for actions to take if this Checkout Session expires.
    */
-  after_expiration: Checkout.Session.AfterExpiration | null;
+  after_expiration: Session.AfterExpiration | null;
 
   /**
    * Enables user redeemable promotion codes.
@@ -675,14 +675,14 @@ export interface Session {
    */
   amount_total: number | null;
 
-  automatic_tax: Checkout.Session.AutomaticTax;
+  automatic_tax: Session.AutomaticTax;
 
   /**
    * Describes whether Checkout should collect the customer's billing address. Defaults to `auto`.
    */
-  billing_address_collection: Checkout.Session.BillingAddressCollection | null;
+  billing_address_collection: Session.BillingAddressCollection | null;
 
-  branding_settings?: Checkout.Session.BrandingSettings;
+  branding_settings?: Session.BrandingSettings;
 
   /**
    * If set, Checkout displays a back button and customers will be directed to this URL if they decide to cancel payment and return to your website.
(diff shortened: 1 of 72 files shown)


testProjects/types-cjs-node16/typescriptTest.ts
```diff
@@ -39,6 +39,23 @@ async (): Promise<void> => {
 let opts: Stripe.RequestOptions;
 let apiList: Stripe.ApiList<Stripe.Customer>;
 
+// Resource sub-types (companion namespace access)
```
xavdid, line 42: can you add these tests to `types-cjs` as well? We should probably generate / static file copy them, but for now we just duplicate

src/resources/Checkout/Sessions.ts
```diff
@@ -976,2475 +976,2551 @@ export interface Session {
   /**
    * Wallet-specific configuration for this Checkout Session.
    */
-  wallet_options: Checkout.Session.WalletOptions | null;
+  wallet_options: Session.WalletOptions | null;
 }
-export namespace Checkout {
```
xavdid, line 981: are these not a breaking change? or they are, but they never worked, so we're fine with it?

xavdid (approved):
Looks good! had a comment about more test coverage, but looks good otherwise

testProjects/types-cjs-node16/typescriptTest.ts
jar-stripe: Ah I thought I did; it looks like it got added to types-cjs-node16.  Let me add them

src/resources/Checkout/Sessions.ts
jar-stripe: My understanding is this never worked through the public API.  See 
image
captured on node v22.2.2.  This change should fix the Alert.Status red squiggle.

testProjects/types-cjs-node16/typescriptTest.ts
jar-stripe: Fixed in [492d2dd](https://github.com/stripe/stripe-node/pull/2765/commits/492d2dd6a75603479e029ea02946e6d87195c91c)

PR #2783, opened 2026-07-10
Make Error fields generated

xavdid:
### Why?

Our main API Error class has always been handwritten. As new fields have been added to `api_errors` in the spec over time (e.g. `advice_code`, `network_advice_code`, `network_decline_code`), we haven't picked them up in the SDKs.

This makes most of those properties generated so they stay in sync.

### What?

- Makes the fields on the API error object generated from the OpenAPI spec
- Adds section markers so codegen can maintain these fields going forward
- add tests

### See Also

- [RUN_DEVSDK-2568](https://go/j/RUN_DEVSDK-2568)
- https://github.com/stripe/stripe-go/issues/2384


jar-stripe (commented):
Approving to unblock, but why do we have the changes in both Error.ts and Types.ts?

jar-stripe (approved):

xavdid:
Error is the actual runtime exception, while `type/RawError` is the interface for the JSON from the API. the JSON is fed into the runtime class during error handling

PR #2785, opened 2026-07-14
Implement extensibility platform

akalinin-stripe:
### Why?
* Support using Stripe SDK as part of the [Scripts](https://docs.stripe.com/extensions/how-extensions-work#scripts)

### What?

* Create a new platform, extensibility, to support this SDK when running scripts in Stripe environment.
* Minimal necessary plumbing to support using `extensibility` as a platform target
* Also implement a bare `StripeEventEmitter`, as it's not part of the runtime.

### DIFF
--- a/src/net/EndpointFetchHttpClient.ts
+++ b/src/net/EndpointFetchHttpClient.ts
@@ -1,0 +1,155 @@
+import {RequestHeaders} from '../Types.js';
+import {parseHttpHeaderAsString} from '../utils.js';
+import {
+  HttpClient,
+  HttpClientResponse,
+  HttpClientResponseInterface,
+  HttpClientRuntimeError,
+} from './HttpClient.js';
+
+type EndpointFetchRequest = {
+  endpoint: 'stripe_api';
+  path: string;
+  method: string;
+  body?: string;
+  headers?: Record<string, string>;
+};
+
+type EndpointFetchResponse = {
+  status: number;
+  body?: string | null;
+};
+
+type EndpointFetchSuccessResponse = EndpointFetchResponse & {
+  ok: true;
+};
+
+type EndpointFetchError = Error & {
+  status?: number;
+  body?: string | null;
+};
+
+type EndpointFetch = (
+  request: EndpointFetchRequest
+) => Promise<EndpointFetchSuccessResponse>;
+
+declare const endpointFetch: EndpointFetch | undefined;
+
+const STRIPE_API_HOST = 'api.stripe.com';
+
+export class EndpointFetchHttpClient extends HttpClient {
+  /** @override */
+  getClientName(): string {
+    return 'endpointFetch';
+  }
+
+  async makeRequest(
+    host: string,
+    port: string,
+    path: string,
+    method: string,
+    headers: RequestHeaders,
+    requestData: string,
+    protocol: string,
+    timeout: number
+  ): Promise<HttpClientResponseInterface> {
+    if (!path.startsWith('/')) {
+      throw new Error(`Only relative paths are supported, got: "${path}"`);
+    }
+
+    if (host !== STRIPE_API_HOST) {
+      throw new HttpClientRuntimeError(
+        `Stripe: This entrypoint only supports Stripe API requests to ${STRIPE_API_HOST}. Received request for ${host}.`
+      );
+    }
+
+    if (typeof endpointFetch !== 'function') {
+      throw new HttpClientRuntimeError(
+        'Stripe: `endpointFetch()` is not available.'
+      );
+    }
+
+    const methodHasPayload =
+      method == 'POST' || method == 'PUT' || method == 'PATCH';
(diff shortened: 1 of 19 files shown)


src/platform/PlatformFunctions.ts
```diff
@@ -2,18 +2,28 @@
 // eslint-disable-next-line wintertc-compat
 import * as http from 'http';
 import {CryptoProvider} from '../crypto/CryptoProvider.js';
-// TODO(DEVSDK-3113): Remove EventEmitter from shared base class in next major version.
```
xavdid, line 5: This comment leads me to believe this is a breaking change, but maybe it's not, given that everything is fine without changing any affected tests?

src/net/EndpointFetchHttpClient.ts
```diff
@@ -0,0 +1,155 @@
...
+type EndpointFetchError = Error & {
+  status?: number;
+  body?: string | null;
+};
+
+type EndpointFetch = (
+  request: EndpointFetchRequest
+) => Promise<EndpointFetchSuccessResponse>;
+
+declare const endpointFetch: EndpointFetch | undefined;
```
xavdid, line 36: Given that `EndpointFetch` is just a type, what does this `cosnt` declaration do?

src/net/EndpointFetchHttpClient.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    if (host !== STRIPE_API_HOST) {
+      throw new HttpClientRuntimeError(
+        `Stripe: This entrypoint only supports Stripe API requests to ${STRIPE_API_HOST}. Received request for ${host}.`
+      );
+    }
+
+    if (typeof endpointFetch !== 'function') {
+      throw new HttpClientRuntimeError(
+        'Stripe: `endpointFetch()` is not available.'
```
xavdid: Is there any more actionable information we can include here? How would a user debug/fix this if they see it?

src/net/EndpointFetchHttpClient.ts
```diff
@@ -0,0 +1,155 @@
...
+    this._res = res;
+  }
+
+  getRawResponse(): EndpointFetchResponse {
+    return this._res;
+  }
+
+  toStream(streamCompleteCallback: () => void): never {
+    throw new HttpClientRuntimeError(
+      'Stripe: Streaming responses are not supported in this runtime.'
```
xavdid: same here: this isn't super user-servicable. will they be aware of the runtime?

src/net/EndpointFetchHttpClient.ts
```diff
@@ -0,0 +1,155 @@
...
+    );
+  }
+
+  toJSON(): Promise<any> {
```
xavdid, line 144: some of these functions (like this one) are duplicated from `FetchHttpClientResponse`. Is it worth extending that instead of copying the code? or at least using a shared util function?

src/platform/ExtensibilityPlatformFunctions.ts
```diff
@@ -0,0 +1,98 @@
...
+      (alternative ? ` ${alternative}` : '')
+  );
+
+/**
+ * Platform functions for extensibility scripts.
+ */
+export class ExtensibilityPlatformFunctions extends PlatformFunctions {
+  /** @override */
+  emitWarning(_warning: string): void {
+    // Warnings have no host-visible sink in this runtime.
```
xavdid: what does "sink" mean in this context?

xavdid (approved):
Left a few small comments and questions, but nothing blocking!

src/net/EndpointFetchHttpClient.ts
akalinin-stripe: It's an ambient declaration, we need it here for compilation for runtime provided global value.

src/net/EndpointFetchHttpClient.ts
akalinin-stripe: I think so... Let me update.

src/net/EndpointFetchHttpClient.ts
akalinin-stripe: Let me make a _parseResponseBody helper

src/platform/ExtensibilityPlatformFunctions.ts
akalinin-stripe: I'll reword it a bit, there's no `console.log` in the extensibility runtime today.

src/platform/PlatformFunctions.ts
akalinin-stripe: It’s not a breaking change: the shared base now uses `StripeEmitterInterface`, which Node `EventEmitter` already satisfies. The runtime behavior remains unchanged.

src/net/EndpointFetchHttpClient.ts
akalinin-stripe: Updated this one as well.

PR #2792, opened 2026-07-21
Add V2SearchResultPromise and V2SearchResult

etienneshi-hub:
### Why?

V2 search api responses have a different shape than v1 search. We are adding the first v2 search api and will need to update sdk generation to return the right fields for the new search endpoint.

### What?

- Adds V2SearchResult<T> interface with object: 'v2.search_result', data, next_page_url, previous_page_url, and total_count fields
  - Adds V2SearchResultPromise<T> interface extending Promise<Response<V2SearchResult<T>>> with autoPagingEach and autoPagingToArray methods like V2ListPromise

### See Also

https://jira.corp.stripe.com/browse/DEVSDK-3211

## Changelog

- Adds V2SearchResult<T> and V2SearchResultPromise<T> types for v2 search endpoint responses with next_page_url/previous_page_url pagination and typed total_count


src/lib.ts
```diff
@@ -297,6 +297,49 @@ export interface V2ListPromise<T>
...
+export interface V2SearchResult<T> {
+  object: 'v2.search_result';
+
+  data: Array<T>;
+
+  /**
+   * Opaque cursor to fetch the next page of results.
+   * Pass back as the `page` request parameter. Null if no more pages.
+   */
+  next_page_url: string | null;
```
xavdid, line 315: Are these the only properties a list response will ever have? 

Also, is this data available in the openapi spec anywhere? It's the sort of thing we'd like to generate, where possible

src/lib.ts
```diff
@@ -297,6 +297,49 @@ export interface V2ListPromise<T>
...
+   */
+  total_count: number;
+}
+
+export interface V2SearchResultPromise<T>
+  extends Promise<Response<V2SearchResult<T>>>,
+    AsyncIterableIterator<T> {
+  autoPagingEach(
+    handler: (item: T) => boolean | void | Promise<boolean | void>,
+    onDone?: (err: any) => void
```
xavdid: the v1 `autoPagingEach` and `autoPagingToArray` doesn't have an `onDone` property. Why the mismatch?

src/lib.ts
```diff
@@ -297,6 +297,49 @@ export interface V2ListPromise<T>
...
+  /**
+   * The total number of objects that match the query.
+   */
+  total_count: number;
+}
+
+export interface V2SearchResultPromise<T>
+  extends Promise<Response<V2SearchResult<T>>>,
+    AsyncIterableIterator<T> {
+  autoPagingEach(
```
xavdid: You'll need to add these to iterator responses in `autoPagination.ts`. We currently guard against `apiMode !== 'v2'`, so that'll presumably change

src/stripe.esm.node.ts
```diff
@@ -31,6 +31,8 @@ import {
   ApiListPromise,
   V2List,
   V2ListPromise,
+  V2SearchResult,
+  V2SearchResultPromise,
```
xavdid, line 35: These need to be added in the matching `cjs` files

xavdid (changes requested):
Thanks for taking a look! There a few more things we need before we can merge.

In addition to my comments, you should add actual tests, plus type tests to ensure v2 search responses have the properties we expect. We don't have any in the generated methods yet (since you're just adding them here) but you can define one in the types test with a TODO to replace it with a real v2 search later.

etienneshi-hub:
The pagination for V2 search hasn't been gavel yet, but we seem to be close to a resolution soon. I'm cleaning this PR in preparation for it.

Are these sufficient [type tests](https://github.com/stripe/stripe-node/pull/2792/changes#diff-acd7a698ffc33cdfb4d141446a611eb0c9709bf974349fd74ae01feefe8c506bR116-R125)? I'm planning on adding tests after we add pagination as well, are there any other that I'm missing.

src/lib.ts
etienneshi-hub: These are the only fields defined in the API pattern review for v2 search.

The only v2 search endpoint currently is custom objects. I am not familiar enough with how the spec generation works for it in sdk generation, but happy to dig into it if you can point me in the right direction.

src/lib.ts
etienneshi-hub: I was originally modelling it against [v2List](https://github.com/stripe/stripe-node/pull/2792/changes/BASE..509e726a3db5c0febfff9cbeeec088873a25f6ac#diff-e59662795ab51f95e1bf2a44d30dd0debbc194a499adffd1bb6a197bd188d4b5R286-R298) which seemed to be the pattern we wanted to emulate. The final decision on the shape of v2 search pagination hasn't been made so I removed it for now

src/lib.ts
etienneshi-hub: When I was looking into this, I saw that [V2ListIterator](https://github.com/stripe/stripe-node/blob/master/src/autoPagination.ts#L172) didn't support bidirectional pagination. When we add support for V2SearchIterator would it make sense to create it as a separate method and support bidirection or update (and rename) V2ListIterator to support it as well. 

This is assuming that v2 search pagination works the same way as v2 list pagination

src/stripe.esm.node.ts
etienneshi-hub: Is that file autogenerated from sdk-codegen?
https://github.com/stripe/stripe-node/blob/master/src/stripe.cjs.node.ts#L14469

PR #2794, opened 2026-07-24
add/adjust event parsing helpers

xavdid:
### Why?

Historically, our webhook parsing logic has been closely tied to signature validation (because the expectation is that users verify webhooks' authenticity before use). But, as the webhook handling experience has grown in complexity over time, our helper methods haven't kept up.

In addition to the classic parse+verify & handle, we also support / encourage :

- using cloud providers (like [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid)), which don't require validation at all
- validating & ACKing an event quickly, then queuing it up for later processing
- unit tests w/ signed test payloads

This gives us 4 use cases that each need a subset of helper methods:

| case                                  | `verify` | `parse` | `parseAndVerify` | `signTestPayload` |
| ------------------------------------- | -------- | ------- | ---------------- | ----------------- |
| process immediately                   |          |         | ✅               | ✅                |
| verify, queue, process later          | ✅       | ✅      |                  | ✅                |
| process immediately (cloud provider)  |          | ✅      |                  |                   |
| queue, process later (cloud provider) |          | ✅      |                  |                   |

To support these workflows, we need all of these methods publicly available in every SDK:

1. `constructEvent(payload, signature, ...)` (parse + verify)
2. `parseEventNotification(payload, signature, ...)` (parse + verify)
3. `constructEventWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
4. `parseEventNotificationWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
5. `verify(payload, signature, secret, ...)`
6. `signTestPayload(payload, secret)`

The exact naming & arguments vary by language, but each SDK should provide roughly the same capabilities.

> [!IMPORTANT]
> For reviewers: the exact error messages are a little inconsistent between SDKs right now. Everything throws in the right place, but we're still iterating on some of the naming. We're trying out "thin event notification" as a descriptor on some of these surfaces. We're planning to have this all cleaned up by the GA launch of interop events; ignore exact messages for now.

### What?

> [!NOTE]
> these notes generic across all the SDKs; see `Changelog` below for the SDK-specific list of changes

- add missing event handling methods to `StripeClient` & `Webhook` classes, where relevant.
- Add an internal-facing `maybe_extract_from_cloud_provider_envelope` for handling wrapped events without the user having to think about it
- centralize some event parsing logic; adjusting accepted types for some methods
- adjust some error messages
- add tests
- update docstrings
- node: move webhook docstrings so they're actually user-facing
- node: make an error message more precise to help steer users

### See Also

- [stripe/stripe-java#2201](https://github.com/stripe/stripe-java/issues/2201)
- [DEVSDK-3085](https://go/j/DEVSDK-3085)
- [DEVSDK-3210](https://go/j/DEVSDK-3210)
- includes a fix for https://github.com/stripe/stripe-node/issues/2803

## Changelog

- Added methods that return their respective `Event`/`EventNotification` objects without verifying authenticity. Use them when you've previously verified an event (e.g. you verified, put the event in a queue, and are now processing). Supports events from [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid) natively.
  - `stripe.webhooks.constructEventWithoutVerification(payload)`
  - `stripe.constructEventWithoutVerification(payload)`
  - `stripe.parseEventNotificationWithoutVerification(payload)`


testProjects/types-cjs-node16/typescriptTest.ts
```diff
@@ -60,6 +60,12 @@ let accountBizRevenue: Stripe.AccountCreateParams.BusinessProfile.AnnualRevenue;
 // @ts-expect-error - unknown config properties should be rejected
 const bad = new Stripe('sk_test_123', {unknownProperty: true});
 
+// Webhook methods: constructEventWithoutVerification and parseEventNotificationWithoutVerification
+event = stripe.webhooks.constructEventWithoutVerification('payload');
+event = stripe.constructEventWithoutVerification('payload');
```
kidus-stripe, line 68: Is event supposed to be overwritten here? same in the other typescriptTest files

kidus-stripe (commented):
Looks good, left one comment

testProjects/types-cjs-node16/typescriptTest.ts
xavdid: it doesn't matter - this code isn't ever executed. It's just there so we can verify that the TS compiler runs without error. We don't even really need to assign it to a variable (since the value is never used)

src/utils.ts
```diff
@@ -465,6 +465,44 @@ export function parseHeadersForFetch(
...
+  // Could add as many checks as we want here, but we'll start simple
+  if ('detail' in parsed) {
+    // AWS
+    // https://docs.stripe.com/event-destinations/eventbridge#event-structure
+    return parsed.detail as Record<string, unknown>;
+  }
+  if ('specversion' in parsed) {
+    // Azure
+    // https://docs.stripe.com/event-destinations/eventgrid#event-structure
+    return parsed.data as Record<string, unknown>;
```
jar-stripe, line 495: could return unknown if data is missing; prefer to throw an error here

jar-stripe (approved):

PR #2798, opened 2026-07-30
Add discriminated union serialization tests

jar-stripe:
### Why?

Validates that the discriminated union serialization design works correctly in stripe-node before codegen produces these patterns. Exercises request-side encoding via `flattenAndStringify` (the SDK's actual form-param encoder) and response-side TypeScript type narrowing.

### What?

- Adds `test/DiscriminatedUnion.spec.ts` with 29 tests covering:
  - **Request standalone DU encoding via `flattenAndStringify`**: Verifies bracket-notation output for nested variant params (`color[model]=rgb`, `color[r]=255`), optional field omission, and discriminator + payload encoding
  - **Request inline DU encoding via `flattenAndStringify`**: Verifies top-level discriminator, nested variant payload encoding (`card[number]=4242`), base fields at top level, and non-selected variant absence
  - **Response standalone DU type narrowing**: TypeScript compile-time verification that discriminator-based narrowing resolves to correct variant types
  - **Response inline DU deserialization**: JSON round-trip into intersection types (`BaseFields & (VariantA | VariantB)`), discriminator-based field access, non-selected variant null check, full base+variant field preservation
  - **Round-trip**: `flattenAndStringify` for request direction, JSON parse for response direction

### DIFF
--- a/test/DiscriminatedUnion.spec.ts
+++ b/test/DiscriminatedUnion.spec.ts
@@ -1,0 +1,260 @@
+// @ts-nocheck
+import {expect} from 'chai';
+
+// ---------------------------------------------------------------------------
+// Type definitions mirroring the shapes the codegen produces for discriminated
+// unions.  These are hand-written here so the test file is self-contained, but
+// they exactly match the object-literal typedef shape the Node generator emits
+// for each variant.
+// ---------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


jar-stripe:
Collapsed into #2801 (coercion PR now includes serialization tests)

PR #2799, opened 2026-07-31
Add `stripe.major_api_version` constant

xavdid:
### Why?

We've been generating information about the current API version into the SDKs for a while (codegen PR `2555`), but accidentally didn't make those constants user-facing.

### What?

- adds a user-facing `major_api_version` constant set to the generated value (currently `"dahlia"`)

### See Also

- (originally from https://github.com/stripe/stripe-java/issues/2001)


jar-stripe (approved):
watch the creep in e.g. CLAUDE.md; otherwise this look good!

PR #2801, opened 2026-08-04
Add discriminatedUnion kind to V2RuntimeSchema coercion

jar-stripe:
### Why?

We need to enhance v2 type coercion to be aware of discriminated unions to ensure variant fields that are of e.g. `int64_string` or `decimal_string` are properly coerced.

### What?

- adds `{kind: 'discriminatedUnion', discriminator, variants}` to `V2RuntimeSchema` in `Types.ts`.
- handles that kind in both `coerceV2RequestData` and `coerceV2ResponseData`: reads `data[schema.discriminator]`, looks up the variant, and recurses into only that variant's fields.
- throws on the request side when the discriminator is missing, or present but not a string. The error names the field and lists the valid variants.
- passes response data through when the discriminator does not resolve, so a variant the API ships after this release does not break existing clients.
- passes an unrecognized but string-valued discriminator through on both sides, preserving support for undocumented params sent with the right shape.
- adds tests for matched variants, and for unknown, missing, and non-string discriminators on both sides.

### See Also

None.

## Changelog

- Adds runtime support for coercing int64/decimal fields within discriminated union variants in V2 API requests and responses.
- ⚠️ Throws when serializing a discriminated union parameter whose discriminator is missing or is not a string. This avoids skipping coercion entirely which could silently truncate `int64` fields above `Number.MAX_SAFE_INTEGER`.


xavdid (approved):
This could also be a good opportunity for type narrowing tests! (which we have already)

PR #2819, opened 2026-08-20
Merge to beta

xavdid:


xavdid:
accidentially re-ran the command and blew this PR away 🤦 recreating

PR #2829, opened 2026-08-28
Harden API requestor code against malicious URLs

xavdid:
### Why?

If an attacker got access to a user's webhook secret (or the user wasn't validating incoming webhook signatures) _and_ a user was calling the built-in `event_notification.fetch_related_object()` method, a malicious webhook payload could cause a user integration to make an authenticated request to an attacker-controlled domain (leaking the secret key)

To solve, we're being more careful during url construction to ensure user input can't be used to send requests to non-`stripe.com` domains (on a per-request basis).

### What?

- add path validation method and call it before the final URL is built
- Encode the `event.id` path param in `fetch_event()` to match the sanitation in the rest of the generated endpoints
- add tests

### See Also

- [RUN_DEVSDK-2808](https://go/j/RUN_DEVSDK-2808)

### DIFF
--- a/src/stripe.core.ts
+++ b/src/stripe.core.ts
@@ -1626,7 +1626,9 @@ export class Stripe {
     parsed.fetchEvent = (): Promise<unknown> => {
       return this._requestSender._rawRequest(
         'GET',
-        `/v2/core/events/${parsed.id}`,
+        // `id` comes from the notification body, so encode it the way the
+        // generated resources do -- otherwise it can inject path or query segments.
+        `/v2/core/events/${encodeURIComponent(parsed.id as string)}`,
         undefined,
         {
           stripeContext: parsed.context as any,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -444,19 +444,16 @@ export function jsonStringifyRequestData(data: RequestData | string): string {
  *
  * Some HTTP clients build the absolute URL by concatenating a host onto this
  * path, and the configured hosts carry no trailing slash. A path like
- * '@evil.example/v1/x' or '.evil.example/v1/x' would then land inside the
- * authority component and send the request -- Authorization header included --
- * to a host of the path's choosing. Some request paths originate in remote data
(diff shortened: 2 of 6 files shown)


src/stripe.core.ts
```diff
@@ -1626,7 +1626,9 @@ export class Stripe {
     parsed.fetchEvent = (): Promise<unknown> => {
       return this._requestSender._rawRequest(
         'GET',
-        `/v2/core/events/${parsed.id}`,
+        // `id` comes from the notification body, so encode it the way the
+        // generated resources do -- otherwise it can inject path or query segments.
+        `/v2/core/events/${encodeURIComponent(parsed.id as string)}`,
```
jar-stripe, line 1631: is there any way parsed.id could be something other than a string?  would this be better as encodeURIComponent(parsed.id.toString()) or something similar?

src/utils.ts
```diff
@@ -438,6 +438,37 @@ export function jsonStringifyRequestData(data: RequestData | string): string {
   return JSON.stringify(data, dateTimeReplacer);
 }
 
+/**
+ * Asserts that a request path is origin-relative: that it begins with a single
+ * '/'.
+ *
+ * Some HTTP clients build the absolute URL by concatenating a host onto this
+ * path, and the configured hosts carry no trailing slash. A path like
+ * '@evil.example/v1/x' or '.evil.example/v1/x' would modify the resulting host
```
jar-stripe, line 447: similar to comment in stripe-java, this feels like a lot of exposition for validatePath.  is there a better place to put this?

jar-stripe (commented):
approving to unblock; check the comments for a couple things to consider before merging.

src/stripe.core.ts
xavdid: not if it's well-formed. plus `encodeURIComponent` is very accepting, but TS wanted me to translate the `unknown` into _something_

jar-stripe (approved):

PR #2831, opened 2026-08-29
Use cryptographically secure boundaries for multipart file uploads

xavdid:
### Why?

Fixes a bug where the boundary markers in multi-part file uploads weren't cryptographically secure. As a result, it could be possible for an attacker-controlled file to specify additional properties in the upload request.

Similarly, a maliciously crafted filename could be used to add additional properties to the request.

This only affects the `files.create` method.

### What?

- use a cryptographically secure method of generating multipart file boundaries
- sanitize filename input before adding to requests
- make UUID cryptographically secure and use it for both idempotency key generation and multipart boundaries
- add fallback for idempotency key generation for environments without the global `crypto` module (which i'm not sure we need to worry about? it's maybe just pre-2022 browsers)
- add tests

### See Also

- [RUN_DEVSDK-2807](https://go/j/RUN_DEVSDK-2807)

### DIFF
--- a/src/platform/NodePlatformFunctions.ts
+++ b/src/platform/NodePlatformFunctions.ts
@@ -21,11 +21,10 @@ class StreamProcessingError extends StripeError {}
 export class NodePlatformFunctions extends PlatformFunctions {
   /** @override */
   uuid4(): string {
-    // available in: v14.17.x+
-    if (crypto.randomUUID) {
-      return crypto.randomUUID();
-    }
-    return super.uuid4();
+    // Available since v14.17, so available on every Node versions we support
+    // The base PlatformFunctions implementation reads `globalThis.crypto`, which is only unflagged as of Node 19 (and we support down to Node 18 until Sept 2026)
+    // TODO(https://go/j/DEVSDK-3253)
+    return crypto.randomUUID();
   }
 
   /** @override */
--- a/src/platform/NodePlatformFunctions.ts
+++ b/src/platform/NodePlatformFunctions.ts
@@ -21,8 +21,7 @@ class StreamProcessingError extends StripeError {}
 export class NodePlatformFunctions extends PlatformFunctions {
   /** @override */
   uuid4(): string {
-    // Available since v14.17, so available on every Node versions we support
-    // The base PlatformFunctions implementation reads `globalThis.crypto`, which is only unflagged as of Node 19 (and we support down to Node 18 until Sept 2026)
+    // we need to override the base `uuid4` implementation because it uses `globalThis.crypto`, which isn't available in unflagged in node until Node 19
     // TODO(https://go/j/DEVSDK-3253)
     return crypto.randomUUID();
   }
--- a/src/platform/PlatformFunctions.ts
+++ b/src/platform/PlatformFunctions.ts
@@ -93,14 +93,21 @@ export class PlatformFunctions {
   }
 
   /**
-   * Generates a v4 UUID. See https://stackoverflow.com/a/2117523
+   * Generates a v4 UUID. Must be cryptographically secure: this seeds both
+   * `Idempotency-Key` values and the multipart/form-data boundary.
+   *
+   * `crypto.randomUUID` is part of the WinterTC Minimum Common API, so it is
+   * available in browsers, Deno, Bun and Cloudflare Workers.
+   *
+   * Deliberately throws rather than degrading to `Math.random()`.
    */
   uuid4(): string {
-    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
-      const r = (Math.random() * 16) | 0;
-      const v = c === 'x' ? r : (r & 0x3) | 0x8;
-      return v.toString(16);
-    });
+    if (typeof crypto === 'undefined' || !crypto.randomUUID) {
+      throw new Error(
+        'Stripe: no cryptographically secure random number generator is available in this runtime; `crypto.randomUUID` is required.'
+      );
+    }
+    return crypto.randomUUID();
   }
 
   /**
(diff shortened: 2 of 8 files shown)


xavdid:
actually tests are failing, so I may need to tweak

src/platform/NodePlatformFunctions.ts
```diff
@@ -21,11 +21,10 @@ class StreamProcessingError extends StripeError {}
 export class NodePlatformFunctions extends PlatformFunctions {
   /** @override */
   uuid4(): string {
-    // available in: v14.17.x+
-    if (crypto.randomUUID) {
-      return crypto.randomUUID();
-    }
-    return super.uuid4();
+    // Available since v14.17, so available on every Node versions we support
```
jar-stripe: i think we can kill this line; node 14 support isnt germane given we support 18+ now

src/platform/PlatformFunctions.ts
```diff
@@ -93,14 +93,21 @@ export class PlatformFunctions {
...
+   * Generates a v4 UUID. Must be cryptographically secure: this seeds both
+   * `Idempotency-Key` values and the multipart/form-data boundary.
+   *
+   * `crypto.randomUUID` is part of the WinterTC Minimum Common API, so it is
+   * available in browsers, Deno, Bun and Cloudflare Workers.
+   *
+   * Deliberately throws rather than degrading to `Math.random()`.
    */
   uuid4(): string {
-    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
```
jar-stripe, line 99: do we know what platforms this will affect?

src/platform/PlatformFunctions.ts
```diff
@@ -93,14 +93,17 @@ export class PlatformFunctions {
...
-      const r = (Math.random() * 16) | 0;
-      const v = c === 'x' ? r : (r & 0x3) | 0x8;
-      return v.toString(16);
-    });
+    if (typeof crypto === 'undefined' || !crypto.randomUUID) {
```
jar-stripe, line 101: we should align with stripe apps to confirm what browser versions they support, or if they are comfortable aligning with the WinterTC interop rules.  Otherwise, we may need to support a legacy browser mode here.

test/PlatformFunctions.spec.ts
```diff
@@ -14,9 +14,10 @@ import {SubtleCryptoProvider} from '../src/crypto/SubtleCryptoProvider.js';
 import {expect} from 'chai';
 import {webcrypto} from 'crypto';
 
-if (process.versions.node < '15') {
+if (process.versions.node < '19') {
+  // Node 18 has no `globalThis.crypto` in CommonJS module scope, so WebPlatformFunctions has no CSPRNG to derive a boundary from.
```
jar-stripe: iiuc the above code uses an override for this, so we don't need globalThis.crypto. Is that right?

jar-stripe: Either way, we should make the comments and/or implementations consistent between here and the NodePlatformFunctions file

test/PlatformFunctions.spec.ts
```diff
@@ -438,3 +445,93 @@ function testPlatform(platformFunctions: PlatformFunctions): void {
     });
   });
 }
+
+describe('PlatformFunctions.uuid4 without crypto.randomUUID', () => {
+  // `crypto.randomUUID` is secure-context-only, so it is absent on plain
```
jar-stripe: small nit; this comment feels unnecessary in the test; if its valuable, maybe move it to the implementation?

test/PlatformFunctions.spec.ts
```diff
@@ -438,3 +445,93 @@ function testPlatform(platformFunctions: PlatformFunctions): void {
...
+    } finally {
+      Math.random = random$;
+    }
+  });
+});
+
+// TODO(https://go/j/DEVSDK-3253) - can remove when we drop node 18
+describe('NodePlatformFunctions.uuid4 without globalThis.crypto', () => {
+  // This is the property that keeps Node 18 working: Node 18 exposes no
+  // `globalThis.crypto` in CommonJS module scope (it landed unflagged in Node
```
jar-stripe: this comment feels overly wordy, and we reference node 18 and globalThis.crypto several times.  can you take a pass thru the comments to make them clearer before merging?

jar-stripe (changes requested):
A lot of small things to touch up, and before we merge we should align with stripe apps.

src/platform/PlatformFunctions.ts
xavdid: This is browsers older than 2022 ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)). All our node versions are covered. I'm checking with extensibility

src/platform/PlatformFunctions.ts
xavdid: asking!

src/platform/PlatformFunctions.ts
xavdid: extensibility doesn't support files and we have a fallback for idempotency keys, so I think we're good here (once we hear from apps)

test/PlatformFunctions.spec.ts
xavdid: The base implementation uses `globalThis.crypto`, which is only available in node 19+. So we skip running the web tests in node 18, since they just fail.

Truth be told we don't have to re-run the web tests across node versions, since they're supposed to be node version agnostic. but it's easier to just run them than pin them to any specific versions.

test/PlatformFunctions.spec.ts
xavdid: fair! this whole test is really only needed for node 18, so i'll scope it down

xavdid:
ok, cleaned up some comments! am waiting to hear from apps, but I'd be surprised if they support browsers back to 2021

jar-stripe (approved):

PR #2850, opened 2026-09-14
Fix event notification callback client context

zacchua-stripe:
### Why?

`EventNotificationHandler` shallow-copied the original Stripe client when applying an event’s Stripe context. Resources and the request sender remained bound to the original client, causing callback API requests to use the original context.

### What?

- Construct a new client for each event context that reuses the original authentication, configuration and HTTP transport
- Add regression test asserting the outbound `Stripe-Context` header. 

### See Also

http://go/j/RUN_DEVSDK-3035

## Changelog

- Fixes event notification callback API requests using the original client’s Stripe context.


src/StripeEventNotificationHandler.ts
```diff
@@ -126,19 +126,43 @@ class BaseEventNotificationHandler {
     return keys;
   }
 
+  private createClientWithContext(
```
xavdid, line 129: nit: the other SDKs call with `withStripeContext`, so we should maybe use that for consistency

xavdid: it should also probably live on the client itself, so it's responsible for creating copies of itself

src/StripeEventNotificationHandler.ts
```diff
@@ -126,19 +126,43 @@ class BaseEventNotificationHandler {
...
+      maxNetworkRetries: this.client.getApiField('maxNetworkRetries'),
+      httpClient: this.client.getApiField('httpClient'),
+      timeout: this.client.getApiField('timeout'),
+      host: this.client.getApiField('host'),
+      port: this.client.getApiField('port'),
+      protocol: this.client.getApiField('protocol'),
+      telemetry: this.client.getTelemetryEnabled(),
+      emitEventBodies: this.client.getEmitEventBodiesEnabled(),
+      appInfo: this.client._appInfo,
+      stripeAccount: this.client.getApiField('stripeAccount') ?? undefined,
```
xavdid, line 150: 🚨 we _don't_ want to copy this from the original client. it's a server error to pass both stripe-context and stripe-account (unless they match). so we want to explicitly drop that field. Apparently python, java, php, and ruby have the same bug (based on when I was looking at this earlier), so you'll want to fix it here and then put up PRs for the rest as well (plus add tests)

src/StripeEventNotificationHandler.ts
```diff
@@ -126,19 +126,43 @@ class BaseEventNotificationHandler {
...
   }
 
+  private createClientWithContext(
+    stripeContext: Stripe.V2.Core.EventNotification['context']
+  ): Stripe {
+    // Use the runtime constructor so this works for the separate Node ESM and
+    // shared core clients. Reuse configuration and the HTTP client, but create
+    // new resources and a new request sender bound to the event-scoped client.
+    const StripeClient = this.client.constructor as typeof Stripe;
+    const eventClient = new StripeClient('', {
```
xavdid, line 136: one thing that's not being copied over is the original `_emitter`, which means we won't send request metrics we've been holding. consider if that's worth fixing

xavdid (approved):
looks good, couple of small comments. but make sure you fix the account issue & add a test! that's a real bug

zacchua-stripe:
Thanks for the comments @xavdid! Do you mind taking a second pass? Thank you! 🙏

.hark/changes/2026-09-15_zacchua_fix-event-callback-context.change.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fix event notification callback client context
+pr_url: https://github.com/stripe/stripe-node/pull/2850
+semver_level: minor
+---
+
+Create event notification callback clients with the event's Stripe context without inheriting a conflicting Stripe account from the original client.
```
xavdid: you have no way of knowing this yet (working on it!) but these should be bulleted items, so add a leading `-`, e.g. `- Create event no...`

.hark/changes/2026-09-15_zacchua_fix-event-callback-context.change.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fix event notification callback client context
+pr_url: https://github.com/stripe/stripe-node/pull/2850
+semver_level: minor
```
xavdid: should be a patch, this is just a bugfix!

.hark/changes/2026-09-15_zacchua_fix-event-callback-context.change.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Fix event notification callback client context
```
xavdid: This title isn't super descriptive to an end user. What's the actual bug we're fixing?

.hark/changes/2026-09-15_zacchua_fix-event-callback-context.change.md
xavdid: Separately, if you want to add this line, make it specific to what a user might experience. Why do they care that we made this change?

Think about how you'd write it, then click this ⬇️ to see what I'd say. fft to use mine, edit it, combine mine with yours, or just use yours if it fits the spirit of what I'd say!

<details>
<summary>What i'd say</summary>


- Fix a bug where an `EventNotificationHandler`'s event-bound client wasn't sending the correct `Stripe-Client` header
- Fix a bug that could cause 400 errors when using a `StripeClient` with the `stripeAccount` arg in an `EventNotificationHandler`


</details>

xavdid (approved):
some nits about the changefile but nothing blocking!

.hark/changes/2026-09-15_zacchua_fix-event-callback-context.change.md
zacchua-stripe: Updated the title to `Fix account scoping for EventNotificationHandler callback clients`!

PR #2856, opened 2026-09-18
Remove V1 fields from V2 list responses

jar-stripe:
### Why?

V2 list responses use `{data, next_page_url, previous_page_url}`, but `Stripe.V2List<T>` temporarily declared the V1-only `object`, `has_more`, and `url` fields for backwards compatibility. Those fields are never populated at runtime and should be removed in the next major release.

### What?

- Remove the V1-only fields from `Stripe.V2List<T>`.
- Update runtime and type tests to enforce the V2 response shape.
- Add v23 migration guidance for users who typed V2 responses as `Stripe.ApiList<T>`.

### Test plan

`just test`

### See Also

- [DEVSDK-2534](https://go/j/DEVSDK-2534)

## Configuration

- [ ] `skip-changefile`: This PR is not a user-facing change, so there's no changefile.


mark-kogan-stripe (approved):
Looks good!

